PR #662, opened 2021-07-28
Feat/button colors

feerglas:
## Preflight Checklist

* [ ] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/master/.github/CONTRIBUTING.md) for this project.
* [ ] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
* [ ] I have searched the [pull request tracker](https://github.com/lyne-design-system/lyne-components/pulls) for a Pull Request (PR) that matches the one I want to submit, without success.

## Issue

This PR Closes #

## Pull request checklist

## Changes

Changes in this pull request:

-

## Browsers

- [ ] Firefox Desktop
- [ ] Chrome Desktop
- [ ] Safari Desktop
- [ ] Firefox Mobile
- [ ] Chrome Mobile
- [ ] Safari Mobile
- [ ] Opera
- [ ] Brave

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

- [ ] Yes
- [ ] No

## Mentions

@lyne-design-system/core


src/global/variables.scss
```diff
@@ -1,5 +1,2 @@
-@import './node_modules/lyne-design-tokens/dist/scss/_variables.scss';
-
-:host {
-  --app-primary-color: blue;
-}
+//@import './node_modules/lyne-design-tokens/dist/scss/_variables.scss';
```
mbleuer, line 1: This one you want to keep for now, just in case?

src/components/lyne-cta-button/lyne-cta-button.tsx
```diff
@@ -24,6 +25,9 @@ export class LyneCtaButton {
   /** Label text to show on the button */
   @Prop() public label = 'Default button text';
 
+  /** Type of the button, like primary, secondary etc. */
+  @Prop() public type?: InterfaceButtonAttributes['type'];
```
mbleuer: I'm wondering, whether we should name this 'variant' or something similar, since type is also used an official attribute already (button, submit, reset).

src/components/lyne-cta-button/lyne-cta-button.tsx
```diff
@@ -24,6 +25,9 @@ export class LyneCtaButton {
   /** Label text to show on the button */
   @Prop() public label = 'Default button text';
 
+  /** Type of the button, like primary, secondary etc. */
+  @Prop() public type?: InterfaceButtonAttributes['type'] = 'primary';
```
mbleuer, line 29: I'm wondering, whether we should better name that something like 'variant' since type is already used as an offical attribute of button (button, submit, reset).

src/global/variables.scss
feerglas: Yes. As we discussed, we most probably will have variables that we need to define in scss. Can't think of a specific example now, but i'd like to keep it in case rather then to rebuild this (and the change in stencil.config) later on.

src/components/lyne-cta-button/lyne-cta-button.tsx
```diff
@@ -46,9 +50,11 @@ export class LyneCtaButton {
   };
 
   public render(): JSX.Element {
-    return <button class='button' onClick={this._buttonClick}>
-      <span class='label'>{this.label}</span>
-      <span class='arrow' innerHTML={ButtonIcon} />
+    const typeClass = `button button--${this.type}`;
```
mbleuer, line 53: If we would decide to go with another naming (see comment above), we would have to adjust here as well.

src/components/lyne-cta-button/lyne-cta-button.tsx
feerglas: Good point! Completely ignored this one, will change it to `variant`

src/components/lyne-cta-button/lyne-cta-button.tsx
feerglas: yes, in some other places as well, like in the interface declarations, the tests, and the stories file

src/components/lyne-cta-button/lyne-cta-button.stories.js
```diff
@@ -1,12 +1,33 @@
 import events from './lyne-cta-button.events.ts';
 import { h } from 'jsx-dom';
 
+const types = {
+  control: {
+    type: 'radio'
```
mbleuer, line 6: See comments below regarding the naming. Maybe variant is not the right fit either, because the actuall variants could be "Text with icon", "Text only", "Icon only", which again could come in all visual characteristics (primary, secondary, etc).

feerglas: I already changed it here, please pull latest

feerglas: As mentioned, let's try to define the component-internal variants "Text with icon", "Text only", "Icon only" via the attributes for `label` (existing) and `icon` (not yet implemented)

mbleuer (approved):

PR #693, opened 2021-08-16
Feat/clock

mbleuer:
## Preflight Checklist

* [ ] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/master/.github/CONTRIBUTING.md) for this project.
* [ ] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
* [ ] I have searched the [pull request tracker](https://github.com/lyne-design-system/lyne-components/pulls) for a Pull Request (PR) that matches the one I want to submit, without success.

## Issue

This PR Closes #

## Pull request checklist

## Changes

Changes in this pull request:

-

## Browsers

- [ ] Firefox Desktop
- [ ] Chrome Desktop
- [ ] Safari Desktop
- [ ] Firefox Mobile
- [ ] Chrome Mobile
- [ ] Safari Mobile
- [ ] Opera
- [ ] Brave

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes

- [ ] Yes
- [ ] No

## Mentions

@lyne-design-system/core


src/components/lyne-clock/lyne-clock.tsx
```diff
@@ -0,0 +1,209 @@
...
+@Component({
+  shadow: true,
+  styleUrl: 'lyne-clock.scss',
+  tag: 'lyne-clock'
+})
+
+export class LyneClock {
+
+  @Prop() public paused = false;
+  @Prop() public initialtime = 'now';
```
feerglas, line 24: Especially here, a Prop description makes sense, to let users now what values they can pass in.
```
/** Pass a string with hours:minutes:seconds as a starting time for the clock, like 04:15:59 */
@Prop() public initialtime? = 'now';
```

feerglas, line 24: Or i might have an even better proposal. Initialtime can have either a string (now) or a time-string (08:15:59). This could be type-checked with literals. But in my opinion this is a logical mismatch, since you're handling the two indescrete types differently. In my opinion, this would work more safely and would be better understandable:
1. if initialtime is not set, you handle it like it would be set to "now"
2. for the initialtime, you define a type:
  -> create a file `lyne-clock.d.ts` with the following: 
``` export type Time = `${number}:${number}:${number}`;```
  
  -> in your main file, import it like so: `import { Time } from './lyne-clock.d';`
  -> then use the type like so: `@Prop() public initialtime?: Time;`
  
That way, we are at least sure about getting a string which has the format "number:number:number". And we have a concise type for initialtime

**Review by feerglas — COMMENTED** [review]
Without digging deeper into the issue: i switched tabs, when returning to the clock, it was 1 minute off

Bildschirmfoto 2021-08-16 um 19 41 13

File: src/components/lyne-clock/lyne-clock.scss [hunk 2]
```diff
@@ -0,0 +1,95 @@
...
+
+.clock__hand-hours {
+  animation-name: rotate;
+  animation-duration: 43200s;
+  animation-iteration-count: infinite;
+  animation-timing-function: linear;
+  animation-play-state: var(--clock-animation-play-state);
+}
+
+.clock__hand-hours.clock__hand-hours--initial-hour {
```
feerglas, line 35: I am confused about these BEM selectors. In my opinion, when using a modifier class, you should not have to use the element class as well, e.g. `.clock__hand-hours--initial-hour` instead of `.clock__hand-hours.clock__hand-hours--initial-hour`. If you're doing it because of specifity, to make sure these styles get precedence over the styles of `.clock__hand-hours`, that probably means that you should have to define kind of a default modifier for the styles you are overwriting here, like so:
```
.clock__hand-hours {
  animation-timing-function: linear;
}

.clock__hand-hours--default {
  animation-name: rotate;
  animation-iteration-count: infinite;
  animation-duration: 43200s;
  animation-play-state: var(--clock-animation-play-state);
}

.clock__hand-hours--initial-hour {
  animation-name: handHoursAnimationInitialHour;
  animation-iteration-count: 1;
  animation-duration: var(--clock-hours-animation-duration);
  animation-play-state: var(--clock-animation-play-state);
}
```

feerglas (changes requested):
Animation looks super smooth, performance is great 🤘👍

src/components/lyne-clock/lyne-clock.scss
mbleuer: You're absolutey right, the specifity was obsolete.

feerglas:
Oh, just one other think: we need to have a solution for Chromatic. Since Chromatic is making screenshots of the default variant (with time="now"), we would have the clock displaying a different time in the screenshots for the different viewports. Furthermore and even worse, we will most likely have a visual difference in the clock for every pull request that we make.

feerglas:
We could workaround it with Javascript...

https://www.chromatic.com/docs/animations

Bildschirmfoto 2021-08-18 um 08 51 51

src/components/lyne-clock/lyne-clock.stories.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+if (isChromatic()) {
+  clock.args = {
+    initialTime: '09:43:59',
+    paused: true
+  };
+} else {
+  clock.args = {
+    initialTime: times.options[0],
+    paused: state.paused
```
feerglas, line 51: I would set this directly to `false` and delete the `state` const

feerglas:
I have another improvement. Currently, when loading the page, the clock is initially at 00:00 and then jumps to the defined time. We can get around this by hiding the handles until the clock is fully initialized:

Import state from stencil:
```
import {
  Component,
  Element,
  h,
  Prop,
  State
} from '@stencil/core';
```
Define a state to keep track of initialisation:
`@State() private _isInitialized = false;`

Set it to true as soon as init is done:
```
private _moveHandsInitially(): void {
    ...
    this._clockHandSeconds.classList.add('clock__hand-seconds--initial-minute');
    this._clockHandHours.classList.add('clock__hand-hours--initial-hour');
    this._element.style.setProperty('--clock-animation-play-state', 'running');

    this._isInitialized = true;
}
```
Use it in the render method to add an appropriate class:
```
    const initClass = this._isInitialized
      ? ''
      : ' clock--not-initialized';

    return <div class={`clock${initClass}`}>
      <span
      ...
```
Then we can hide the handles based on that class:
```
.clock--not-initialized .clock__hand-hours,
.clock--not-initialized .clock__hand-minutes,
.clock--not-initialized .clock__hand-seconds  {
  display: none;
}
```

feerglas (approved):

PR #750, opened 2021-09-24
Feat/teaser hero

feerglas:


mbleuer (changes requested):
Nice work! Let's have a chat over the suggestions.

PR #832, opened 2021-11-01
Feat/lyne link

mbleuer:
This PR also includes a suggestion how we could deal with i18n text which is closely related to the module (e.g. "Link target opens in new window.") and should be consistend across the various locations of use.

### DIFF
--- a/src/components/lyne-link/lyne-link.e2e.ts
+++ b/src/components/lyne-link/lyne-link.e2e.ts
@@ -6,7 +6,7 @@ describe('lyne-link', () => {
 
   it('renders', async () => {
     page = await newE2EPage();
-    await page.setContent('<lyne-link></lyne-link>');
+    await page.setContent('<lyne-link href-value="https://github.com/lyne-design-system/lyne-components" iconFlip="false"></lyne-link>');
 
     element = await page.find('lyne-link');
     expect(element)
--- a/src/components/lyne-link/lyne-link.e2e.ts
+++ b/src/components/lyne-link/lyne-link.e2e.ts
@@ -6,7 +6,7 @@ describe('lyne-link', () => {
 
   it('renders', async () => {
     page = await newE2EPage();
-    await page.setContent('<lyne-link href-value="https://github.com/lyne-design-system/lyne-components" iconFlip="false"></lyne-link>');
+    await page.setContent('<lyne-link href-value="https://github.com/lyne-design-system/lyne-components"></lyne-link>');
 
     element = await page.find('lyne-link');
     expect(element)
--- a/src/components/lyne-link/lyne-link.e2e.ts
+++ b/src/components/lyne-link/lyne-link.e2e.ts
@@ -1,0 +1,16 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('lyne-link', () => {
+  let element,
+    page;
+
+  it('renders', async () => {
+    page = await newE2EPage();
+    await page.setContent('<lyne-link></lyne-link>');
+
+    element = await page.find('lyne-link');
+    expect(element)
+      .toHaveClass('hydrated');
+  });
+
+});
--- a/src/components/lyne-link/lyne-link.stories.js
+++ b/src/components/lyne-link/lyne-link.stories.js
@@ -2,22 +2,9 @@ import {
   ColorCharcoalDefault,
   ColorWhiteDefault
 } from 'lyne-design-tokens/dist/js/tokens.es6';
+import getMarkupForSvg from '../../global/helpers/get-markup-for-svg';
 import { h } from 'jsx-dom';
-import lyneIcons from 'lyne-icons/dist/icons.json';
 import readme from './readme.md';
-
-// --- Helper methods
-const getMarkupForSvg = (svgName) => {
-  if (!svgName) {
-    return '';
-  }
-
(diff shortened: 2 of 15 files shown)


src/components/lyne-link/lyne-link.e2e.ts
```diff
@@ -0,0 +1,16 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('lyne-link', () => {
+  let element,
+    page;
+
+  it('renders', async () => {
+    page = await newE2EPage();
+    await page.setContent('<lyne-link href-value="https://github.com/lyne-design-system/lyne-components" iconFlip="false"></lyne-link>');
```
4aficiona2, line 9: `iconFlip="false"` required?

4aficiona2 (commented):
Looks great, thank you!

I mentioned some smaller things and we might have to introduce some more attributes or define how we handle them.

One thing is the naming which I already mentioned in the corresponding files. The other thing is, do we have to introduce also `rel`, `role`, `id`. 

Had a look over here which brought up my questions.

https://www.duetds.com/components/link/
https://github.com/carbon-design-system/carbon-web-components/blob/master/src/components/link/link.ts

mbleuer:
> Looks great, thank you!
> 
> I mentioned some smaller things and we might have to introduce some more attributes or define how we handle them.
> 
> One thing is the naming which I already mentioned in the corresponding files. The other thing is, do we have to introduce also `rel`, `role`, `id`.
> 
> Had a look over here which brought up my questions.
> 
> https://www.duetds.com/components/link/ https://github.com/carbon-design-system/carbon-web-components/blob/master/src/components/link/link.ts

Thanks for the in depth review @4aficiona2. I will add the id attribute. The rel attribute is set by the module itself for links which open in a new window. I'm not sure if we should expose the rel attribute to the consumers to populate it by themselves. I see very few cases, where this could make sense, but then again we could also solve this differently (e.g. allow/disallow noreferrer). Since we strictly separate the markup for the button and the link, the module should never receive another role as its html default role. Or do I overlook something here?

src/components/lyne-link/styles/lyne-link.component.scss
```diff
@@ -0,0 +1,70 @@
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin, available for shared
```
feerglas, line 2: I did not look into it any further, but the storybook plugin for switching direction ltr <> rtl is not working. But the semantic of your component is same as with the button, and there it is working (flex element with 2 spans in it). Did you notice it too? any idea how we can fix it?

src/components/lyne-link/styles/lyne-link.component.scss
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.link[dir="rtl"].link--icon-flip .link__text_icon {
+  transform: scaleX(-1);
+}
+
+.link__text_icon slot {
+  fill: currentColor;
+
+  @include ifForcedColors {
```
feerglas, line 55: wouldn't it make more sense to include it on the link color itself? the slot's fill will be inhertied then via currentColor, right?

src/components/lyne-link/lyne-link.stories.js
```diff
@@ -0,0 +1,185 @@
...
+    return `background-color: ${ColorWhiteDefault};`;
+  }
+
+  return `background-color: ${ColorCharcoalDefault};`;
+
+};
+
+const Template = (args) => (
+  <lyne-link {...args}>
+    <span slot='icon'>{getMarkupForSvg(args.icon)}</span>
```
feerglas, line 21: Mini-improvement: this way, the slot is still rendered if no icon is configured via attribute. If you change it to this, we get rid of the empty slot:
```
{args.icon &&
      <span slot='icon'>{getMarkupForSvg(args.icon)}</span>
    }
```

src/components/lyne-link/lyne-link.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  public render(): JSX.Element {
+
+    const textSizeClass = ` link--text-${this.textSize}`;
+    const currentLanguage = getDocumentLang();
+    const currentWritingMode = getDocumentWritingMode();
+
+    let openInNewWindow = false;
+
+    if (!this.hrefValue.includes('sbb.ch')) {
```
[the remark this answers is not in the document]
feerglas: Same opinion as @4aficiona2 . And what if other projects, that don't have sbb.ch in their url, want to use our components, like europaalle etc?

feerglas (changes requested):

src/components/lyne-link/lyne-link.tsx
```diff
@@ -0,0 +1,142 @@
...
+          `link
+          ${textSizeClass}
+          ${iconPositionClass}
+          ${iconFlipClass}
+          ${variantClass}`
+        }
+        download={this.download}
+        dir={currentWritingMode}
+        href={this.hrefValue}
+        id={this.idValue}
```
feerglas, line 127: i would handle this via `additionalLinkAttributes`, otherwise if no `id` is provided via attributes, we end up with an empty `id` attribute on the `a` tag

feerglas (changes requested):

PR #835, opened 2021-11-02
Feat/lyne link list

mbleuer:
Is based on the lyne link PR. The button variant of the link component is missing compared to the screendesign. Was unsure whether I should ad the links as Json Config, as slot or even as separate component which again would slot the link component. What do you think about the current implementation?


src/components/lyne-link-list/lyne-link-list.tsx
```diff
@@ -0,0 +1,93 @@
+import {
+  Component,
+  h,
+  Prop
+} from '@stencil/core';
+import { guid } from '../../global/guid';
+import { InterfaceLyneLinkListAttributes } from './lyne-link-list.custom.d';
+import { InterfaceTitleAttributes } from '../lyne-title/lyne-title.custom.d';
+
+/**
```
feerglas, line 10: you can remove this block, otherwise we get an "empty" doc entry

src/components/lyne-link-list/lyne-link-list.tsx
```diff
@@ -0,0 +1,93 @@
...
+    if (this.titleText === '') {
+      showTitle = false;
+    } else {
+      additionalAttributes = {
+        'aria-labelledby': id
+      };
+    }
+
+    return (
+      // the role="list" is needed for voice over: https://bit.ly/3CDiZaG
```
feerglas, line 69: to be aria compliant, shouldn't we have `role="listitem"` on the chhildren as well? or is ios smart enough for that? (just doubting, because it seems to have issues with the ul anyway...)

feerglas (changes requested):

src/components/lyne-link-list/lyne-link-list.tsx
mbleuer: I thought the same but in the referenced article they were not used. I have tested it, with or without role="listitem" and the results are the same.

PR #978, opened 2022-02-02
Feature/horizontal direction for link-list

4aficiona2:
\#\# Preflight Checklist
<\!\-\- Please ensure you've completed the following steps by replacing \[ \] with \[x\]\-\->

\* \[X\] I have read the \[Contributing Guidelines\]\(https://github.com/lyne\-design\-system/lyne\-components/blob/master/.github/CONTRIBUTING.md\) for this project.
\* \[X\] I agree to follow the \[Code of Conduct\]\(https://github.com/lyne\-design\-system/lyne\-components/blob/master/.github/CODE\_OF\_CONDUCT.md\) that this project adheres to.
\* \[X\] I have searched the \[pull request tracker\]\(https://github.com/lyne\-design\-system/lyne\-components/pulls\) for a Pull Request \(PR\) that matches the one I want to submit, without success.

\#\# Pull request checklist

\- \[X\] Tests for the changes have been added \(for bug fixes / features\)
\- \[X\] Docs have been reviewed and added / updated if needed \(for bug fixes / features\)
\- \[X\] Build \(\`npm run build\`\) was run locally and any changes were pushed
\- \[X\] Lint \(\`npm run lint\`\) has passed locally and any fixes were made for failures

\#\# Changes

Changes in this pull request:
- adds direction attribute for link-list component
- enables to switch between vertical (default) and horizontal representation

\-

\#\# Browsers

\- \[X\] Firefox Desktop
\- \[X\] Chrome Desktop
\- \[X\] Safari Desktop
\- \[X\] Firefox Mobile
\- \[ \] Chrome Mobile
\- \[ \] Safari Mobile
\- \[X\] Opera
\- \[X\] Brave

\#\# Pull request type

<\!\-\- Please do not submit updates to dependencies unless it fixes an issue. \-\->

<\!\-\- Please try to limit your pull request to one type, submit multiple pull requests if needed. \-\->

\- \[ \] Bugfix
\- \[X\] Feature
\- \[ \] Code style update \(formatting, renaming\)
\- \[ \] Refactoring \(no functional changes, no api changes\)
\- \[ \] Build related changes
\- \[ \] Documentation content changes
\- \[ \] Other \(please describe\):

\#\# Does this introduce a breaking change?

\- \[ \] Yes
\- \[X\] No

\#\# Mentions

<\!\-\- Please leave this mention \-\->

@lyne\-design\-system/core

<\!\-\- Add other mentions \-\->


src/components/lyne-link-list/styles/lyne-link-list.component.scss
```diff
@@ -15,6 +15,14 @@ lyne-title {
   @include listReset();
 }
 
+.link-list--horizontal {
```
mbleuer, line 18: This is a challenging one. I would recommend to remove the media query here. For the current use case you are working on, this makes sense but for other occasions we eventually would like to have a different behaviour. I would define this as a general rule for link-list--horizontal and define the explicit logic for the bottom footer part in a separate component. What do you think?

src/components/lyne-link-list/styles/lyne-link-list.component.scss
```diff
@@ -15,6 +15,14 @@ lyne-title {
...
+  @include mq($from: large) {
+    display: flex;
+    flex-wrap: wrap;
+    gap: pxToRem(var(--spacing-fixed-8x));
+  }
+}
+
 ::slotted(.link-list__item:not(:last-child)) {
```
mbleuer, line 26: For the horizontal variant, we would need to remove or unset the margin-block-end.

src/components/lyne-link-list/lyne-link-list.custom.d.ts
```diff
@@ -1,3 +1,4 @@
 export interface InterfaceLyneLinkListAttributes {
   variant?: 'positive' | 'negative';
```
mbleuer, line 2: To follow your recommendation, we could change **variant** to **appearance**. Then we could still decide whether vertical | horizontal are better named variant or direction. Direction is very specific and makes great sense for the current case while variant could offer more options for the future.

mbleuer (changes requested):

src/components.d.ts
```diff
@@ -443,6 +443,10 @@ export namespace Components {
         "variant": InterfaceLinkButtonAttributes['variant'];
     }
     interface LyneLinkList {
+        /**
```
mbleuer, line 446: Tiny detail, duplicated description.

src/components/lyne-link-list/lyne-link-list.custom.d.ts
4aficiona2: This would definitely be an option to rename the **variant** into **appearance** but this would also require to rename the `lyne-title` since the variant gets passed into `lyne-title`. I won't do the refactoring within this branch but we can definitely consider to refactor these components. 

WIth merging the direction property into the variant/appearance property I do not fully agree. Because if we would do so, we would loose the opportunity to combine the variant with the direction like we can at the moment. We would have to create four variants ... 

- positive-vertical
- positive-horizontal
- negative-vertical
- negative-horizontal

src/components/lyne-link-list/styles/lyne-link-list.component.scss
4aficiona2: I removed the breakpoint and will implement the direction swap within the footer context if this will be possible.

src/components/lyne-link-list/styles/lyne-link-list.component.scss
4aficiona2: Thanks for this one!

`margin-block-end` is unset for `.link-list--horizontal`

4aficiona2 (commented):
I made the suggested change of handling the breakpoint in the footer context.

4aficiona2:
I actually had to reintroduce the vertical/horizontal handling of the link list since it is not possible to handle this from outside (within the footer context) since we have double nested / slotted components (grand children) where we have absolute no control from outside. 

Due to this I introduced in the stack git branch a third direction variant `horizontal-from-large` which indicates that it will be horizontal from the large breakpoint on.

PR #1179, opened 2022-05-24
Feat/toast 1026

### LINKED ISSUE #1026: story(toast): initial implementation
### Feature Description

We want to provide a toast component.

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=2928%3A459049)

**References**
- https://material.angular.io/components/snack-bar/overview
- https://angular.app.sbb.ch/angular/components/notification-toast/overview

**Design Spec**
- Icon is optional and configurable
- Should have an optional close button/icon
- Should have an optional action, which can either be a button or a link
- The positioning of the toast should be configurable (defaults to bottom center)
- @mcilurzo checks max width
  - Message text should wrap if max width exceeded and icon and action should be positioned at the top
- Animation should use ease-in and ease-out to show/hide toast

**Technical Spec**
- Default timeout should be 3 seconds (configurable)
- Inside the `class LyneToast { static { ... } }` anchor the API to the global scope and also export the API
- Check how can the Toast be loaded (via import or standalone usage)

```ts
interface LyneToastConfiguration {
  /** Message to display. */
  message: string;
  /** Hide the toast after defined milliseconds. */
  timeout?: number = 3000;
  /** Either SVG string or reference to a SVG element. */
  icon?: string | HTMLElement;
  /** Id of <template> to use for the icon. */
  iconTemplate?: string;
  /** Action configuration. */
  action?: 'close' | { label: string; action: () => void; } | { label: string; link: string; };
  /** Where the toast should be displayed vertically. Defaults to 'bottom'. */
  verticalPosition?: 'top' | 'bottom' = 'bottom';
  /** Where the toast should be displayed horizontally. Defaults to 'center'. */
  horizontalPosition?: 'left' | 'center' | 'right' | 'start' | 'end' = 'center';
}

globab const LyneToastApi: {
  open(config: LyneToastConfiguration): LyneToastRef;
}
```

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

DavideMininni-Fincons:
## Preflight Checklist
* [x] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/master/.github/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/master/.github/CODE_OF_CONDUCT.md) that this project adheres to.
* [x] I have searched the [pull request tracker](https://github.com/lyne-design-system/lyne-components/pulls) for a Pull Request (PR) that matches the one I want to submit, without success.

## Issue

This PR Closes #1026

## Pull request checklist

## Changes

Changes in this pull request:

- Implementation of toast component

## Browsers

- [ ] Firefox Desktop
- [ ] Chrome Desktop
- [ ] Safari Desktop
- [ ] Firefox Mobile
- [ ] Chrome Mobile
- [ ] Safari Mobile
- [ ] Opera
- [ ] Brave

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Mentions

@lyne-design-system/core
@kyubisation 
@jeripeierSBB


src/components/lyne-toast/animations/toast.enter.ts
```diff
@@ -0,0 +1,29 @@
...
+  const baseAnimation = createAnimation();
+  const wrapperAnimation = createAnimation();
+
+  const root = baseEl.shadowRoot;
+  const wrapperEl = root.querySelector('.toast-wrapper') as HTMLElement;
+
+  wrapperAnimation.addElement(wrapperEl);
+
+  switch (position) {
+    case 'top':
```
jeripeierSBB, line 14: I don't see any difference to the default case?

src/components/lyne-toast/animations/toast.enter.ts
```diff
@@ -0,0 +1,29 @@
...
+      wrapperAnimation.fromTo('opacity', 0.01, 1);
+      wrapperAnimation.duration(750);
```
jeripeierSBB, line 16: durations should be taken from design tokens, but probably such a long duration does not exist right now, if it really should be that slow, we can ask @mcilurzo and add some more tokens in lyne-design-tokens

src/components/lyne-toast/animations/toast.enter.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    default:
+      wrapperAnimation.fromTo('opacity', 0.01, 1);
+      wrapperAnimation.duration(750);
+      break;
+  }
+
+  return baseAnimation
+    .easing('cubic-bezier(.36,.66,.04,1)')
+    .duration(1500)
```
jeripeierSBB, line 27: same comment for durations

src/components/lyne-toast/lyne-toast.custom.d.ts
```diff
@@ -0,0 +1,75 @@
+import { AnimationBuilder } from '../../global/core/components/animations/animation-interface';
+
+export interface InterfaceLyneToastAttributes {
+  verticalPosition: 'top' | 'bottom';
+  horizontalPosition: 'left' | 'center' | 'right' | 'start' | 'end';
+  politeness: 'off' | 'assertive' | 'polite';
```
jeripeierSBB, line 6: in lyne-alert I named it "ariaLivePoliteness" to explicitly mention the aria context. what are your thoughts?

src/components/lyne-toast/lyne-toast.custom.d.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type InterfaceToastType = 'link' | 'action' | 'icon';
```
jeripeierSBB, line 9: should that no also be exported?

src/components/lyne-toast/lyne-toast.custom.d.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+export interface InterfaceToastConfiguration {
+  message: string;
+  timeout?: number;
+  icon?: string | HTMLElement;
+  iconTemplate?: string;
+  action?: InterfaceToastLink | InterfaceToastAction | InterfaceToastIcon;
+  verticalPosition?: InterfaceLyneToastAttributes['verticalPosition'];
+  horizontalPosition?: InterfaceLyneToastAttributes['horizontalPosition'];
+  politeness?: InterfaceLyneToastAttributes['politeness'];
```
jeripeierSBB, line 65: same comment as above about aria-live prefix

jeripeierSBB (commented):
just some little comments about the first files in the PR :) I will need more time to get through all the PR.

src/components/lyne-toast/animations/toast.enter.ts
DavideMininni-Fincons: Resolved using variables '--animation-easing', '--animation-duration-6x' and '--animation-duration-3x'.

Added a FIXME:
@mcilurzo could be possible remove the 'ms' suffix for '--animation-duration-n' values?

src/components/lyne-toast/lyne-toast.custom.d.ts
DavideMininni-Fincons: In my opinion, InterfaceToastCommonAction is extended by InterfaceToast (Link / Action / Icon) so it's not useful on its own for consumers. 
The same for InterfaceToastType; it only serves to define the type property in InterfaceToast (Link / Action / Icon).

But on the other hand, I don't really know how the compilation happens: without 'export', are the types defined anyway (ie, does the consumer see the correct types)? if not, I add the export..

src/components/lyne-toast/animations/toast.enter.ts
jeripeierSBB: Opened a discussion in lyne teams chat.

Instead of parsing the css variables here it's also possible to directly import the tokens in javascript (but still a string with unit): 

`import {
  AnimationDuration6x
} from 'lyne-design-tokens/dist/js/tokens.es6';

jeripeierSBB: Importing javascript would be better in terms of performance for a consumer, but less flexible when a variable should change at runtime. This would probably not be a problem here as we don't expect overridden animation times during runtime(?). I tend to use the solution you chose because we are sure we can react to runtime changes

src/components/lyne-toast/lyne-toast.custom.d.ts
jeripeierSBB: @kyubisation Can you maybe answer as you know more in this topic?

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+  }
+  &.toast-left {
+    left: pxToRem(var(--spacing-responsive-s-zero));
+    transform: none;
+    @include mq($from: medium) {
+      left: pxToRem(var(--spacing-responsive-s-medium));
+    }
+  }
+  &.toast-right {
+    right: pxToRem(var(--spacing-responsive-s-zero));
```
jeripeierSBB, line 65:
```suggestion
    right: pxToRem(var(--spacing-responsive-s));
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+    left: auto;
+    transform: none;
+    @include mq($from: medium) {
+      right: pxToRem(var(--spacing-responsive-s-medium));
+    }
+  }
+
+  background: var(--color-midnight-default);
+  @include shadow-level-5-hard;
+  border-radius: pxToRem(var(--spacing-fixed-2x));
```
jeripeierSBB, line 75:
```suggestion
  border-radius: pxToRem(var(--border-radius-4x));
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+    & > button {
+      color: var(--color-default-text);
+      cursor: pointer;
+      background-color: transparent;
+      border: 0;
+      padding: 0;
+      span {
+        display: block;
+        line-height: 1;
+        svg {
```
jeripeierSBB, line 119: as mentioned in the beginning of the document, please avoid nesting as far as possible (maybe it's not here, I don't know ;) )

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+      //line-height: var(--typo-line-height-body-text);
+      text-decoration: none;
+    }
+    & > button {
+      color: var(--color-default-text);
+      cursor: pointer;
+      background-color: transparent;
+      border: 0;
+      padding: 0;
+      span {
```
jeripeierSBB, line 116: class name available?

jeripeierSBB (changes requested):
This review only includes comments for the styles. Sorry, at the end I lost a little bit the overview. After you made the requested changes, I will have a second look at it.

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+}
+
+.toast-wrapper {
+  // temporary -2 rem for storybook only
+  width: calc(100vw - 2rem);
+  height: calc(100vh - 2rem);
+  contain: content;
+}
+.toast {
+  &.toast-item--visible {
```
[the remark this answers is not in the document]
DavideMininni-Fincons: Class removed because unused.

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+    right: pxToRem(var(--spacing-responsive-s-zero));
+    left: auto;
+    transform: none;
+    @include mq($from: medium) {
+      right: pxToRem(var(--spacing-responsive-s-medium));
+    }
+  }
+
+  background: var(--color-midnight-default);
+  @include shadow-level-5-hard;
```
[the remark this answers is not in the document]
DavideMininni-Fincons: Sure, includes are on top and properties are better rearranged.

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+  flex-direction: row;
+  align-items: start;
+  gap: pxToRem(var(--spacing-fixed-4x));
+  span.toast-text {
+    color: var(--color-default-text);
+    flex-grow: 1;
+  }
+  .toast-icon {
+    color: var(--color-default-text);
+    flex: 0 0 19px; //width
```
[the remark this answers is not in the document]
DavideMininni-Fincons: For this one, Federico and I decided to keep it as it is, and add a FIXME waiting for the new sbb-icon.

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+
+:host {
+  .overlay-hidden {
+    display: none;
+  }
+
+}
+
+.toast-wrapper {
+  // temporary -2 rem for storybook only
```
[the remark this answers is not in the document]
DavideMininni-Fincons: We found out that we missed some props, removed because useless.

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+  position: fixed;
+
+  @include text-xs--bold;
+  --color-default-text: var(--color-white-default);
+  padding-block: pxToRem(var(--spacing-responsive-xxxs-zero));
+  padding-inline: pxToRem(var(--spacing-responsive-xs-zero));
+  width: calc(100% - (2 * pxToRem(var(--spacing-responsive-s-zero))));
+  margin: pxToRem(var(--spacing-responsive-s-zero)) 0;
+  @include mq($from: small) {
+    width: 400px;
```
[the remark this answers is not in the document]
DavideMininni-Fincons: 400 is defined in Figma under "Edge cases - max width" on the left side. 
Rewritten as `--toast-max-width: #{pxToRem(400)};` in :host

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+    line-height: var(--typo-line-height-body-text);
+    margin-left: auto;
+    cursor: pointer;
+    & > a {
+      color: var(--color-default-text);
+      //line-height: var(--typo-line-height-body-text);
+      text-decoration: none;
+    }
+    & > button {
+      color: var(--color-default-text);
```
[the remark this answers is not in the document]
federicoisepponfincons: we need that because button element resets some css instructions (color is one of them)

src/components/lyne-toast/styles/lyne-toast.component.scss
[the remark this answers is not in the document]
federicoisepponfincons: yes, to use flex alignment we would need at least one more wrapper; is that much better instead of absolute positioning?

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+  }
+  flex-direction: column;
+  align-items: flex-start;
+  position: fixed;
+
+  @include text-xs--bold;
+  --color-default-text: var(--color-white-default);
+  padding-block: pxToRem(var(--spacing-responsive-xxxs-zero));
+  padding-inline: pxToRem(var(--spacing-responsive-xs-zero));
+  width: calc(100% - (2 * pxToRem(var(--spacing-responsive-s-zero))));
```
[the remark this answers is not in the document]
federicoisepponfincons: yes, this is needed in mobile breakpoints ( below "medium" ) to have the full width

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,126 @@
...
+    color: var(--color-default-text);
+    flex: 0 0 19px; //width
+    height: 19px;
+    align-self: start;
+    fill: currentColor;
+    margin-right: pxToRem(var(--spacing-fixed-2x));
+    @include mq($from: medium) {
+      flex: 0 0 23.8px; //width
+      height: 23.8px;
+    }
```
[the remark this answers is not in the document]
federicoisepponfincons: same as the previous, temporary instructions, waiting for sbb-icon component

src/components/lyne-toast/styles/lyne-toast.component.scss
federicoisepponfincons: reduced nesting, where possible :)

src/components/lyne-toast/animations/toast.enter.ts
jeripeierSBB: I like it, thank you :)

src/components/lyne-toast/styles/lyne-toast.component.scss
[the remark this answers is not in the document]
jeripeierSBB: ok, but doesn't it generate wrong output styles, because it should be 24 and not 19?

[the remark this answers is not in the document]
federicoisepponfincons: this is because the example icon (on the left part of the toast) in Figma is 19x19... this is temporary, anyway, until we have the icon component

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+.toast {
+  @include text-s--bold;
+  @include shadow-level-5-hard;
+  @include absoluteCenterX;
+
+  position: fixed;
+  padding-block: var(--toast-padding-block);
+  padding-inline: var(--toast-padding-inline);
+  margin: var(--toast-margin) 0;
+  width: calc(100% - (2 * pxToRem(var(--spacing-responsive-s-zero))));
```
jeripeierSBB, line 40:
```suggestion
  width: calc(100% - (2 * var(--toast-margin)));
```
Removing '-zero' doesn't make any difference, but looks cleaner for me because it's specified as spacing s

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin, available for shared
+// and default mode
+@include host-component-properties;
+
+:host {
+  --color-default-text: var(--color-white-default);
```
jeripeierSBB, line 7:
```suggestion
  --toast-text-color: var(--color-white-default);
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+
+.toast {
+  @include text-s--bold;
+  @include shadow-level-5-hard;
+  @include absoluteCenterX;
+
+  position: fixed;
+  padding-block: var(--toast-padding-block);
+  padding-inline: var(--toast-padding-inline);
+  margin: var(--toast-margin) 0;
```
jeripeierSBB, line 39:
```suggestion
  margin-block: var(--toast-margin);
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+    transform: none;
+  }
+}
+
+.toast-text {
+  flex-grow: 1;
+}
+
+// FIXME temporary, waiting for sbb-icon
+.toast-icon {
```
jeripeierSBB, line 77: as mentioned in call: spacing not correct and value should probably be 24, respectively its css var.

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+  flex: 0 0 19px;
+  height: 19px;
```
jeripeierSBB, line 79: maybe `line-height: 0` helps here

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+}
+
+.toast-link {
+  color: var(--color-default-text);
+  text-decoration: none;
+}
+
+.toast-button {
+  @include buttonReset;
+  color: var(--color-default-text);
```
jeripeierSBB, line 102:
```suggestion
  color: var(--toast-text-color);
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+  flex-direction: row;
+  align-items: start;
+  bottom: 0;
+  gap: var(--toast-gap);
+  background: var(--toast-background-color);
+  color: var(--color-default-text);
+  border-radius: var(--toast-border-radius);
+
+  &.toast-top {
+    top: 0;
```
jeripeierSBB, line 56:
```suggestion
    inset-block-start: 0;
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+    bottom: auto;
+  }
+
+  &.toast-left {
+    left: var(--toast-margin);
+    transform: none;
+  }
+
+  &.toast-right {
+    left: auto;
```
jeripeierSBB, line 66:
```suggestion
    inset-inline-start: auto;
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+    right: var(--toast-margin);
```
jeripeierSBB, line 67:
```suggestion
    inset-inline-end: var(--toast-margin);
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
...
+:host {
+  --color-default-text: var(--color-white-default);
+  --toast-background-color: var(--color-midnight-default);
+  --toast-border-radius: #{pxToRem(var(--border-radius-4x))};
+  --toast-gap: #{pxToRem(var(--spacing-fixed-4x))};
+  --toast-padding-block: #{pxToRem(var(--spacing-responsive-xxxs))};
+  --toast-padding-inline: #{pxToRem(var(--spacing-responsive-xs))};
+  --toast-margin: #{pxToRem(var(--spacing-responsive-s))};
+  --toast-max-width: #{pxToRem(400)};
+  --toast-width: #{calc(100% - (2 * pxToRem(var(--spacing-responsive-s-zero))))};
```
jeripeierSBB, line 15:
```suggestion
```

src/components/lyne-toast/styles/lyne-toast.component.scss
```diff
@@ -0,0 +1,113 @@
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin, available for shared
+// and default mode
+@include host-component-properties;
+
+:host {
+  --color-default-text: var(--color-white-default);
+  --toast-background-color: var(--color-midnight-default);
+  --toast-border-radius: #{pxToRem(var(--border-radius-4x))};
```
jeripeierSBB, line 9: I suggest to move pxToRem to the usages (as in other components). e.g. if you make a further calculation, it won't create a lot of outputs.

jeripeierSBB (commented):
Very nice work of your css refactoring, thank you, I like the clean file now :)

I made another small suggestions and comments.

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,330 @@
...
+  }
+
+  /**
+   * Returns a promise that resolves when the toast will dismiss.
+   */
+  @Method()
+  public onWillDismiss<T = any>(): Promise<InterfaceOverlayEventDetail<T>> {
+    return eventMethod(this.el, 'lyne-toast_will-dismiss');
+  }
+
```
jeripeierSBB, line 183: Hm I don't see the reason why those methods are used? Is it only because of the interface? what are de differences to didDismiss and WillDismiss?

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,330 @@
...
+   * Creates the CSS classes for the action button.
+   * @returns A CssClassMap with two default CSS classes and all the CSS classes
+   * added by consumers in _internalConfig.action.cssClass.
+   */
+  private _buttonClass(): CssClassMap {
+    const fn: (acc: CssClassMap, curr: string) => CssClassMap = (acc: CssClassMap, curr: string) => {
+      acc[curr] = true;
+
+      return acc;
+    };
```
jeripeierSBB, line 208: Can't we simplify that to Object.fromEntries ?

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,330 @@
...
+      'toast-button': true,
+      ...classArray.reduce(fn, {})
+    };
+  }
+
+  /**
+   * Triggers the action handler and dismiss the toast.
+   * @param action The action provided bu consumers in the toast config.
+   */
+  private _handleButtonClick(action: InterfaceToastAction): Promise<boolean> {
```
jeripeierSBB, line 223:
```suggestion
  private _handleActionButtonClick(action: InterfaceToastAction): Promise<boolean> {
```

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,330 @@
...
+          : 'right';
+      }
+      default: {
+        return this._internalConfig.horizontalPosition;
+      }
+    }
+  }
+
+  public render(): JSX.Element {
+    const zIndex = this.overlayIndex + 60000;
```
jeripeierSBB, line 288: We probably have to document somehow why 60000 is chosen, or have a centralized file where the z-indexes for the overlays are stored?

jeripeierSBB (commented):
Another partial review :)

src/components/lyne-toast/lyne-toast.custom.d.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+type InterfaceToastType = 'link' | 'action' | 'icon';
+
+/**
+ * Common interface for toast action.
+ */
+interface InterfaceToastCommonAction {
+  type: InterfaceToastType;
+  role: 'cancel' | string;
```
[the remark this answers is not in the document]
DavideMininni-Fincons: You are totally right; 'cancel' is not present in the MDN role list. 
However, Ionic uses it (I think) for two reason (see usages of https://github.com/ionic-team/ionic-framework/blob/1641dbc52a07567953bc2f07c4bc828ed95a5185/core/src/utils/overlays.ts#L570):
- since they could have multiple button in alert/toast/etc, it's used to connect the right handler to the button with role 'cancel';
- when the alert/toast/ecc is dismissed via cancel button, the role 'cancel' is emitted in the will/didDismiss events, so maybe consumers can bind their own logic (eg. toast autoclose emits a 'timeout' value instead; see also https://62a9b9a402dd4701997253a8--lyne-components-storybook.netlify.app/?path=/story/components-overlay-lyne-toast--icon-and-close-icon-action-with-event-listeners ).

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+  /**
+   * Renders the action/icon button.
+   * @param onClickFn Callback on button click.
+   * @param innerToastEl The HTML rendered as button content.
+   * @returns The button that will be rendered as action/icon.
+   */
+  private _renderActionCommonButton(onClickFn: () => Promise<boolean>, innerToastEl: JSX.Element): JSX.Element {
+    return (
+      <button type='button' tabIndex={0} onClick={onClickFn} role={this._internalConfig.action.role} class={this._buttonClass()}>
```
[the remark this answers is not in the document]
DavideMininni-Fincons: Hmm.. just checked, useless, maybe a typo from an old implementation? it will be removed. 
A check for action button focus is underway with @federicoisepponfincons .

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,332 @@
...
+
+  /**
+   * Renders the action button (link/action/close icon).
+   */
+  private _renderAction(): JSX.Element {
+
+    switch (this._internalConfig.action.type) {
+      case 'link': {
+        return (
+          <a class='toast-link lyne-focusable' href={this._internalConfig.action.href} target='_blank' rel='noreferrer' tabIndex={0} role={this._internalConfig.action.role} onClick={this.dismiss.bind(this, null, 'link')}>
```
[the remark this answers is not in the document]
DavideMininni-Fincons: It will be removed.

jeripeierSBB:
The Toast Figma specification has changed in a way that lyne-button and lyne-link should be used. The variant in lyne link does not yet exist, I'm going to create it today because I will need it in alert too.

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,330 @@
...
+   * Emitted before the toast has presented.
+   */
+  @Event({
+    eventName: 'lyne-toast_will-present'
+  }) public willPresent: EventEmitter<void>;
+
+  /**
+   * Internal timeout.
+   */
+  private _durationTimeout: NodeJS.Timeout;
```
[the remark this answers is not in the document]
jeripeierSBB: Yes if it works :-)

src/components/lyne-toast/lyne-toast.custom.d.ts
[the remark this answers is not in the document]
jeripeierSBB: Ok, thanks for your clarification, so we leave it for now :-)

src/components/lyne-toast/lyne-toast.tsx
```diff
@@ -0,0 +1,330 @@
...
+  },
+  tag: 'lyne-toast'
+})
+
+export class LyneToast implements ComponentInterface, InterfaceOverlay {
+
+  /**
+   * If `true`, the toast will animate.
+   */
+  public animated = false;
```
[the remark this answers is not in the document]
DavideMininni-Fincons: The flag is used to disable animation (now it's not working due a bug, so thank you for your comments, I'm fixing it and maybe I will add a new story for the no-animation case).
Yes, I could rename it and invert the logic, will be changed in the next commit.

src/components/lyne-toast/lyne-toast.tsx
DavideMininni-Fincons: I think I didn't get it.. 
Basically, I want to convert the classArray into an object like
` {element1OfTheArray: true, element2OfTheArray: true, ...}`
Maybe I can do something like this:

```
const obj = {}
classArray.forEach(e => obj[e] = true);
return { 'lyne-focusable': true, 'toast-button': true, ...obj}
```

but I think the reduce fn does pretty the same..

src/components/lyne-toast/lyne-toast.tsx
DavideMininni-Fincons: They came from Ionic: from what I could see, onDidDismiss is mandatory because it's used in the focusPreviousElementOnDismiss method in overlay.tsx (funny thing: in Ionic it's not defined in OverlayInterface, so if you forget it in an overlay component, the focusPreviousElementOnDismiss method breaks); onWillDismiss appears not to be used explicitly.

For both, I assume they are defined because they might be useful in framework implementations, such as Angular, instead of attaching a listener to the event, something like:
`<my-overlay-comp [config] = 'config' (onWillDismiss)="onWillDismissHandler($event)"> </my-overlay-comp>`
I took a look at the Ionic codebase but I'm not 100% sure about it.

src/components/lyne-toast/lyne-toast.stories.js
```diff
@@ -0,0 +1,401 @@
+import getMarkupForSvg from '../../global/helpers/get-markup-for-svg';
```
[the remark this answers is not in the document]
DavideMininni-Fincons: A new story has been added with only the template, no interaction (open/close).

src/components/lyne-toast/lyne-toast.tsx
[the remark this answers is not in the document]
DavideMininni-Fincons: Bug fixed, story with no animation added.

src/components/lyne-toast/lyne-toast.tsx
DavideMininni-Fincons: Added new file: src/global/z-index-list.ts

src/components/sbb-toast/sbb-toast.scss
```diff
@@ -0,0 +1,140 @@
...
+    );
+  }
+
+  .toast-spacer {
+    display: none;
+  }
+}
+
+.toast-text {
+  word-break: break-all;
```
dauriamarco, line 92: This layout on small screens does not look right. The message text should wrap to the words without breaking the letters, and the button maybe should have enough space not to break (?)



```suggestion
  word-break: break-word;
```

dauriamarco (commented):
Hi 👋 thank you for your work! I just left an initial comment on a particular visualization, I will be continuing the review in the next few days.

src/components/sbb-toast/sbb-toast.scss
DavideMininni-Fincons: Text fixed & pushed;
about the button, I changed the word-break prop for sbb-link in commit cbc9d3b0529c56ebb7aa1159e3d4bfa88164109d, but then I reverted it because OFC it was not a solution..

DavideMininni-Fincons:
@kyubisation @jeripeierSBB 
Some infos about accessibility: using aria-live="assertive" (or "polite"), the toast is not announced using JAWS/NVDA; same thing happens in Ionic (see bug https://github.com/ionic-team/ionic-framework/issues/25866).

In the overlay system as now developed, opening a toast means adding a new `sbb-toast` element directly to the DOM, unlike what happens eg. in Material (if I remember correctly), where there is always a 'container overlay' in the body and components are added to this container.
According to [this comment](https://github.com/nvaccess/nvda/issues/8873#issuecomment-432487794), this behavior could be the cause of the problem: screenreaders announce messages when something changes **inside** a region, but not when the region itself is added to the DOM.

As a partial fix, I tried [this solution](https://github.com/lyne-design-system/lyne-components/commit/07859b78f28ec41bd4d375eae20e082dca28d559) and it works, toast is announced, but it's a simple JS solution. If we decided to go with this/something similar to this, we have to revise the overlay system.

kyubisation:
@DavideMininni-Fincons Thanks for the investigation. It's good to gather some experiences with the overlay system. I'm still contemplating going with `<dialog>` in the long run, but I would like to have some more experience before we do a final decision.

PR #1272, opened 2022-06-24
fix: update design tokens and adapt imports

kyubisation:
Update `@sbb-esta/lyne-design-tokens` to version 0.1.5, which allows imports directly from the package root.


DavideMininni-Fincons (commented):
Looks good 👍

PR #1277, opened 2022-06-28
build: add stylelint

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
Should the stylelint packages be grouped in renovate?

PR #1281, opened 2022-07-01
refactor: remove scss standalone mode

kyubisation:
Previously there were two modes: default (standalone/encapsulated) and
shared. This PR removes the encapsulated mode as we currently have no
need and expect consumers to use the global stylesheet, which populates
the necessary CSS variables.


DavideMininni-Fincons (commented):
It's ok for me, thanks :)

PR #1287, opened 2022-07-08
docs: mark all unfinished components accordingly in storybook

kyubisation:
To give an indication of the state of the components, this PR marks everything that is currently not finished accordingly in the storybook. This allows us in the future, once we complete a component, to remove the corresponding remark and indicate production readiness for a specific component.


dauriamarco:
LGTM 👌

PR #1292, opened 2022-07-14
feat(sbb-link): refactor sbb-link according to new specification

### LINKED ISSUE #1239: story(sbb-link): review/refactor
### Feature Description
We want to review/refactor the lyne-button implementation.

[Figma Spec] (https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=615%3A530767)

- dev: https://lyne.sbb.ch/components/lyne-link

**References**

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

**Design Spec**
- See Figma spec

**Technical Spec**
- Slot instead of label
- Should internally either use `<button>` or `<a>` (or `<span>` with visual only)
  - Should render an `<a>` element when `href` is used, when `href` is not provided use `<button>` element
  - Should support all attributes from both `<button>` or `<a>` but only use the appropriate ones
  - Should add property `static` to indicate, that internally the `<span>` should be used (To discuss: Discussion about ancester detection)
- Error should not short circuit
- Internally check if icon is used and use named slot for icons
- Refactor negative variant according to guidelines (add property negative: boolean separate from variant property)

```html
<sbb-link href="...">Action</sbb-link>
```

### Definition of Done
 - [x]  Component is implemented
 - [x]  Tests are implemented
 - [x]  E2E Tests are implemented
 - [x]  Storybook stories are implemented
 - [x]  Navigation via keyboard is tested
 - [x]  Screen reader output is tested
 - [x]  Remaining accessibility is tested
 - [x]  UX approved

HendrikExtSBB:
## Preflight Checklist

## Issue

This PR Closes #1239

## Pull request checklist

- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Changes

Changes in this pull request:
- change link text from property to slot
- add button element condition if no href is present and add corresponding attribute support
- add static-span support

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Safari Desktop
- [x] Firefox Mobile
- [x] Chrome Mobile
- [ ] Safari Mobile
- [ ] Opera
- [ ] Brave

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] Other (please describe):

- [x] Yes
- [ ] No

- the text is now set via slot (default slot)
- the aria-label has its own property instead of the former text property


src/components/sbb-link/sbb-link.tsx
```diff
@@ -67,25 +73,38 @@ export class SbbLink {
...
+
+  /**
+   * Name attribute if link is used as button
+   */
+  @Prop() public buttonName?: string;
+
+  /**
+   * Form attribute if link is used as button
+   */
+  @Prop() public formId?: string;
```
kyubisation, line 89: We don't have a common pattern yet for internal ids. For form elements we agreed on `inputId`. Since this is not strictly speaking a form element, I would suggest to use `linkId` and use this id in all variants.
Also I would recommend initializing the id property. See e.g. https://github.com/lyne-design-system/lyne-components/blob/cc7e9a99253e64eaa3e6b479999b03d8f4fcc47f/src/components/sbb-title/sbb-title.tsx

src/components/sbb-link/sbb-link.tsx
```diff
@@ -46,14 +46,20 @@ export class SbbLink {
    */
   @Prop() public iconFlip?: boolean;
 
+  /**
+   * If this is set to true an span element will be used
+   * instead of a anchor or a button
+   */
+  @Prop() public staticSpan?: boolean = false;
```
kyubisation, line 53: I'd prefer to just use `static` or was there a problem with that?

src/components/sbb-link/sbb-link.tsx
```diff
@@ -67,25 +73,38 @@ export class SbbLink {
...
     return this.variant === 'inline' || this.variant === 'inline-negative';
   }
 
-  public render(): JSX.Element {
+  /**
+   * Get the constructed class string for the given config of the element
+   * @private
+   * @return <string>
+   */
+  private get _getClassString(): string {
```
kyubisation, line 105: Nitpick: I prefer methods to be placed after their usage

kyubisation (changes requested):
Thank you very much for your work!
I was not able to fully finish the review, but I commented some things I noticed.

src/components/sbb-link/sbb-link.tsx
sebastiencloss, line 53: if it is initialised per default ("=false"),  no need of the "?". It is anyway optional,because it has a default initalizer.

src/components/sbb-link/sbb-link.tsx
```diff
@@ -100,51 +119,138 @@ export class SbbLink {
...
+   */
+  private get _getAttributeList(): object {
+    let ariaLabelText = this.ariaText;
+    const currentLanguage = getDocumentLang();
+    const currentWritingMode = getDocumentWritingMode();
+
+    let attributeList = {};
+
+    attributeList = {
+      ...attributeList,
```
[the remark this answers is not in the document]
sebastiencloss: so to add myValue, we need to add the property..
so like "value[myProperty] = myValue"  ?

src/components/sbb-link/sbb-link.tsx
```diff
@@ -46,14 +46,20 @@ export class SbbLink {
    */
   @Prop() public iconFlip?: boolean;
 
+  /**
+   * If this is set to true an span element will be used
+   * instead of a anchor or a button
```
jeripeierSBB, line 51:
```suggestion
   * instead of an anchor or a button
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -67,25 +73,38 @@ export class SbbLink {
...
+
+  /**
+   * Form attribute if link is used as button
+   */
+  @Prop() public formId?: string;
+
+  /**
+   * Type attribute if link is used as button
+   */
+  @Prop() public buttonType?: InterfaceLinkAttributes['buttonType'];
```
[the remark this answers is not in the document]
HendrikExtSBB: Changed

src/components/sbb-link/sbb-link.tsx
```diff
@@ -67,25 +73,38 @@ export class SbbLink {
...
 
+  /**
+   * Disabled attribute if link is used as button
+   */
+  @Prop() public disabled?: boolean;
+
+  /**
+   * Name attribute if link is used as button
+   */
+  @Prop() public buttonName?: string;
```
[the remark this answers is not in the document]
HendrikExtSBB: Changed

src/components/sbb-link/sbb-link.tsx
HendrikExtSBB: Its a reserved keyword.

src/components/sbb-link/sbb-link.tsx
```diff
@@ -100,51 +119,138 @@ export class SbbLink {
...
+   */
+  public render(): JSX.Element {
+    let TAG_NAME;
+    if (!this.hrefValue && !this.staticSpan) {
+      TAG_NAME = 'button';
+    } else if (!this.hrefValue && this.staticSpan) {
+      TAG_NAME = 'span';
+    } else {
+      TAG_NAME = 'a';
+    }
```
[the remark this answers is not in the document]
HendrikExtSBB: My approach prevent delivering spans with href.

src/components/sbb-link/sbb-link.tsx
```diff
@@ -100,51 +119,138 @@ export class SbbLink {
...
+      if (this.disabled) {
+        attributeList = {
+          ...attributeList,
+          tabIndex: '-1',
+        };
+      }
+    }
+
+    // Button specific attributes
+    if (!this.hrefValue && !this.staticSpan) {
```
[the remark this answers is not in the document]
HendrikExtSBB: I like!

HendrikExtSBB:
DO NOT MERGE. I forgot to adjust the sbb-link-list, sorry. :D

sebastiencloss (approved):

Acika00mk (approved):

src/components/sbb-link/sbb-link.tsx
```diff
@@ -17,43 +17,56 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
...
    * If set to true, the browser will
-   * show the download dialog on click.
+   * show the download dialog on click (optional).
    */
   @Prop() public download?: boolean;
 
-  /** The href value you want to link to */
-  @Prop() public hrefValue!: string;
+  /** The href value you want to link to (mandatory otherwise the link becomes a button)*/
+  @Prop() public href!: string;
```
kyubisation, line 25: Nitpick: This is not generally mandatory, just for the case when it `should` render as an ancher element.

src/components/sbb-link/sbb-link.tsx
```diff
@@ -17,43 +17,56 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
...
    * the text.
    */
   @Prop() public iconPlacement: InterfaceLinkAttributes['iconPlacement'] = 'start';
 
-  /** The link text we want to visually show. */
-  @Prop() public text!: string;
+  /**
+   * The aria-label attribute text. (optional)
+   */
+  @Prop() public accessibilityLabel?: string;
```
kyubisation, line 64: Could you also add the other accessibility properties:
```ts
/** This will be forwarded as aria-describedby to the relevant nested element. */
@Prop() public accessibilityDescribedby?: string;
/** This will be forwarded as aria-label to the relevant nested element. */
@Prop() public accessibilityLabel?: string;
/** This will be forwarded as aria-labelledby to the relevant nested element. */
@Prop() public accessibilityLabelledby?: string;
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -63,88 +76,134 @@ export class SbbLink {
   @Prop() public textSize: InterfaceLinkAttributes['textSize'] = 's';
 
   /**
-   * Choose the link style variant.
+   * Applies link inline styles (underline, inherit coloring/font-size etc).
    */
-  @Prop() public variant: InterfaceLinkAttributes['variant'] = 'block';
+  @Prop() public inline = false;
```
kyubisation, line 81: This should still be variant (with `block` and `inline`)?

src/components/sbb-link/sbb-link.tsx
```diff
@@ -63,88 +76,134 @@ export class SbbLink {
...
+
+  /**
+   * Type attribute if link is used as button (optional)
+   */
+  @Prop() public type?: InterfaceLinkAttributes['buttonType'];
 
+  /**
+   * Render element
+   */
   public render(): JSX.Element {
```
kyubisation, line 106: As mentioned in chat; Can be placed at the bottom if you prefer it that way.

src/components/sbb-link/sbb-link.tsx
```diff
@@ -63,88 +76,134 @@ export class SbbLink {
...
+   */
+  private get _getAttributeList(): object {
+    let ariaLabelText = this.accessibilityLabel;
     const currentLanguage = getDocumentLang();
     const currentWritingMode = getDocumentWritingMode();
+    const attributeList: Record<string, string> = {};
+
+    Object.assign(attributeList, {
+      dir: currentWritingMode || undefined,
+      class: this._getClassString || undefined,
```
kyubisation, line 143:
```suggestion
      class: this._getClassString(),
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -63,88 +76,134 @@ export class SbbLink {
...
+   * @return <object>
+   */
+  private get _getAttributeList(): object {
+    let ariaLabelText = this.accessibilityLabel;
     const currentLanguage = getDocumentLang();
     const currentWritingMode = getDocumentWritingMode();
+    const attributeList: Record<string, string> = {};
+
+    Object.assign(attributeList, {
+      dir: currentWritingMode || undefined,
```
kyubisation, line 142: If you are certain that the value cannot evaluate to falsy, you can remove the `|| undefined`

src/components/sbb-link/sbb-link.tsx
```diff
@@ -63,88 +76,134 @@ export class SbbLink {
...
+        target: '_blank',
+        'aria-label': ariaLabelText || undefined,
+      });
     }
 
-    let iconFlipClass = '';
-
-    if (this.icon && this.iconFlip) {
-      iconFlipClass = ' sbb-link--icon-flip';
+    if (!this.isStatic) {
```
kyubisation, line 165: I'd prefer to avoid if nesting. Could we do an early return if it is static?

kyubisation (changes requested):
Thanks again for your work.
Some minor issues still, after these are fixed/addressed I will merge this.

src/components/sbb-link/sbb-link.scss
```diff
@@ -39,26 +39,26 @@
...
+.sbb-link--negative {
+  --sbb-focus-outline-color: var(--sbb-focus-outline-color-dark);
   --text-color-normal: var(--sbb-color-white-default);
   --text-color-hover-focus: var(--sbb-color-cloud-default);
   --text-color-active: var(--sbb-color-silver-default);
 }
 
 .sbb-link--inline,
-.sbb-link--inline-negative {
+.sbb-link--inline.sbb-link--negative {
```
jeripeierSBB, line 54: Is line 54 really needed? Without testing I think .sbb-link--inline is sufficient here.

src/components/sbb-link/sbb-link.tsx
jeripeierSBB: I'd prefer to use `variant: 'block' | 'inline'` instead of a boolean

src/components/sbb-link/sbb-link.tsx
```diff
@@ -17,43 +17,56 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
...
 
   /**
    * The icon name we want to use,
    * choose from the small icon variants from
    * the ui-icons category from here
-   * https://lyne.sbb.ch/tokens/icons/.
+   * https://lyne.sbb.ch/tokens/icons/ (optional).
    * Inline variant doesn't support icons.
    */
   @Prop() public icon?: string;
```
jeripeierSBB, line 34: At the moment there is a slot and a property for the icon. The property is only used to toggle showing the icon slot, so there is completely no matter what you write in the property. With the new sbb-icon component we can handle now the property and the slot. Please change the logic to use a sbb-icon inside the icon slot as a default and set the name attribute of it to the icon property. Whenever a consumer provides its own icon slot, it is overwritten and if not it takes the string from the icon property. Please also change the corresponding stories if necessary.

src/components/sbb-link/sbb-link.tsx
jeripeierSBB: @kyubisation Do we really want to change hrefValue into href? There are a lot of components which use hrefValue.

jeripeierSBB (commented):
@kyubisation Should the logic which recognizes if a link is used inside another link or button and then becomes static be part of this PR?

src/components/sbb-link/sbb-link.tsx
HendrikExtSBB: How do you guys want it now (@kyubisation)?^^ Sorry I don't get it fully from the comments.

jeripeierSBB: @HendrikExtSBB Sorry for the confusion, it should be "href" as @kyubisation previously defined and you already applied.

src/components/sbb-link/sbb-link.tsx
HendrikExtSBB: There is no if-nesting and this condition is related to the icon position. Not if the element appears as a span.

kyubisation: With if-nesting I mean there is an if inside an if.
This could be refactored to:
```ts
if (this.isStatic) {
  return attributeList;
}
```

HendrikExtSBB: You were at the wrong line. ;)  But found the pos that you mentioned.

src/components/sbb-link/sbb-link.scss
HendrikExtSBB: True words, thx! :)

src/components/sbb-link/sbb-link.tsx
```diff
@@ -98,53 +180,34 @@ export class SbbLink {
...
-        href={this.hrefValue}
-        {...additionalLinkAttributes}
-      >
-        {this.icon && !this._inlineVariant ? (
-          <span class="sbb-link__icon">
-            <slot name="icon" />
-          </span>
+      <TAG_NAME {...this._getAttributeList}>
+        {this.icon && this.variant !== 'inline' ? (
+          <sbb-icon name={this.icon}></sbb-icon>
```
jeripeierSBB, line 205: Also you can remove the check for "this.icon".
```suggestion
          <slot name='icon'>
            <sbb-icon name={this.icon}></sbb-icon>
          </slot>
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -98,53 +180,34 @@ export class SbbLink {
...
-        dir={currentWritingMode}
-        href={this.hrefValue}
-        {...additionalLinkAttributes}
-      >
-        {this.icon && !this._inlineVariant ? (
-          <span class="sbb-link__icon">
-            <slot name="icon" />
-          </span>
+      <TAG_NAME {...this._getAttributeList}>
+        {this.icon && this.variant !== 'inline' ? (
```
jeripeierSBB, line 204:
```suggestion
        {this.variant !== 'inline' && (
```
and the you can remove the part after the ':'

src/components/sbb-link/sbb-link.tsx
```diff
@@ -17,43 +17,51 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
 export class SbbLink {
```
jeripeierSBB, line 17:
```suggestion
export class SbbLink implements AccessibilityProperties {
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -63,33 +71,107 @@ export class SbbLink {
...
+  @Prop() public type?: InterfaceLinkAttributes['buttonType'];
+
+  /** This will be forwarded as aria-describedby to the relevant nested element. */
+  @Prop() public accessibilityDescribedby?: string;
+
+  /** This will be forwarded as aria-label to the relevant nested element. */
+  @Prop() public accessibilityLabel?: string;
+
+  /** This will be forwarded as aria-labelledby to the relevant nested element. */
+  @Prop() public accessibilityLabelledby?: string;
```
jeripeierSBB, line 105: Changes due to the AccessibilityProperties interface. Please rebase your PR if you don't find AccessibilityProperties interface.
```suggestion
  /** This will be forwarded as aria-label to the relevant nested element. */
  @Prop() public accessibilityLabel: string | undefined;

  /** This will be forwarded as aria-describedby to the relevant nested element. */
  @Prop() public accessibilityDescribedby: string | undefined;

  /** This will be forwarded as aria-labelledby to the relevant nested element. */
  @Prop() public accessibilityLabelledby: string | undefined;
```
```

File: src/components/sbb-link/sbb-link.scss [hunk 23]
```diff
@@ -81,15 +80,15 @@
   }
 }
 
-.sbb-link__icon {
+sbb-icon {
   width: var(--sbb-size-icon-ui-small);
   height: var(--sbb-size-icon-ui-small);
```
jeripeierSBB, line 85: Is setting width and height really needed? Icon should have the right size itself.

src/components/sbb-link/sbb-link.scss
```diff
@@ -81,15 +80,15 @@
...
 }
 
-.sbb-link--icon-placement-end .sbb-link__icon {
+.sbb-link--icon-placement-end sbb-icon {
   order: 2;
 }
 
-.sbb-link[dir='rtl'].sbb-link--icon-flip .sbb-link__icon {
+.sbb-link.sbb-link--icon-flip sbb-icon {
```
jeripeierSBB, line 92: As reading the iconFlip docs, I think it should only be flipped on rtl mode. Why did you remove the `[dir='rtl']` part?

src/components/sbb-link/sbb-link.tsx
```diff
@@ -17,43 +17,51 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
...
    * If set to true, the browser will
-   * show the download dialog on click.
+   * show the download dialog on click (optional).
    */
   @Prop() public download?: boolean;
 
-  /** The href value you want to link to */
-  @Prop() public hrefValue!: string;
+  /** The href value you want to link to (mandatory otherwise the link becomes a button)*/
+  @Prop() public href!: string;
```
jeripeierSBB, line 25: Should it not be optional now (if it is a button)?
```suggestion
  @Prop() public href?: string;
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -17,43 +17,51 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
...
   /**
    * The icon can either be place before or after
    * the text.
    */
   @Prop() public iconPlacement: InterfaceLinkAttributes['iconPlacement'] = 'start';
 
-  /** The link text we want to visually show. */
-  @Prop() public text!: string;
+  /**
+   * Negative coloring variante flag
```
jeripeierSBB, line 62:
```suggestion
   * Negative coloring variant flag
```

jeripeierSBB (changes requested):

src/components/sbb-link/sbb-link.scss
HendrikExtSBB: [dir='rtl'] is used to set the reading direction of text and should not be connected to to and icon that has no contextual  meaning.

jeripeierSBB: But for what do we need the icon flip property then? Or you have at least to update the docs there. 

Your changes made that you now can flip an icon regardless of the writing direction.

jeripeierSBB (commented):
General comment: I think there are more stories / usages of sbb-link inside lyne-components which have to be adapted to the new api.

src/components/sbb-link/sbb-link.scss
HendrikExtSBB: I did not write the docs. They do not make sense to me. If you are using Icons that has a contextual meaning it should get a description somehow via an attribute (alt/aria-label or so). I dont get the idea behind that. But to flip the orientation of the icon depending on the position sounds logical to me.

src/components/sbb-link/sbb-link.tsx
```diff
@@ -14,46 +15,53 @@ import { i18nTargetOpensInNewWindow } from '../../global/i18n';
...
    */
   @Prop() public idValue?: string;
 
   /**
    * Decide whether the icon should get flipped
-   * horizontally if the document writing mode
-   * is changed from ltr to rtl or vice versa.
+   * horizontally to the left or to the right.
    */
   @Prop() public iconFlip?: boolean;
```
jeripeierSBB, line 47: I think we have a misunderstanding regarding the iconFlip property. Can it be that you are mix the properties iconPlacement and iconFlip? Icon flip does mirror an icon which for me still only reall makes sense in rtl mode. But I am unsure if we even want to support this edge case in the sbb-link component. @kyubisation your opinion?

jeripeierSBB: @HendrikExtSBB  @kyubisation and I decided to remove the iconFlip property, so you can delete it and remove the corresponding css classes.

kyubisation (approved):
LGTM 👍 
Thank you for your work!
(FYI: I applied a few changes that I discussed yesterday with @jeripeierSBB )

PR #1297, opened 2022-07-18
feat(sbb-icon): add width and height CSS variables and refactor

dauriamarco:
## Preflight Checklist

## Issue

This PR Closes #

## Pull request checklist

- [x] Build (`npm run build`) was run locally and any changes were pushed
- [x] Lint (`npm run lint`) has passed locally and any fixes were made for failures

## Changes

Changes in this pull request:

- add width and height CSS vars in order to resize the icon
- fix Safari behaviour on initial rendering
- refactor ARIA attributes properties

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Safari Desktop
- [x] Firefox Mobile
- [x] Chrome Mobile
- [x] Safari Mobile
- [x] Opera
- [ ] Brave

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/sbb-icon/sbb-icon.scss
+++ b/src/components/sbb-icon/sbb-icon.scss
@@ -3,31 +3,33 @@
 @include host-component-properties;
 
 :host {
-  --sbb-icon-margin-inline: 0;
-  --sbb-icon-margin-block: 0;
-
   display: inline-block;
   line-height: 0;
-  margin-inline: var(--sbb-icon-margin-inline-override, var(--sbb-icon-margin-inline));
-  margin-block: var(--sbb-icon-margin-block-override, var(--sbb-icon-margin-block));
+  margin-inline: var(--sbb-icon-margin-inline-override);
+  margin-block: var(--sbb-icon-margin-block-override);
 }
 
-svg:not(.color-immutable) {
-  [fill]:not([fill='none']) {
-    fill: currentcolor;
-  }
+svg {
+  width: var(--sbb-icon-width-override);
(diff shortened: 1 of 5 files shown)


src/components/sbb-icon/sbb-icon.scss
```diff
@@ -3,31 +3,33 @@
...
-svg:not(.color-immutable) {
-  [fill]:not([fill='none']) {
-    fill: currentcolor;
-  }
-
-  [stroke]:not([stroke='none']) {
-    stroke: currentcolor;
-  }
+svg {
+  width: var(--sbb-icon-width-override);
```
kyubisation, line 13: I'm a bit torn on naming here. (Also applies to height.)
What do you think about `--sbb-icon-svg-width`?
My thought is; Since we don't define a value ourselves, it's not really an override.
What do you think?

kyubisation (commented):
Nice work!
See my comment about css variable naming.

src/components/sbb-icon/sbb-icon.scss
dauriamarco: Exactly, I would have raised the same question in the next refinement. I agree with you for the most part, then I thought maybe we wanted to have the same naming convention to express the concept of "overriding the default value/behaviour", which in this case would be: "the size is defined by the retrieved icon, but you can override it". But I would probably change it to `--sbb-icon-svg-width` as you suggest anyway...

kyubisation (approved):
LGTM 👍

PR #1304, opened 2022-07-19
docs: add context detection section to CODING_STANDARDS.md

kyubisation:


dauriamarco:
Thank you very much, this is a good alternative to the `:host-context` selector. I just had some thoughts: what if we want to check whether a component is nested in another component of the same type (e.g. nested `sbb-tab-group`)? Should we consider starting traversing from the parent element to exclude the current element, or provide an option for that?

kyubisation:
@dauriamarco I think in that case, as you suggested, we can just use `this.el.parentElement`.

PR #1317, opened 2022-07-27
fix(sbb-toggle-check): fix various issues

jeripeierSBB:
- Removes space around component
 - Adds windows high contrast support
 - Simplifies code

### DIFF
--- a/src/components/sbb-toggle-check/sbb-toggle-check.tsx
+++ b/src/components/sbb-toggle-check/sbb-toggle-check.tsx
@@ -44,7 +44,7 @@ export class SbbToggleCheck implements AccessibilityProperties {
   /** The aria-describedby prop for the hidden input. */
   @Prop() public accessibilityDescribedby: string | undefined;
 
-  /** Event for emiting whenever selection is changed. */
+  /** Emits whenever the selection has changed.  */
   @Event() public sbbChange: EventEmitter;
 
   @Watch('checked')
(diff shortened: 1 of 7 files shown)


src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -58,9 +58,10 @@ export class SbbToggleCheck implements AccessibilityProperties {
   public render(): JSX.Element {
     const disabled = this.disabled ? 'toggle-check--disabled' : '';
     const checked = this.checked ? 'toggle-check--checked' : '';
+    const labelPosition = `toggle-check--${this.labelPosition}`;
```
kyubisation, line 61: I'd prefer this to be directly used in the template string

src/global/mixins.scss
```diff
@@ -44,3 +44,8 @@
   top: 50%;
   transform: translateY(-50%);
 }
+
+@mixin zeroWidthSpace() {
```
kyubisation, line 48: Could you maybe add a little documentation as to why this is needed?
And should we maybe inlcude the ::before inside this mixin?

src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -78,19 +79,18 @@ export class SbbToggleCheck implements AccessibilityProperties {
...
-                <sbb-icon name={this.icon}></sbb-icon>
-              </slot>
-            ) : (
-              ''
-            )}
+        <span class={'toggle-check__container'}>
+          <slot />
+          <span class={`toggle-check__slider`}>
+            <span class="toggle-check__circle">
+              {this.checked && (
```
kyubisation, line 86: This is probably a nitpick; I've noticed that due to this being loaded on demand, the tick is only visible after a few milliseconds after setting it to checked. Could we maybe use display: none to hide the tick instead of removing it from the DOM entirely?

kyubisation (changes requested):
Thank you for the investigation and improvements!
Some thoughts and nitpicks from my side.

osminaz (approved):
Nice work! thank you :)

kyubisation (approved):
LGTM 👍

PR #1342, opened 2022-08-18
fix(sbb-form-field): windows high contrast mode

jeripeierSBB:


DavideMininni-Fincons (commented):
Looks good, thanks :)

Issue #946, opened 2022-01-20
Deprecated Stencil output target: dist-custom-elements-bundle


feerglas:
Stencil will deprecate the output target `dist-custom-elements-bundle`, which we heavily rely on our project and other projects like lyne-documentation, lyne-getting-started etc.

Stencil has open issues regarding this deprecation:
https://github.com/Esri/calcite-components/issues/3957
https://github.com/ionic-team/stencil/issues/3136

We should transition as soon as possible to the output target `dist-custom-elements`.

DavideMininni-Fincons:
Useful link:
- the issue in Stencil
https://github.com/ionic-team/stencil/issues/3136
- a open PR in Stencil for an export-all-components 
https://github.com/ionic-team/stencil/pull/3368
- a closed PR in Ionic; note that a custom script is used to export everything (see folder scripts/custom-elements)
https://github.com/ionic-team/ionic-framework/pull/22863
- a closed PR in Ionic to update the starter component
https://github.com/ionic-team/stencil-component-starter/pull/107

Some bug related to the new setting:
https://github.com/ionic-team/stencil/issues?q=is%3Aissue+is%3Aopen+dist-custom-element

jeripeierSBB:
Related to #1568

jeripeierSBB:
Closed by #1568

PR #1360, opened 2022-08-30
feat: add @sbb-esta/lyne-components-react for react consumers

kyubisation:
This PR adds an additional package `@sbb-esta/lyne-components-react` intended for react consumers.


.github/workflows/release.yml
```diff
@@ -66,6 +66,19 @@ jobs:
...
+      - name: 'Download artifacts'
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const fs = require('fs');
+            const version = require('./package.json').version;
+            const path = `${process.env.GITHUB_WORKSPACE}/react-library/package.json`;
+            const pkgJson = fs.readFileSync(path, 'utf8');
+            fs.writeFileSync(path, pkgJson.replace(/0.0.0-PLACEHOLDER/g, version), 'utf8');
+      - name: 'Release: Publish @sbb-esta/lyne-components'
```
jeripeierSBB, line 78:
```suggestion
      - name: 'Release: Publish @sbb-esta/lyne-components-react'
```

stencil.config.ts
```diff
@@ -19,7 +20,7 @@ export const config: Config = {
       type: 'dist',
     },
     {
-      type: 'dist-custom-elements-bundle',
+      type: 'dist-custom-elements',
```
jeripeierSBB, line 23: Can we so "easily" remove the "-bundle"? https://github.com/lyne-design-system/lyne-components/issues/946

stencil.config.ts
```diff
@@ -2,6 +2,7 @@ import { inlineSvg } from 'stencil-inline-svg';
 import jestConfig from './.jest.config.js';
 import { sass } from '@stencil/sass';
 import { Config } from '@stencil/core';
+import { reactOutputTarget as react } from '@stencil/react-output-target';
```
jeripeierSBB, line 5: seems not to be used?

jeripeierSBB (changes requested):

stencil.config.ts
kyubisation: I'm actually not sure.
@DavideMininni-Fincons If I remember correctly, quite a while ago you checked what the differences were. Do you remember any details?

stencil.config.ts
kyubisation: See line 38

stencil.config.ts
DavideMininni-Fincons: The issue about deprecating dist-custom-elements-bundle is still open. Link: https://github.com/ionic-team/stencil/issues/3136
It seems that there are still some bugs that Stencil team should fix.

kyubisation: I'll revert it for now.

jeripeierSBB (approved):
LGTM

PR #1367, opened 2022-09-01
refactor: refine link button properties interfaces

jeripeierSBB:
BREAKING CHANGE:
- The link does not anymore himself decides to set target="_blank". It's now up to the consumer to decide if a link should be opened in a new window.


src/global/interfaces/link-button-properties.ts
```diff
@@ -126,56 +115,88 @@ export function getLinkButtonBaseAttributeList(
...
+  const baseAttributeList = getLinkButtonBaseAttributeList(linkProperties);
+
+  if (!linkProperties.href) {
+    return baseAttributeList;
+  }
+
+  const linkAttributeList = Object.assign(baseAttributeList, {
+    href: linkProperties.href,
+    ...(linkProperties.download && { download: '' }),
+    ...(buttonProperties.disabled && { tabIndex: '-1' }),
```
kyubisation, line 148: This pattern is not necessary. Object.assign filters out undefined values.
```suggestion
    download: linkProperties.download ? '' : undefined,
    tabIndex: buttonProperties.disabled ? '-1' : undefined,
```

src/global/interfaces/link-button-properties.ts
```diff
@@ -126,56 +115,88 @@ export function getLinkButtonBaseAttributeList(
...
+  }
+
+  return Object.assign(linkAttributeList, {
+    rel: linkProperties.rel || 'external noopener nofollow',
+    target: linkProperties.target || '_blank',
+    ...(linkProperties.accessibilityLabel && {
+      'aria-label': `${linkProperties.accessibilityLabel}. ${
+        i18nTargetOpensInNewWindow[getDocumentLang()]
+      }`,
+    }),
```
kyubisation, line 162:
```suggestion
    'aria-label': linkProperties.accessibilityLabel
      ? `${linkProperties.accessibilityLabel}. ${i18nTargetOpensInNewWindow[getDocumentLang()]}`
      : i18nTargetOpensInNewWindow[getDocumentLang()],
```

src/global/interfaces/link-button-properties.ts
```diff
@@ -126,56 +115,88 @@ export function getLinkButtonBaseAttributeList(
...
+ */
+export function getButtonAttributeList(buttonProperties: ButtonProperties): Record<string, string> {
+  const baseAttributeList = getLinkButtonBaseAttributeList(buttonProperties);
+
+  return Object.assign(baseAttributeList, {
+    name: buttonProperties.name,
+    type: buttonProperties.type || 'button',
+    onClick: buttonProperties.emitButtonClick.bind(buttonProperties),
+    ...(buttonProperties.form && { form: buttonProperties.form }),
+    ...(buttonProperties.disabled && { disabled: 'true' }),
```
kyubisation, line 178:
```suggestion
    form: buttonProperties.form || undefined,
    disabled: buttonProperties.disabled ? 'true' : undefined,
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -160,21 +176,22 @@ export class SbbLink implements LinkButtonProperties, ComponentInterface {
 
   public render(): JSX.Element {
     let TAG_NAME: string;
-    let attributeList: object = getLinkButtonBaseAttributeList(
-      this.idValue,
-      this._getClassString(),
-      this
-    );
+    let attributeList: object = getLinkButtonBaseAttributeList(this);
```
kyubisation, line 179: Is it not possible to use `Record<string, string>` here?
```suggestion
    let attributeList: Record<string, string> = getLinkButtonBaseAttributeList(this);
```

src/components/sbb-link/sbb-link.tsx
```diff
@@ -160,21 +176,22 @@ export class SbbLink implements LinkButtonProperties, ComponentInterface {
...
       TAG_NAME = this.href ? 'a' : 'button';
-      attributeList = getLinkButtonAttributeList(this.idValue, this._getClassString(), this);
+      attributeList = getLinkButtonAttributeList(this);
     }
 
     // See https://github.com/ionic-team/stencil/issues/2703#issuecomment-1050943715 on why form attribute is set with `setAttribute`
     return (
-      <TAG_NAME {...attributeList} ref={(btn) => this.form && btn?.setAttribute('form', this.form)}>
+      <TAG_NAME
+        {...(this.idValue && { id: this.idValue })}
```
kyubisation, line 190: I'm not sure I like this usage. Could we not just use `id={this.idValue}`?
I assume jsx will validate this anyway.

src/global/interfaces/accessibility-properties.ts
```diff
@@ -11,3 +11,23 @@ export interface AccessibilityProperties {
...
 }
+
+/**
+ * Lists all attributes for the AccessibilityProperties interface
+ * @param accessibilityProps accessibility props
+ */
+export function getAccessibilityAttributeList(
+  accessibilityProps?: AccessibilityProperties | null
+): Record<string, string> {
+  return Object.assign({
```
kyubisation, line 22: With the current usage, Object.assign will do nothing.
```suggestion
  return Object.assign({}, {
```

src/components/sbb-link/sbb-link.tsx
kyubisation, line 179: Minor performance nitpick; Are we not iterating the accessibility properties twice with the calls to `getLinkButtonBaseAttributeList` and `getLinkButtonAttributeList`?

jeripeierSBB: Yes it's called twice. I don't know which solution would be the best to avoid. generally I like if it is possible to just call `getLinkButtonAttributeList` and you have all properties in it (also the base props). I thought the performance impact would be minimal as it is (also removed Object assign now in accessibility properties). Otherwise we could pass the base attributes into `getLinkButtonAttributeList` and make a switch there?

jeripeierSBB: Forget my answer, is much easier to solve on callers side :D

DavideMininni-Fincons (approved):
Looks good, thanks! 👍

kyubisation (approved):
LGTM 👍

PR #1376, opened 2022-09-14
fix: use correct jsx factory for react output target

kyubisation:
This PR should fix the output format of our react output, by using the correct jsx factory for react.


jeripeierSBB (approved):
LGTM

PR #1378, opened 2022-09-15
feat(sbb-menu): introduce sbb-menu component

### LINKED ISSUE #1310: story(sbb-menu): initial implementation
### Feature Description

We want to provide the menu component.

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=355%3A953891)

**References**

- Positioning: https://material.angular.io/cdk/overlay/api#FlexibleConnectedPositionStrategy

**Design Spec**
- With size zero-small, the menu should be displayed as a sheet with a backdrop
- With size medium-ultra, the menu should be displayed as an overlay (attached to the trigger) without a visible backdrop
- `<sbb-divider>` should have appropriate margins
- If no icon is provided for a menu item action the space for it should still be reserved
- The default padding is `3x` (`2x` is to be discussed with @mcilurzo, and is not part of this issue)
- The menu should have flexible width depending on its content (except in sheet mode where it is full width) with a max width of  320px
- If the menu does not fit on the screen the content should be scrollable
- Create proposal implemenation for animation and verify with @mcilurzo (e.g. slide in for sheet, unfold for dropdown)

**Technical Spec**

- Try to extract the positioning logic into its own class/functionality as we will need it for other components

- `sbb-menu`
  - Uses `<dialog>` internally
  - Content should be provided via unnamed slot
  - `@Prop() trigger: string | HTMLElement` should accept both a string (id of an element) or a HTML element
  - Should attach click handler to trigger and remove it, if the trigger is removed or changed
  - With size medium-ultra the dialog should be "attached" to the trigger with a distance of `2x`
    - The menu should calucate the optimal position to display the dialog depending on available space with the following priority: start/below, start/above, end/below, end/above (maybe use [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) )
    - Logic to do this should be activated/deactivated depending on viewport size medium-ultra/zero-small
  - `ESC` key press should close the menu
  - On a click on the backdrop should close the menu
  - Focus should be on the first focusable element
  - `<sbb-menu>` should define padding/margin for slotted elements (`0` for `<sbb-divider>`, `3x` for `<sbb-menu-action>` and `5x` for everything else)
  - Investigate gesture close possibility when swiping down on sheet variant. If not possible or too complicated to do, document investigation results.
  - Investigate whether to use the backdrop to close the dropdown menu or register a once event on the document if anything else was clicked or interacted with.
  - Provide `show()` and `close()` methods on the `<sbb-menu>`
  - Any click on an interactive element (e.g. button, a) inside the `<sbb-menu>` that bubbles to the container should close the menu

- `sbb-menu-action`
  - Implementation should be similar to `<sbb-button>` and should internally either render a button (no href input) or anchor (href input defined) element (no static variant here, as a consumer can use other elements for that purpose)
  - `@Prop() icon?: string` for providing an icon which is internally rendered as a `<sbb-icon>`. Is optional and should render nothing but still reserve space if no icon is defined.
  - Content should be provided via an unnamed slot
  - `@Prop() amount?: string` can optionally be provided and is displayed as white text in a red circle (the circle should be able to stretch into a [stadium](https://en.wikipedia.org/wiki/Stadium_(geometry)))

```html
<sbb-button id="menu-trigger">Menu</sbb-button>

<sbb-menu trigger="menu-trigger">
  <sbb-menu-action icon="eye-small" href="./view">View</sbb-menu-action>
  <sbb-menu-action icon="pen-small" href="./edit">Edit</sbb-menu-action>
  <sbb-menu-action icon="trash-small" onclick="delete()">Delete</sbb-menu-action>
  <sbb-divider />
  <sbb-menu-action icon="cross-small" onclick="close()">Cancel</sbb-menu-action>
</sbb-menu>
```

```html
<sbb-button id="menu-trigger">Menu</sbb-button>

<sbb-menu trigger="menu-trigger">
  <strong>User Name</strong>
  <span>UserId</span>
  <sbb-link href="/profile">Profile</sbb-link>
  <sbb-divider />
  <sbb-menu-action icon="person-small" href="/account">Account</sbb-menu-action>
  <sbb-menu-action icon="tickets-small" href="/tickets">Tickets</sbb-menu-action>
  <sbb-divider />
  <sbb-menu-action icon="cross-small" onclick="logOut()">Log out</sbb-menu-action>
</sbb-menu>
```

### Definition of Done

- [x] Documentation is provided
- [x] Tests are implemented
- [x] E2E Tests are implemented
- [x] Storybook stories are implemented
- [x] Navigation via keyboard is tested
- [x] Screen reader output is tested
- [x] High-contrast is tested
- [x] Remaining accessibility is tested
- [x] UX approved

dauriamarco:
## Preflight Checklist

## Issue

This PR Closes #1310

## Pull request checklist

- [x] Build (`yarn build`) was run locally and any changes were pushed
- [x] Lint (`yarn lint`) has passed locally and any fixes were made for failures

## Changes

Changes in this pull request:

- implement menu component

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [x] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/sbb-menu/readme.md
```diff
@@ -0,0 +1,87 @@
+The `sbb-menu` is a component that you can attach to any element (trigger) to display a context menu. The menu appears on trigger left click. On mobile, the menu is displayed as a sheet with a backdrop; on desktop it will be shown as a floating menu and will calucate the optimal position relative to the trigger element by evaluating the available space with the following priority: start/below, start/above, end/below, end/above.
```
DavideMininni-Fincons, line 1: Maybe split in more lines?

src/components/sbb-menu/sbb-menu.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        <sbb-divider>
+          <sbb-menu-action icon="cross-small">
+            Cancel
+          </sbb-menu-action>
+        </sbb-divider>
+      </sbb-menu>
+    `);
+  });
+
+  // it('opens the menu', async () => {
```
DavideMininni-Fincons, line 55: If can't be tested, could we remove it?

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,290 @@
...
+    newValue: string | HTMLElement,
+    oldValue: string | HTMLElement
+  ): void {
+    if (newValue !== oldValue) {
+      this._triggerController.abort();
+    }
+  }
+
+  // Returns true if it is a DOM element.
+  private static _isElement(o: string | HTMLElement): boolean {
```
DavideMininni-Fincons, line 151: If it could be reused, move it to an utility file.

DavideMininni-Fincons (commented):
Looks good, thank you. ;)
Some minor issues noticed.

dauriamarco:
### Additional notes

I would add to the PR some additional notes to consider regarding some points in the implementation:

- tests of the `sbb-menu` component are limited by the fact that the dialog element `HTMLDialogElement` is not supported by Jest, preventing basic methods such as `showModal` from being recognized;

- chromatic has some problems making snapshots correctly due to absolute positioning of the element and transition on opening, but setting the `inlineStories: false` property and suspending animations almost completely fixes it (sometimes it fails to capture the opened menu in time for some breakpoint).

src/components/sbb-menu-action/sbb-menu-action.tsx
```diff
@@ -0,0 +1,150 @@
+import { Component, Event, EventEmitter, h, JSX, Prop } from '@stencil/core';
```
DavideMininni-Fincons, line 1: Add type 'undefined' on props

src/components/sbb-menu/sbb-menu.scss
```diff
@@ -0,0 +1,163 @@
...
+      duration: var(--sbb-animation-duration-6x);
+      timing-function: ease;
+    }
+
+    &.sbb-menu--no-animation {
+      animation-duration: 0s;
+    }
+  }
+
+  &::backdrop {
```
dauriamarco, line 58: On the `::backdrop` it is impossible to use any design tokens, so inline values must be used (for background and animation duration in this case).

To learn more: [https://stackoverflow.com/questions/58818299/css-variables-not-working-in-dialogbackdrop](url)

src/components/sbb-menu/sbb-menu.scss
```diff
@@ -0,0 +1,163 @@
...
+      animation: none;
+    }
+
+    &::after {
+      display: none;
+    }
+  }
+}
+
+@keyframes show {
```
dauriamarco, line 121: It is necessary to use animations with keyframes instead of transitions in order not to alter the default `display: block` of the modal otherwise it causes several problems, especially for accessibility.

src/components/sbb-menu-action/sbb-menu-action.e2e.ts
```diff
@@ -0,0 +1,13 @@
+import { newE2EPage } from '@stencil/core/testing';
```
DavideMininni-Fincons, line 1: Add tests for click

dauriamarco:
> ### Additional notes
> I would add to the PR some additional notes to consider regarding some points in the implementation:
> 
> * tests of the `sbb-menu` component are limited by the fact that the dialog element `HTMLDialogElement` is not supported by Jest, preventing basic methods such as `showModal` from being recognized;
> * chromatic has some problems making snapshots correctly due to absolute positioning of the element and transition on opening, but setting the `inlineStories: false` property and suspending animations almost completely fixes it (sometimes it fails to capture the opened menu in time for some breakpoint).

More notes here: 
[https://github.com/lyne-design-system/lyne-components/pull/1378#discussion_r985499900](url)
[https://github.com/lyne-design-system/lyne-components/pull/1378/files#r985508287](url)

src/components/sbb-menu-action/sbb-menu-action.scss
```diff
@@ -0,0 +1,94 @@
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin
+@include host-component-properties;
+
+:host {
+  --sbb-menu-action-border-radius: var(--sbb-border-radius-8x);
+}
+
+:host(:focus-within) {
+  outline: none;
```
DavideMininni-Fincons, line 10: Check if `@include focusOutline;` could be used.

src/components/sbb-menu-action/sbb-menu-action.scss
```diff
@@ -0,0 +1,94 @@
...
+    color: var(--sbb-color-graphite-default);
+  }
+
+  .sbb-menu-action__label {
+    color: var(--sbb-color-graphite-default);
+    text-decoration: line-through;
+  }
+
+  .sbb-menu-action__amount {
+    display: none;
```
DavideMininni-Fincons, line 92: Use variable (see line 69 `display: flex;`).

src/components/sbb-menu/readme.md
```diff
@@ -0,0 +1,97 @@
+# sbb-menu
+
+The `sbb-menu` is a component that you can attach to any element (trigger) to display a context menu.
```
DavideMininni-Fincons, line 3:
```suggestion
The `sbb-menu` is a component that can be attached to any element (trigger) to display a context menu. 
```

src/components/sbb-menu/readme.md
```diff
@@ -0,0 +1,97 @@
...
+
+On mobile, the menu is displayed as a sheet with a backdrop; 
+on desktop it will be shown as a floating menu and will calculate the optimal position relative to the trigger element 
+by evaluating the available space with the following priority: start/below, start/above, end/below, end/above.
+
+## Usage
+
+The menu component allows you to present a custom menu that allows you to perform actions relevant to the current task 
+or to navigate within or outside the application by using the `sbb-menu-action` component along with it 
+or providing custom content inside the dialog as shown below:
```
DavideMininni-Fincons, line 14:
```suggestion
or providing custom content inside the component as shown below:
```

src/components/sbb-menu/readme.md
```diff
@@ -0,0 +1,97 @@
...
+    
+    <sbb-menu-action icon="cross-small">Cancel</sbb-menu-action>
+</sbb-menu>
+```
+
+Clicking in the backdrop or pressing the `ESC` key closes the menu.
+
+### Accessibility
+
+As the menu opens, the focus will automatically be set to the first focusable item within the dialog.
```
DavideMininni-Fincons, line 42:
```suggestion
As the menu opens, the focus will automatically be set to the first focusable item within the component.
```

src/components/sbb-menu/sbb-menu.scss
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+@include ifForcedColors {
+  dialog {
+    outline: var(--sbb-border-width-1x) solid CanvasText;
+  }
+}
+
+@include mq($from: medium) {
```
DavideMininni-Fincons, line 95: As a general suggestion:

`:host {
  --sbb-menu-position: fixed;
}

.sbb-menu {
  position: var(--sbb-menu-position);
}

@include mq($from: medium) {
  .sbb-menu {
    --sbb-menu-position: absolute;
  }
}`

src/components/sbb-menu/sbb-menu.stories.js
```diff
@@ -0,0 +1,227 @@
...
+    </sbb-menu-action>
+    <sbb-divider />
+    <sbb-menu-action icon="cross-small">Cancel</sbb-menu-action>
+  </sbb-menu>,
+];
+
+export const Default = DefaultTemplate.bind({});
+Default.argTypes = defaultArgTypes;
+Default.args = { ...defaultArgs };
+Default.args.disabled = true;
```
DavideMininni-Fincons, line 180: You can use spread operator to override values (same for CustomContent & LongContent).

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,292 @@
...
+  // Check if the trigger is valid and attach click event listener.
+  private _configureTriggerElement(trigger: string | HTMLElement): void {
+    if (!trigger) {
+      return;
+    }
+
+    // check whether it's a string or an HTMLElement
+    if (typeof trigger === 'string') {
+      this._triggerEl = document.getElementById(trigger);
+    } else if (trigger instanceof window.Element) {
```
DavideMininni-Fincons, line 190: Is the check mandatory? Since `trigger: string | HTMLElement`, if it's not a string, it must be the latter.

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,292 @@
...
+      if (!el.hasAttribute('disabled')) {
+        el.addEventListener('click', () => this.close());
+      }
+    });
+  }
+
+  // Close menu on backdrop clicked.
+  private _dismissOnBackdropClick(): void {
+    this._dialog.addEventListener('mousedown', (event: MouseEvent) => {
+      const rect = this._dialog.getBoundingClientRect();
```
DavideMininni-Fincons, line 216: Could it be reused elsewhere? if yes, create an utility method.

src/global/helpers/breakpoint.ts
```diff
@@ -0,0 +1,19 @@
+type Breakpoint = 'zero' | 'micro' | 'small' | 'medium' | 'wide' | 'large' | 'ultra';
```
DavideMininni-Fincons, line 1: @kyubisation @mcilurzo 
should we consider screen resolution above ultra? (eg. my home monitor is 2580px wide)

src/components/sbb-menu/sbb-menu.scss
kyubisation: Good find 👍 
Please add this as documentation to the above line

src/components/sbb-menu/sbb-menu.tsx
dauriamarco: Yes, as we have already discussed it could also be passed an invalid value or object so maybe it is worth leaving the control in any case.

src/global/helpers/breakpoint.ts
dauriamarco: I improved the `isBreakpoint()` function so that it also covers larger screens, by allowing the possibility of setting only a `min-width`. 👍

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,297 @@
...
+   */
+  @Method()
+  @Listen('sbb-menu-action_click')
+  public async close(): Promise<void> {
+    this.willClose.emit();
+    this._isDismissing = true;
+  }
+
+  // Resets position on window resize.
+  @Listen('resize', { target: 'window', passive: true })
```
kyubisation, line 128: This causes the event listener to be bound to the window, even if the menu is not used.
I would probably prefer, if this is only done, when the menu is opened (and removed when closed/disconnected).
@dauriamarco, @jeripeierSBB Thoughts?
(This also applies to the keydown event.)

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,297 @@
...
+    newValue: string | HTMLElement,
+    oldValue: string | HTMLElement
+  ): void {
+    if (newValue !== oldValue) {
+      this._menuController?.abort();
+      this._configure(this.trigger);
+    }
+  }
+
+  private _dialog: HTMLDialogElement;
```
kyubisation, line 156: Please move the properties above the methods

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,297 @@
...
+  // Check if the trigger is valid and attach click event listeners.
+  private _configure(trigger: string | HTMLElement): void {
+    if (!trigger) {
+      return;
+    }
+
+    // check whether it's a string or an HTMLElement
+    if (typeof trigger === 'string') {
+      this._triggerEl = document.getElementById(trigger);
+    } else if (trigger instanceof window.Element) {
```
kyubisation, line 187: Nitpick
```suggestion
    } else if (trigger instanceof HTMLElement) {
```

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,297 @@
...
+      this._triggerEl = trigger;
+    }
+
+    if (this._triggerEl) {
+      this._menuController = new AbortController();
+      this._triggerEl.addEventListener('click', () => this.open(), {
+        signal: this._menuController.signal,
```
kyubisation, line 194: I really like this pattern 😃 👍

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,297 @@
...
+    this._dialog.addEventListener('pointerdown', this._pointerDownListener, {
+      signal: this._menuController.signal,
+    });
+    this._dialog.addEventListener('pointerup', this._dismissOnBackdropClick, {
+      signal: this._menuController.signal,
+    });
+  }
+
+  // Close menu at any click on an interactive element inside the <sbb-menu> that bubbles to the container.
+  private _dismissOnInteractiveElementClick(): void {
```
kyubisation, line 208: I wonder if it might be better to just apply a click event listener on the wrapper and check if the event originated from one of the interactive elements instead of adding an event listerener to every element?

src/global/mixins.scss
```diff
@@ -55,3 +55,39 @@
...
+    }
+  }
+
+  /* set button(top and bottom of the scrollbar) */
+  &::-webkit-scrollbar-button {
+    display: none;
+  }
+
+  /* style for Firefox */
+  scrollbar-width: thin;
```
kyubisation, line 92: No `scrollbar-color`?

src/global/mixins.scss
```diff
@@ -55,3 +55,39 @@
     height: 0;
   }
 }
+
+@mixin scrollbar($color: 'light') {
```
kyubisation, line 59: Since there are only really two cases, I would prefer to have maybe two mixins `scrollbar-light` and `scrollbar-dark`, but I am also open to other suggestions.
(I would like to avoid the need to check every time, what parameter should be used to apply the other case.)

src/global/mixins.scss
```diff
@@ -55,3 +55,39 @@
...
+  }
+
+  /* scrollbar itself */
+  &::-webkit-scrollbar-thumb {
+    background-color: var(--sbb-color-black-alpha-30);
+    border: #{px-to-rem-build(7)} solid transparent;
+    border-radius: var(--sbb-border-radius-4x);
+    background-clip: padding-box;
+
+    @if $color == 'dark' {
```
kyubisation, line 72: It might be easier to use the if function `if($color == 'dark, ... , ...)` (https://www.tutorialspoint.com/sass/if_function.htm)

kyubisation (changes requested):
Thank you very much for your work!
Some suggestions and questions. Feel free to discuss them with @jeripeierSBB next week.

src/components/sbb-menu/sbb-menu.tsx
dauriamarco: True, I did as you suggested for the resize event. While for the keydown event I just removed the `target: 'window'` since it wasn't really needed.

src/components/sbb-menu/sbb-menu.tsx
dauriamarco: That's right, it's much simpler this way. Thank you! 👍

src/global/mixins.scss
dauriamarco: I tried to find an alternative to this problem. I introduced the `scrollbar-ligh`t and `scrollbar-dark` mixins, but you can also use the simple `scrollbar` mixin which will take the currentcolor by default, otherwise it will take the color defined through the variables `-sbb-scrollbar-color` and `--sbb-scrollbar-color-hover`, where you can pass any color to customize the scrollbar before including the mixin in your CSS.

It would be helpful if you could use currentcolor with opacity to automatically define the scrollbar color or at least manage opacity in a simpler way. But the `opacity` property doesn't work in Chrome (strangely, it only works in Safari).

src/global/mixins.scss
dauriamarco: Thanks, though I think it's no longer needed once the `scrollbar-light` and `scrollbar-dark` mixins are introduced.

src/components/sbb-menu/sbb-menu.scss
```diff
@@ -0,0 +1,178 @@
...
+  --sbb-menu-max-width: #{px-to-rem-build(320)};
+  --sbb-menu-min-width: #{px-to-rem-build(240)};
+  --sbb-menu-border-radius: var(--sbb-border-radius-4x);
+}
+
+// On the `::backdrop` it is impossible to use any design tokens,
+// so inline values must be used (for background and animation duration in this case).
+// To learn more: https://stackoverflow.com/questions/58818299/css-variables-not-working-in-dialogbackdrop
+::backdrop {
+  --sbb-menu-backdrop-color: rgb(0 0 0 / 20%);
```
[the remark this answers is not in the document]
dauriamarco: I don't know if I understand the suggestion correctly, but I don't think it is possible to use variables already defined outside `backdrop` in general. You can only define new ones to be reused...

src/components/sbb-menu/sbb-menu.tsx
dauriamarco: I changed it to `window.HTMLElement` because otherwise I get this error:
`Rollup: Missing Export: ./src/components/sbb-menu/sbb-menu.tsx:1:131
           'HTMLElement' is not exported by
           ./node_modules/@stencil/core/internal/hydrate/index.js,
           imported by ./src/components/sbb-menu/sbb-menu.tsx`

Changing `HTMLElement` to `window.HTMLElement` quickly fix this.

jeripeierSBB: We did not find a solution but added a TODO, maybe you know more? :)

jeripeierSBB (approved):
LGTM, expecting some other tests :-)

dauriamarco:
> LGTM, expecting some other tests :-)

Tests on focus have also been added  😃 👍

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,371 @@
...
+      this._dialog.firstElementChild.scrollTo(0, 0);
+      this._dialog.close();
+      this.didDismiss.emit();
+      this._resizeController?.abort();
+    }
+  }
+
+  // Avoid possible double focus on some elements.
+  private _focusMenuElement(element: HTMLElement): void {
+    if (['SBB-BUTTON', 'SBB-LINK', 'SBB-MENU-ACTION'].includes(element.nodeName)) {
```
kyubisation, line 300:
```suggestion
    // TODO: Consider handling/forwarding focus event on these components
    if (['SBB-BUTTON', 'SBB-LINK', 'SBB-MENU-ACTION'].includes(element.nodeName)) {
```

kyubisation (approved):
One question about event listeners.
Other than that great work! 😃 👍

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -0,0 +1,371 @@
...
+      this._configure(this.trigger);
+    }
+  }
+
+  public connectedCallback(): void {
+    // Validate trigger element and attach event listeners
+    this._configure(this.trigger);
+  }
+
+  public componentDidLoad(): void {
```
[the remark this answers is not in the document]
dauriamarco: Done 👍

kyubisation (approved):
LGTM 👍

PR #1379, opened 2022-09-16
refactor: introduce functionality to mark slotted elements

kyubisation:
We have some use cases where it would be beneficial to have the context of a slotted element that can be addressed via a CSS selector. This PR adds a `SlotMarker` class, which can set the `data-slot-context` attribute of slotted elements.


kyubisation:
Won't do for now

PR #1394, opened 2022-10-03
fix(sbb-link): expand clickable area in block variant

### LINKED ISSUE #1387: [Feature request]: sbb-link variant block: the <a> should be width 100%
### Request type

### Functionality

When we use the sbb-link, variant block, 
the <sbb-link> itself is taking 100% width, ok
but the <a> in it, is still inline, and taking only the size of the text instead of the size of the container. 
-> the <a> should also be display:block, so the entire line should be clickable

See the anchor navigation ("Inhalt") : https://int-www.sbb.ch/de/hilfe-und-kontakt/meine-abos/ga-kuendigen.html

### Is there a design proposal?

No

### Link to design proposal file (optional)

_No response_

### Preflight Checklist

- [X] I have searched the issue tracker for an issue that matches the one I want to file, without success.

jeripeierSBB:
Fixes #1387

### DIFF
--- a/src/components/sbb-link/sbb-link.tsx
+++ b/src/components/sbb-link/sbb-link.tsx
@@ -8,6 +8,7 @@ import {
   JSX,
   Listen,
   Prop,
+  State,
 } from '@stencil/core';
 import {
   ButtonType,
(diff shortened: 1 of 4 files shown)


src/components/sbb-link/sbb-link.tsx
```diff
@@ -153,19 +160,12 @@ export class SbbLink implements LinkButtonProperties, ComponentInterface {
...
-    const iconPositionClass =
-      this.iconPlacement === 'start'
-        ? ' sbb-link--icon-placement-start'
-        : ' sbb-link--icon-placement-end';
-    const inlineClass = this.variant === 'inline' ? ' sbb-link--inline' : '';
-    const negativeClass = this.negative ? ' sbb-link--negative' : '';
-
-    return `sbb-link${textSizeClass}${iconPositionClass}${inlineClass}${negativeClass}`;
+  /** State of listed named slots, by indicating whether any element for a named slot is defined. */
+  @State() private _namedSlots = createNamedSlotState('icon');
```
kyubisation, line 164: Please move this property to the other properties

kyubisation (changes requested):

src/components/sbb-link/sbb-link.tsx
jeripeierSBB: done

PR #1399, opened 2022-10-05
refactor: forward click on host to inner action element

jeripeierSBB:


src/components/sbb-button/sbb-button.tsx
```diff
@@ -151,6 +152,20 @@ export class SbbButton implements LinkButtonProperties, ComponentInterface {
     }
   }
 
+  @Listen('click')
+  public handleClick(event: Event): void {
+    if (this.disabled) {
```
kyubisation, line 157: I wonder if we should provide an additional function `preventOrForwardHostClick` which additionally checks the disabled state?

kyubisation (approved):
LGTM 👍

src/components/sbb-button/sbb-button.tsx
jeripeierSBB: I'd leave it as it is, as then the disabled state becomes a little bit hidden. Maybe I have to explain it in a call ;)

PR #1403, opened 2022-10-06
feat: add possibility to submit forms with sbb-button/sbb-link

dauriamarco:
Provide `sbb-button` and `sbb-link` with `type="submit"` with the ability to submit the form in which they are embedded.


jeripeierSBB (commented):
Looks good, but I think link and card should be updated as well? I think whenever we provide the type property, we also should handle this?

dauriamarco:
> Looks good, but I think link and card should be updated as well? I think whenever we provide the type property, we also should handle this?

Sure, unless we want to limit the ability to submit the form to the button and perhaps the link button. I don't know what are the use cases of a card involving a form to submit. But in any case it sounds good to me.

jeripeierSBB (approved):
LGTM

PR #1404, opened 2022-10-06
refactor(sbb-teaser): use LinkProperties interface

jeripeierSBB:


kyubisation (approved):
LGTM 👍

src/components/sbb-teaser/sbb-teaser.tsx
```diff
@@ -17,7 +24,28 @@ import { AccessibilityProperties } from '../../global/interfaces/accessibility-p
...
+   */
+  @Prop({ reflect: true }) public isStacked: boolean;
+
+  /**
+   * Heading level of the sbb-title element (e.g. h1-h6).
+   */
+  @Prop() public titleLevel: InterfaceTeaserAttributes['titleLevel'] = '5';
+
+  /** The href value you want to link to. */
+  @Prop() public href!: string | undefined;
```
DavideMininni-Fincons, line 41: Since it's mandatory, is the undefined type needed?

DavideMininni-Fincons (commented):
Looks good, added just a comment;
should we change the event type to `Event` in src/global/interfaces/link-button-properties.ts line 76 (the `emitButton` function)?

src/components/sbb-teaser/sbb-teaser.tsx
jeripeierSBB: I removed the mandatory flag, as it renders to span if there is no href

PR #1412, opened 2022-10-19
feat: add experimental sass output

kyubisation:
With this PR our sass code can be used by consumers (e.g. via `@import "@sbb-esta/lyne-components";`).
This is still experimental and sass code in our code base WILL change in the future.


DavideMininni-Fincons (approved):
Looks good, thanks!

PR #1426, opened 2022-10-25
feat(sbb-slider): component implementation

### LINKED ISSUE #1163: story(sbb-slider): initial implementation
### Feature Description
We want to provide a sbb-slider.

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=140%3A30)

**References**
- PoC Branch: https://github.com/lyne-design-system/lyne-components/tree/feat-slider/src/components/sbb-slider
- Styling considerations: https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range
- https://material.angular.io/components/slider/overview
- https://react.carbondesignsystem.com/?path=/story/components-slider--default

**Design Spec**
- ~~Both a normal and a stepped variant should be supported~~ (@mcilurzo is checking whether stepped variant is needed visually)
- Icon at both start and end of slider
- **Tooltip out of scope**
- **Input out of scope**
- Open question: Should the icons be interactive? @mcilurzo checks this with sbb.ch UX

**Technical Spec**
- Use native `<input type="range">` internally
- Properties:
  - value?: string (Should be forwarded to and updated from the native input) (Investigate whether type number works for this property/attribute)
  - valueAsNumber?: number (Should be forwarded to and updated from the native input)
  - disabled: boolean (@mcilurzo will provide designs)
  - readonly: boolean (@mcilurzo will provide designs)
  - step?: number
  - max?: number (Should default to 100)
  - min?: number (Should default to 0)
- `sbbChange` event 
- Should be supported inside a `<sbb-form-field>`. This will require adaptation inside the `<sbb-form-field>` (check focus management, accessibility aria-describedby update, etc.).
- Provide properties for start and end icons with default icons (@mcilurzo will defined default icons)
  - Sizes of icons should be fixed to 24px (use appropriate design token)
- Use default width (with the input range being 400px, to be challenged by @mcilurzo)
  - When used within an sbb-form-field, it should stretch to the dimension of the form field
  - @mcilurzo will provide the minimum width

### Definition of Done

 - [ ]  Component is implemented
 - [ ]  Documentation is provided
 - [ ]  Tests are implemented
 - [ ]  UX approved

DavideMininni-Fincons:
## Preflight Checklist

## Issue

This PR Closes #1163

## Pull request checklist

## Changes

Changes in this pull request:

- component implementation

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


mcilurzo:
ux-review

- [ ] issue-1: https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj?node-id=13931:466933#295643073
- [ ] issue-2: https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj?node-id=13931:466933#295655713
- [ ] issue-3: https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj?node-id=13931:466982#295649116
- [ ] issue-4: https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj?node-id=13931:467738#295657837

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,149 @@
...
+
+  height: var(--sbb-slider-line-height);
+  width: 100%;
+  background-color: var(--sbb-slider-line-color);
+  border-radius: var(--sbb-slider-line-height);
+  overflow: hidden;
+  pointer-events: none;
+}
+
+.sbb-slider__line-disabled,
```
kyubisation, line 85: Prefer host selectors for variations
```scss
.sbb-slider__line {
  ...

  :host([disabled]) & {
    ...
  }
}
```

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,149 @@
...
+}
+
+.sbb-slider__selected-line {
+  position: relative;
+  height: 100%;
+  background-color: var(--sbb-slider-selected-line-color);
+  width: calc(var(--sbb-slider-value-fraction) * 100%);
+}
+
+.sbb-slider__selected-line-disabled,
```
kyubisation, line 104: Prefer host selectors for variations
```scss
.sbb-slider__selected-line {
  ...

  :host([disabled]) & {
    ...
  }
}
```

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,150 @@
...
+  transition-property: width, height;
+  transition-duration: var(--sbb-animation-duration-1x);
+  transition-timing-function: ease-in;
+  background-color: var(--sbb-color-white-default);
+  border: var(--sbb-slider-knob-border-size) solid var(--sbb-slider-selected-line-color);
+  border-radius: 50%;
+  pointer-events: none;
+}
+
+.sbb-slider__knob-disabled {
```
kyubisation, line 133: Prefer host selectors for variations
```scss
.sbb-slider__knob {
  ...

  :host([disabled]) & {
    ...
  }
}
```

src/components/sbb-slider/sbb-slider.tsx
```diff
@@ -0,0 +1,197 @@
...
+  @Prop() public accessibilityLabel: string | undefined;
+
+  /** This will be forwarded as aria-describedby to the relevant nested element. */
+  @Prop() public accessibilityDescribedby: string | undefined;
+
+  /** This will be forwarded as aria-labelledby to the relevant nested element. */
+  @Prop() public accessibilityLabelledby: string | undefined;
+
+  /** Event emitted when the value of the inner HTMLInputElement changes. */
+  @Event() public sbbChange: EventEmitter<SbbSliderChange>;
```
kyubisation, line 70: I'm inclined to drop custom form events (for change and input), similar to what we decided for action elements.
@jeripeierSBB, @DavideMininni-Fincons, @simone-sabato Thoughts?

src/components/sbb-slider/sbb-slider.tsx
```diff
@@ -0,0 +1,197 @@
...
+      disabled: this.disabled || this.readonly || null,
+      valueAsNumber: this.valueAsNumber || null,
+      value: this.value || null,
+      'aria-label': this.accessibilityLabel || null,
+      'aria-describedby': this.accessibilityDescribedby || null,
+      'aria-labelledby': this.accessibilityLabelledby || null,
+    };
+
+    return (
+      <Host class={{ 'sbb-form-field-element': this._isInFormField }}>
```
kyubisation, line 143: I've been thinking about the integration into `sbb-form-field`. My suggestion at the time was `class="sbb-form-field-element"`.
I now feel like that might have been a mistake, as it interferes with the class attribute on the host. While not a major issue, I still feel like a separate attribute would be cleaner.
My suggestion now would be an attribute called `data-sbb-form-field`, which could be set via `this._element.dataset.sbbFormField = '';`.
@jeripeierSBB, @DavideMininni-Fincons, @dauriamarco, @federicoisepponfincons, @simone-sabato Thoughts?

src/components/sbb-slider/sbb-slider.tsx
```diff
@@ -0,0 +1,197 @@
...
+        <div class="sbb-slider__wrapper">
+          {this.startIcon && (
+            <slot name="prefix">
+              <sbb-icon name={this.startIcon} />
+            </slot>
+          )}
```
kyubisation, line 149: The slot should always be rendered.
```suggestion
          <slot name="prefix">
            {this.startIcon && <sbb-icon name={this.startIcon} />}
          </slot>
```

src/components/sbb-slider/sbb-slider.tsx
```diff
@@ -0,0 +1,197 @@
...
+                'sbb-slider__knob-disabled': this.disabled,
+                'sbb-slider__knob-readonly': this.readonly,
+              }}
+            ></div>
+          </div>
+          {this.endIcon && (
+            <slot name="suffix">
+              <sbb-icon name={this.endIcon} />
+            </slot>
+          )}
```
kyubisation, line 192: The slot should always be rendered.
```suggestion
          <slot name="suffix">
            {this.endIcon&& <sbb-icon name={this.endIcon} />}
          </slot>
```

kyubisation (changes requested):
Nice work!

src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -5,8 +5,9 @@
 
 :host {
   display: inline-block;
-  width: #{px-to-rem-build(300)};
+  max-width: #{px-to-rem-build(300)};
```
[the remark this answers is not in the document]
federicoisepponfincons: The width is still 300px, but this make the form field shrink in smaller breakpoint situations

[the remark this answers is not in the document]
kyubisation: Could we maybe use `width: max(#{px-to-rem-build(300)}, 100%);` instead of max-width? I would prefer if consumers only need to override one property to change the size.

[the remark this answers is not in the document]
federicoisepponfincons: unfortunately no, because max() sets the largest possible between the two values, going over 300px
but I understand the point, I see if there is a solution that involves only one property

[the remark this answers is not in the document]
kyubisation: Maybe using `min` instead of `max`. I always confuse them on how to use them in CSS.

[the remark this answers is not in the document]
federicoisepponfincons: exactly :)

src/components/sbb-form-field/sbb-form-field.tsx
```diff
@@ -133,14 +135,16 @@ export class SbbFormField implements ComponentInterface {
...
+    const slot = (event.target as HTMLSlotElement).assignedElements();
+    this._isNativeInput = slot.some(
+      (e): boolean => e.tagName === 'INPUT' || e.tagName === 'SELECT'
+    );
+    this._input = slot.find(
+      (e): e is HTMLElement =>
+        e.tagName === 'INPUT' ||
+        e.tagName === 'SELECT' ||
+        e.classList.contains('sbb-form-field-element')
+    );
```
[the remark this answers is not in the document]
DavideMininni-Fincons: `_focusInputElement` has been removed, so no more need for `_isNativeInput`

src/components/sbb-slider/sbb-slider.tsx
jeripeierSBB: I'd vote for the data attribute :)

jeripeierSBB (changes requested):

src/components/sbb-slider/sbb-slider.tsx
DavideMininni-Fincons: Me too; let's discuss it tomorrow during refinement..

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,148 @@
+// Default component properties, defined for :host. Properties which can not
```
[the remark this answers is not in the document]
DavideMininni-Fincons: Probably is not desired; @mcilurzo can you confirm if should we turn it off?

src/components/sbb-slider/sbb-slider.e2e.ts
```diff
@@ -0,0 +1,104 @@
...
+      <sbb-slider start-icon="walk-slow-small" end-icon="walk-fast-small" max="500" min="100" value="400"/>
+    `);
+    element = await page.find('sbb-slider');
+  });
+
+  it('renders', async () => {
+    expect(element).toHaveClass('hydrated');
+    await page.waitForChanges();
+
+    // NOTE: value at line 22 is not reflected; probably related to bug https://github.com/ionic-team/stencil/issues/2946
```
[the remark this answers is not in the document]
DavideMininni-Fincons: It seems not; in the browser is still missing (see image) and the test still fails if value is added in the input tag.

image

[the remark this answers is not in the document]
jeripeierSBB: I meant the line number is incorrect, not the comment itself ;)

src/components/sbb-slider/sbb-slider.tsx
kyubisation: I agree, but most of these values could also be read from the target element.

src/components/sbb-slider/sbb-slider.e2e.ts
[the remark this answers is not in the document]
DavideMininni-Fincons: Comment restored and line number fixed (sorry for the misunderstanding) 😄

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,148 @@
...
+  --sbb-slider-selected-line-disabled-color: var(--sbb-color-smoke-default);
+  --sbb-slider-line-color: var(--sbb-color-smoke-default);
+  --sbb-slider-line-disabled-color: var(--sbb-color-graphite-default);
+
+  @include ifForcedColors {
+    --sbb-slider-selected-line-color: Highlight;
+    --sbb-slider-line-color: CanvasText;
+  }
+
+  line-height: var(--sbb-typo-line-height-body-text);
```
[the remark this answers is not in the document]
DavideMininni-Fincons: I did various tests and as you said it seems useless, so i removed it.

src/components/sbb-slider/sbb-slider.scss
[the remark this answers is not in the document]
mcilurzo: In my opinion, the animations should be switched off.

src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -155,11 +156,12 @@ slot[name='label'] > span {
 
 // Input
 /* stylelint-disable plugin/stylelint-bem-namics */
-.form-field__input ::slotted(:where(input, select, .sbb-form-field-element)) {
+.form-field__input ::slotted(:where(input, select, [data-sbb-form-field])) {
   @include text-m--regular;
   @include placeholder;
   @include ellipsis;
 
+  overflow: var(--form-field-overflow);
```
jeripeierSBB, line 164: I'm just interested, which is the use case where an overflow hidden is necessary?

src/components/sbb-form-field/sbb-form-field.tsx
```diff
@@ -139,8 +139,9 @@ export class SbbFormField implements ComponentInterface {
         (e): e is HTMLElement =>
           e.tagName === 'INPUT' ||
           e.tagName === 'SELECT' ||
-          e.classList.contains('sbb-form-field-element')
+          e.getAttribute('data-sbb-form-field') !== null
```
jeripeierSBB, line 142: would it be possible to use hasAttribute?
```suggestion
          e.hasAttribute('data-sbb-form-field')
```

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,148 @@
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin
+@include host-component-properties;
+
+:host {
+  --sbb-slider-line-height: #{px-to-rem-build(4)};
```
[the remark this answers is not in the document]
jeripeierSBB: What confuses me is, that if you use display: inline, it still takes full width and following text is on the next line. I'd expect that text comes directly after the slider (as it would be with inline-block). But maybe it's ok for this case, but then maybe it can be only display: block from default values, as it doesn't seem to make any difference

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,141 @@
...
+  width: 100%;
+  background-color: var(--sbb-slider-line-color);
+  border-radius: var(--sbb-slider-line-height);
+  overflow: hidden;
+  pointer-events: none;
+
+  :host([disabled]) &,
+  :host([readonly]) & {
+    --sbb-slider-line-color: var(--sbb-slider-line-disabled-color);
+  }
```
jeripeierSBB, line 86: I'd prefer this definition directly on host level, I mean at the moment at line 42

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,141 @@
...
+.sbb-slider__selected-line {
+  position: relative;
+  height: 100%;
+  background-color: var(--sbb-slider-selected-line-color);
+  width: calc(var(--sbb-slider-value-fraction) * 100%);
+
+  :host([disabled]) &,
+  :host([readonly]) & {
+    --sbb-slider-selected-line-color: var(--sbb-slider-selected-line-disabled-color);
+  }
```
jeripeierSBB, line 98: I'd prefer this definition directly on host level, I mean at the moment at line 42

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,141 @@
...
+  pointer-events: none;
+
+  :host([disabled]) & {
+    --sbb-slider-knob-color: var(--sbb-color-milk-default);
+    --sbb-slider-knob-border-color: var(--sbb-slider-line-disabled-color);
+  }
+
+  :host([readonly]) & {
+    --sbb-slider-knob-border-color: var(--sbb-slider-selected-line-disabled-color);
+  }
```
jeripeierSBB, line 126: I'd prefer this definitions on the most outer level as the other comments

src/components/sbb-slider/sbb-slider.stories.js
```diff
@@ -0,0 +1,261 @@
+import events from './sbb-slider.events.ts';
+import { h } from 'jsx-dom';
+import readme from './readme.md';
+
+const TemplateSbbSlider = (args) => <sbb-slider {...args}></sbb-slider>;
+
+const templateSlottedIcons = (args) => (
```
jeripeierSBB, line 7:
```suggestion
const TemplateSlottedIcons = (args) => (
```

jeripeierSBB (changes requested):

src/components/sbb-form-field/sbb-form-field.scss
federicoisepponfincons: overflow hidden comes from the @include ellipsis, it's necessary for ellipsis to work
on the other side (slider) this instruction hides part of the knob

jeripeierSBB: ah I see now what confused me, it was the new assignment on line 24. `--form-field-overflow: hidden;`. because you override the mixins value, it's necessary to redefine it. It's ok.

src/components/sbb-slider/sbb-slider.scss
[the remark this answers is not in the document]
federicoisepponfincons: @jeripeierSBB probably the best solution is to keep display: inline-block but also move these

```suggestion
  max-width: #{px-to-rem-build(400)};
  min-width: #{px-to-rem-build(150)};
  width: 100%;
```

from .sbb-slider__wrapper to :host

in this way it behaves as a inline-block element and still has correct dimensions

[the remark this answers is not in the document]
jeripeierSBB: But then it's again 100% width as defined, and we should avoid defining width on host. So maybe there is no perfect solution, or I just accept that it is a block element.
Other question: why is width 100% needed on host (if you would change) ?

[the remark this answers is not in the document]
federicoisepponfincons: in this case width 100% is needed to make the element shrink between min and max width, otherwise it sets to the minimum (150px)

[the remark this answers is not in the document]
jeripeierSBB: Yes but this is what I'd expect and the consumer has to set the width he likes. If there is no width, take minimum. in form-field then it should be 100%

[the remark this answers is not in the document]
federicoisepponfincons: Also the form-field element has min and max width already set to 150 and 300
Can the consumer eventually overrides max-width property?

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,139 @@
...
+    --sbb-slider-knob-border-color: var(--sbb-slider-line-disabled-color);
+  }
+}
+
+:host([readonly]) {
+  --sbb-slider-icon-color: var(--sbb-color-smoke-default);
+
+  .sbb-slider__knob {
+    --sbb-slider-knob-border-color: var(--sbb-slider-selected-line-disabled-color);
+  }
```
jeripeierSBB, line 107: selector should not be necessary here
```suggestion
    --sbb-slider-knob-border-color: var(--sbb-slider-selected-line-disabled-color);

```

src/components/sbb-slider/sbb-slider.scss
```diff
@@ -0,0 +1,139 @@
...
+:host([readonly]) {
+  @include ifForcedColors {
+    --sbb-slider-icon-color: GrayText;
+    --sbb-slider-selected-line-disabled-color: GrayText;
+    --sbb-slider-line-disabled-color: GrayText;
+  }
+
+  .sbb-slider__line {
+    --sbb-slider-line-color: var(--sbb-slider-line-disabled-color);
+  }
```
jeripeierSBB, line 120: selector should not be necessary
```suggestion
    --sbb-slider-line-color: var(--sbb-slider-line-disabled-color);
```

jeripeierSBB (changes requested):
some style issues

jeripeierSBB (changes requested):
some style issues

src/components/sbb-slider/sbb-slider.scss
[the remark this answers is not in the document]
jeripeierSBB: @federicoisepponfincons In your solution, you would then apply display: inline-block to the host?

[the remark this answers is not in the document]
jeripeierSBB: The more I think, I guess Federicos idea is a good solution that the width can be set directly outside, otherwise we should provide css vars. But anyways, style on host should always be considered carefully :)

src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -5,8 +5,8 @@
 
 :host {
   display: inline-block;
-  width: #{px-to-rem-build(300)};
   min-width: #{px-to-rem-build(150)};
+  width: min(#{px-to-rem-build(300)}, 100%);
```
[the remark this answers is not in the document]
jeripeierSBB: With the current solution it's probably easier to overwrite from outside. So I think it's ok.

src/global/interfaces/link-button-properties.ts
```diff
@@ -275,3 +275,15 @@ export function forwardHostEvent(
...
   nestedActionElement.dispatchEvent(copiedEvent);
 }
+
+/**
+ * Forwards a change on the nested input element to the host element.
+ * In this way an event triggered in the ShadowDOM can cross its boundary and can be listened on the host component.
+ */
+export function forwardInnerEventToHost(event: Event, host: HTMLElement): void {
+  event.preventDefault();
+  event.stopImmediatePropagation();
```
jeripeierSBB, line 285: do we need to stop it as it doesn't bubble shadow dom boundaries anyways?. But it's also ok to stop it :)

src/global/interfaces/link-button-properties.ts
```diff
@@ -275,3 +275,15 @@ export function forwardHostEvent(
 
   nestedActionElement.dispatchEvent(copiedEvent);
 }
+
+/**
+ * Forwards a change on the nested input element to the host element.
+ * In this way an event triggered in the ShadowDOM can cross its boundary and can be listened on the host component.
+ */
+export function forwardInnerEventToHost(event: Event, host: HTMLElement): void {
```
jeripeierSBB, line 283: I'm not sure if this is the right place as it has nothing to do with link-button-properties

jeripeierSBB (changes requested):
Please also fix test

src/global/interfaces/link-button-properties.ts
DavideMininni-Fincons: Yes, moved to a new file 👍

src/global/helpers/forward-event.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Forwards a change on the nested input element to the host element.
+ * In this way an event triggered in the ShadowDOM can cross its boundary and can be listened on the host component.
+ */
+export function forwardInnerEventToHost(event: Event, host: HTMLElement): void {
+  const eventConstructor = Object.getPrototypeOf(event).constructor;
+  const copiedEvent: Event = new eventConstructor(event.type, event);
+  host.dispatchEvent(copiedEvent);
+}
```
jeripeierSBB, line 9:
```suggestion
/**
 * Forwards an event to the host element provided.
 * With this way, an event triggered in the ShadowDOM can cross its boundary and can be listened on the host component.
 */
export function forwardEventToHost(event: Event, host: HTMLElement): void {
  const eventConstructor = Object.getPrototypeOf(event).constructor;
  const copiedEvent: Event = new eventConstructor(event.type, event);
  host.dispatchEvent(copiedEvent);
}
```

src/components/sbb-slider/sbb-slider.tsx
jeripeierSBB: It was not that easy as the target becomes the sbb-slider. but now we update the value there as well, so it's readable from outside.

src/global/helpers/forward-event.ts
DavideMininni-Fincons: Done in [af92fa8](https://github.com/lyne-design-system/lyne-components/pull/1426/commits/af92fa81c5233debb291861981681a74de186fea)

jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍 
Thanks to everyone for your work!

PR #1454, opened 2022-11-14
feat(sbb-dialog): add `sbb-dialog-close` attribute option to close the dialog

dauriamarco:
Extend `sbb-dialog` with option to close dialog via `sbb-dialog-close` attribute.

### DIFF
--- a/src/components/sbb-dialog/sbb-dialog.tsx
+++ b/src/components/sbb-dialog/sbb-dialog.tsx
@@ -22,6 +22,7 @@ import {
 import { IS_FOCUSABLE_QUERY, FocusTrap } from '../../global/helpers/focus';
 import { i18nCloseDialog, i18nGoBack } from '../../global/i18n';
 import getDocumentLang from '../../global/helpers/get-document-lang';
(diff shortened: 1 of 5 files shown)


src/components/sbb-dialog/sbb-dialog.tsx
```diff
@@ -274,6 +277,21 @@ export class SbbDialog implements AccessibilityProperties {
...
 
+  // Close the dialog on click of any element that has the 'sbb-dialog-close' attribute.
+  private _closeOnSbbDialogCloseClick(event: Event): void {
+    const target = event.target as HTMLElement;
+
+    // Check if the target is a submit element within a form
+    const closestForm =
+      target.getAttribute('type') === 'submit'
+        ? (hostContext('form', target) as HTMLFormElement)
+        : undefined;
```
jeripeierSBB, line 288: The closest form is only used inside the next if condition, I'd suggest to wrap it into it.

jeripeierSBB (approved):
LGTM

PR #1457, opened 2022-11-17
refactor(sbb-signet): adapt to coding standards

jeripeierSBB:
BREAKING CHANGE:
- sbb-logo no longer has different color variants
- sbb-signet no longer has white-on-black variant


kyubisation (approved):
LGTM 👍

PR #1461, opened 2022-11-17
fix(sbb-toggle-check): fix toggle vertical alignment

federicoisepponfincons:
## Preflight Checklist

## Issue

sbb-toggle-check missed vertical alignment for the toggle element

## Pull request checklist

## Changes

Changes in this pull request:

- vertical alignment for toggle element
- added variant with long label 

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [ ] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

- [ ] JAWS Firefox Desktop
- [ ] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [ ] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/sbb-toggle-check/sbb-toggle-check.scss
+++ b/src/components/sbb-toggle-check/sbb-toggle-check.scss
@@ -1,61 +1,64 @@
 :host {
   display: inline-block;
 
-  --toggle-check-color: var(--sbb-color-charcoal-default);
-  --toggle-check-background: var(--sbb-color-smoke-default);
-  --toggle-check-fill: var(--sbb-color-red-default);
-  --toggle-check-circle-background: var(--sbb-color-white-default);
-  --toggle-check-circle-offset: var(--sbb-border-width-2x);
-  --toggle-check-circle-diameter: var(--sbb-size-icon-ui-small);
-  --toggle-check-height: calc(
-    var(--toggle-check-circle-diameter) + 2 * var(--toggle-check-circle-offset)
+  --sbb-toggle-check-color: var(--sbb-color-charcoal-default);
+  --sbb-toggle-check-background: var(--sbb-color-smoke-default);
+  --sbb-toggle-check-fill: var(--sbb-color-red-default);
+  --sbb-toggle-check-circle-background: var(--sbb-color-white-default);
+  --sbb-toggle-check-circle-offset: var(--sbb-border-width-2x);
+  --sbb-toggle-check-circle-diameter: var(--sbb-size-icon-ui-small);
+  --sbb-toggle-check-height: calc(
+    var(--sbb-toggle-check-circle-diameter) + 2 * var(--sbb-toggle-check-circle-offset)
   );
-  --toggle-check-circle-transform: translate(var(--toggle-check-circle-offset), -50%);
-  --toggle-check-cursor: pointer;
-  --toggle-check-flex-direction: row-reverse;
-  --toggle-check-icon-opacity: 0;
+  --sbb-toggle-check-circle-transform: translate(var(--sbb-toggle-check-circle-offset), -50%);
+  --sbb-toggle-check-cursor: pointer;
+  --sbb-toggle-check-flex-direction: row-reverse;
+  --sbb-toggle-check-icon-opacity: 0;
 
   @include ifForcedColors {
-    --toggle-check-circle-background: Canvas;
-    --toggle-check-background: CanvasText;
+    --sbb-toggle-check-circle-background: Canvas;
+    --sbb-toggle-check-background: CanvasText;
   }
 }
 
-.toggle-check--checked {
-  --toggle-check-background: var(--sbb-color-red-default);
-  --toggle-check-circle-transform: translate(calc(100% - var(--toggle-check-circle-offset)), -50%);
-  --toggle-check-icon-opacity: 1;
+:host([checked]) {
+  --sbb-toggle-check-background: var(--sbb-color-red-default);
+  --sbb-toggle-check-circle-transform: translate(
+    calc(100% - var(--sbb-toggle-check-circle-offset)),
+    -50%
+  );
+  --sbb-toggle-check-icon-opacity: 1;
 
   @include ifForcedColors {
-    --toggle-check-fill: Highlight;
-    --toggle-check-background: Highlight;
+    --sbb-toggle-check-fill: Highlight;
+    --sbb-toggle-check-background: Highlight;
   }
 }
 
-.toggle-check--disabled {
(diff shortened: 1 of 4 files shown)


jeripeierSBB (commented):
And please fix tests

src/components/sbb-toggle-check/sbb-toggle-check.scss
```diff
@@ -81,7 +81,16 @@ input[type='checkbox'] {
...
+  align-items: start;
+}
+
+.toggle-check__label {
+  padding-block-start: calc(
+    (
+        var(--toggle-check-height) -
+          (var(--sbb-typo-line-height-body-text) * var(--sbb-font-size-text-s))
+      ) / 2
+  );
```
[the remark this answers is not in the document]
DavideMininni-Fincons: Hmm, it seems not working; the computed value is 0.4px using the calculation vs. 6px using the 1em solution..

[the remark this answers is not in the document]
jeripeierSBB: I think I accidentaly removed also the line height, which still should be multiplied with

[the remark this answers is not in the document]
DavideMininni-Fincons: Fixed as suggested :)

src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -80,7 +80,9 @@ export class SbbToggleCheck implements AccessibilityProperties {
           aria-labelledby={this.accessibilityLabelledby}
         />
         <span class="toggle-check__container">
-          <slot />
+          <span class="toggle-check__label">
```
jeripeierSBB, line 83: This class should be conditionally be set if there is slotted content or not. See https://www.chromatic.com/test?appId=5ea7d9e48c451f00223b52ff&id=6380cec29ac24f83b5e6afcb . Now there is an invisible space on the right side, if there is no label.

jeripeierSBB (changes requested):

src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -80,9 +94,13 @@ export class SbbToggleCheck implements AccessibilityProperties {
...
-          <span class="toggle-check__label">
-            <slot />
-          </span>
+          {this._hasLabelText ? (
+            <span class="toggle-check__label">
+              <slot onSlotchange={(event): void => this._onLabelSlotChange(event)} />
+            </span>
+          ) : (
+            <slot onSlotchange={(event): void => this._onLabelSlotChange(event)} />
+          )}
```
jeripeierSBB, line 103: what's the purpose of the alternative here? I think we can live without?

src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -47,12 +47,26 @@ export class SbbToggleCheck implements AccessibilityProperties {
...
 
+  @State() private _hasLabelText = false;
+
   /** Method triggered on toggle change. */
   public checkedChanged(event: Event): void {
     this.checked = this._checkbox?.checked;
     forwardEventToHost(event, this._element);
   }
 
+  public connectedCallback(): void {
```
jeripeierSBB, line 58: Can we not use our default approach here with namedSlots? (e.g. in sbb-button, there for the icon)

DavideMininni-Fincons: I tried but the label slot is unnamed, so I can't use it (createNamedSlotState requires an input string, hasNamedSlotElement looks for `slot="${name}"` and so on).
I could update the slot in the toggle-check to `<slot name='label' />`, but it means that consumers should set the label like `<span slot=label'>My label</span>` and I don't really like it.
The _hasLabelText approach is taken and renamed from the sbb-button (_hasText).

src/components/sbb-toggle-check/sbb-toggle-check.tsx
DavideMininni-Fincons: The 'sbb-toggle-check__container' has a gap rule; if the span is in the DOM, rule is applied, even if the nested slot is empty, and this causes the unwanted padding.

src/components/sbb-toggle-check/sbb-toggle-check.tsx
jeripeierSBB: @kyubisation would it be possible to extend your script to listen to an unnamed slot?

kyubisation: I would recommend to just render the slot and apply `hidden` to the containing span, depending on whether there is content or not.
While I think it's important to be efficient, it also adds unneeded complexity in this case (at least to me).

jeripeierSBB: Sometimes I forget the simple solutions, I agree with lukas

src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -85,10 +92,12 @@ export class SbbToggleCheck implements AccessibilityProperties {
           aria-describedby={this.accessibilityDescribedby}
           aria-labelledby={this.accessibilityLabelledby}
         />
-        <span class="toggle-check__container">
-          <slot />
-          <span class="toggle-check__slider">
-            <span class="toggle-check__circle">
+        <span class="sbb-toggle-check__container">
+          <span class="sbb-toggle-check__label" hidden={!this._hasLabelText}>
```
jeripeierSBB, line 96: was a little bit surprised how it's going to be rendered, because i thought it will end up in hidden=false, which would be wrong for a boolean attribute. But it seems to be rendered to correct way, so if false, hidden is not there at all :-)

jeripeierSBB (approved):
LGTM, thanks a lot for also integrating coding standards!

src/components/sbb-toggle-check/sbb-toggle-check.tsx
DavideMininni-Fincons: Double-checked in all desktop browsers, Safari included, seems to work fine 🚀

jeripeierSBB: I think this is already done be tsx somehow so it's browser independent :-)

PR #1476, opened 2022-12-01
feat: add title margins

jeripeierSBB:
BREAKING CHANGE:
Title mixins and `<sbb-title>`-component now have a default margin block to meet design specs. In the mixin you can opt out by passing the flag $exclude-spacing with true (e.g. `@include text-1($exclude-spacing: true)`).
In the `sbb-title` component you can just set any margin from outside on the `<sbb-title>` itself, e.g. `<sbb-title style='margin:0'>`.

### DIFF
--- a/src/components/sbb-alert/sbb-alert.scss
+++ b/src/components/sbb-alert/sbb-alert.scss
@@ -77,7 +77,8 @@
 }
 
 .sbb-alert__title {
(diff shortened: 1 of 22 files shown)


kyubisation (approved):
LGTM 👍

kyubisation (approved):
LGTM 👍

PR #1481, opened 2022-12-05
feat(sbb-toggle): introduce toggle component

### LINKED ISSUE #1169: story(sbb-toggle): initial implementation
### Feature Description
We want to provide a toggle component.
**This excludes the `sbb-toggle-check`.**

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=140%3A26)

**References**

- https://angular.app.sbb.ch/angular/components/toggle/overview
- https://kittygiraudel.com/2021/04/05/an-accessible-toggle/
- https://5f0ca4e61af3790022cad2fe-anufvgkxlz.chromatic.com/?path=/story/components-data-entry-toggle--basic

**Design Spec**

- It should be possible to display label and toggle with width 100%
- Item width should be dynamic or configureable 50%
- The white area should animate to the clicked/activated option
- It should be possible to overwrite the padding (left/right) arrount the label/icon

**Technical Spec**

- native input type="radio" is used internally => probably not possible because of shadow DOM, check with @dauriamarco.
- should have property size `'m' | 's' = 'm'`
- should support `sbb-icon`

- `sbb-toggle`
  - Should support properties `name`, `disabled` and `even` (for fixed 50% width)
  - Should emit native `input` and `change` events (`input` might already bubble from the native `<input type="radio">` element)
  - Should support property `value`, which is the value of the currently selected option, otherwise select the first.
  - If `value` is set programmatically, the corresponding option should be selected

- `sbb-toggle-option`
  - Should support content
  - Should provide property `value`

**Questions**
@mcilurzo : if no value is specified should the first option be selected or no option? material and ionic don't select any option.

```html
<sbb-toggle name="example1">
  <sbb-toggle-option value="val1">Option 1</sbb-toggle-option>
  <sbb-toggle-option value="val2">Option 2</sbb-toggle-option>
</sbb-toggle>
```

### Definition of Done

 - [x]  Tests are implemented
 - [x]  E2E Tests are implemented
 - [x]  Storybook stories are implemented
 - [x]  Navigation via keyboard is tested
 - [x]  Screen reader output is tested
 - [x]  High-contrast is tested
 - [x]  Remaining accessibility is tested
 - [ ]  UX approved

dauriamarco:
## Preflight Checklist

## Issue

This PR Closes #1169

## Pull request checklist

## Changes

Changes in this pull request:

- implement `sbb-toggle` component

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [x] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+    this._toggleResizeObserver.observe(this._element.firstElementChild);
+  }
+
+  public disconnectedCallback(): void {
+    this._toggleResizeObserver.disconnect();
+  }
+
+  private _updateToggle(): void {
+    const options = this._options;
+    const value = this.value ?? options.find((toggleOption) => toggleOption.checked)?.value;
```
jeripeierSBB, line 134: use your getter `_checked` here

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  @Listen('keydown')
+  public handleKeyDown(evt: KeyboardEvent): void {
+    if (!this._options) {
+      return;
+    }
+
+    const enabledToggleOptions = this._options.filter((t) => !t.disabled);
+    const checked = enabledToggleOptions.findIndex((radio) => radio.checked);
```
jeripeierSBB, line 162:
```suggestion
    const checked = enabledToggleOptions.findIndex((toggleOption) => toggleOption.checked);
```

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+      this._options[0].tabIndex = 0;
+    }
+  }
+
+  private _getOptionTabIndex(option: HTMLSbbToggleOptionElement): number {
+    return option.checked && !option.disabled && !this.disabled ? 0 : -1;
+  }
+
+  @Listen('keydown')
+  public handleKeyDown(evt: KeyboardEvent): void {
```
jeripeierSBB, line 156: in unselected variant, I'd expect to let an toggleOption to be selected by space or enter key. what do you think?

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+    } else if (evt.key === nextKey || evt.key === 'ArrowDown') {
+      enabledToggleOptions[next].select();
+      enabledToggleOptions[next].focus();
+      evt.preventDefault();
+    }
+  }
+
+  public render(): JSX.Element {
+    return (
+      <Host class={{ 'sbb-toggle--checked': this._hasCheckedOption }}>
```
jeripeierSBB, line 193: I'm not sure whether we should set classes on host. @kyubisation ? On the other hand, it can be nice as we use :host for real properties and can toggle variables on the same level.

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+  private get _checked(): HTMLSbbToggleOptionElement {
+    return this._options.find((toggle) => toggle.checked);
+  }
+
+  @Listen('did-select', { passive: true })
+  public onToggleOptionSelect(event: CustomEvent<Set<string>>): void {
+    this.value = event.detail;
+  }
+
+  private _setCheckedPillPosition(): void {
```
jeripeierSBB, line 110: resizing the window does not stable update the pill position, so I often get into situation where the pill is calculated at the wrong time (probably). Can we make this more reliable?

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,200 @@
...
+    this._hasCheckedOption = true;
+    this._element.style.setProperty('--sbb-toggle-option-left', `${checked?.offsetLeft - 2}px`);
+    this._element.style.setProperty(
+      '--sbb-toggle-option-right',
+      `${this._toggleElement.clientWidth - (checked?.offsetLeft + checked?.clientWidth) - 2}px`
+    );
+  }
+
+  public connectedCallback(): void {
+    this._toggleResizeObserver.observe(this._element.firstElementChild);
```
jeripeierSBB, line 125:
```suggestion
    this._toggleResizeObserver.observe(this._toggleElement);
```

src/components/sbb-toggle/sbb-toggle.tsx
jeripeierSBB: See https://638f217768c512082108d28d--lyne-components-storybook.netlify.app/?path=/story/components-form-elements-toggle-sbb-toggle--dynamic-width

jeripeierSBB: Possible solution, maybe wait for another tick.

jeripeierSBB (changes requested):
Good impression, some details open

- is the empty file sbb-toggle-option.custom.d.ts really needed?

src/components/sbb-toggle-option/sbb-toggle-option.scss
```diff
@@ -0,0 +1,64 @@
...
+  overflow: hidden;
+  flex-grow: 1;
+  min-width: var(--sbb-spacing-fixed-16x);
+  z-index: 1;
+}
+
+:host(:focus-visible) {
+  @include focusOutline;
+
+  --sbb-focus-outline-offset: #{px-to-rem-build(6)};
```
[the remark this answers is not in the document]
jeripeierSBB: I asked manuel to add focus styles, here they are: https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=16063%3A477487&t=cSoQNp4yQgrcA5Y7-0
Maybe your implementation was correct before

src/components/sbb-toggle/sbb-toggle.tsx
kyubisation: I'm a bit late on this; I'd say to generally avoid setting classes on host.
I'm also not sure how this applies here, as a state without a selection does not exist (according to spec).

dauriamarco: Agreed with avoiding classes on the host, and true, it's probably no longer necessary in this case (previously we handled the unselected status anyway, but now we always select the first option if none is specified). 👍

src/components/sbb-toggle/sbb-toggle.scss
```diff
@@ -0,0 +1,94 @@
...
+  display: flex;
+  align-items: center;
+  width: var(--sbb-toggle-width);
+  max-width: 100%;
+  min-width: calc(var(--sbb-toggle-min-width) * 2);
+  height: var(--sbb-toggle-height);
+  user-select: none;
+  -webkit-tap-highlight-color: transparent;
+  background: var(--sbb-color-cloud-default);
+  border-radius: var(--sbb-border-radius-16x);
```
jeripeierSBB, line 62: As the border radius is used at least two times, I'd suggest to extract it into a variable.

src/components/sbb-toggle/sbb-toggle.scss
```diff
@@ -0,0 +1,94 @@
...
+
+  &::after {
+    content: '';
+    padding-inline: var(--sbb-toggle-padding-inline);
+    display: inline-block;
+    opacity: 1;
+    background-color: var(--sbb-toggle-selected-option-background-color);
+    border: var(--sbb-border-width-1x) solid var(--sbb-toggle-selected-option-border-color);
```
jeripeierSBB, line 70: As the border width is used at least two times, I'd suggest to extract it into a variable.

src/components/sbb-toggle/sbb-toggle.scss
```diff
@@ -0,0 +1,94 @@
...
+    inset-block: calc(-2 * var(--sbb-border-width-1x));
+    inset-inline-end: calc(var(--sbb-toggle-option-right) - #{sbb.px-to-rem-build(2)});
+    inset-inline-start: calc(var(--sbb-toggle-option-left) - #{sbb.px-to-rem-build(2)});
+    transition: {
+      duration: var(--sbb-toggle-animation-duration);
+      timing-function: ease;
+      property: opacity, inset-inline-end, inset-inline-start;
+    }
+
+    @include sbb.if-forced-colors {
```
jeripeierSBB, line 86: I'd suggest to just redefine --sbb-toggle-selected-option-border-color on most upper level (host)

src/components/sbb-toggle/sbb-toggle.e2e.ts
```diff
@@ -0,0 +1,13 @@
+import { newE2EPage } from '@stencil/core/testing';
```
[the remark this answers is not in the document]
jeripeierSBB: I have still the same meaning here ;)

src/components/sbb-toggle/sbb-toggle.spec.ts
```diff
@@ -0,0 +1,72 @@
+import { SbbToggle } from './sbb-toggle';
```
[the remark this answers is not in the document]
jeripeierSBB: I have still the same meaning here ;)

src/components/sbb-toggle/sbb-toggle.stories.js
```diff
@@ -0,0 +1,217 @@
...
+  labelTwo: 'Zurich',
+  iconName: undefined,
+};
+
+const DefaultTemplate = ({ label, labelTwo, iconName, ...args }) => (
+  <sbb-toggle {...args}>
+    <sbb-toggle-option icon-name={iconName} value="Value 1">
+      {label}
+    </sbb-toggle-option>
+    <sbb-toggle-option icon-name={iconName && 'arrows-right-left-small'} value="Value 2">
```
jeripeierSBB, line 94: what's the sense of `{iconName && 'arrows-right-left-small'}`?

src/components/sbb-toggle/sbb-toggle.stories.js
```diff
@@ -0,0 +1,217 @@
...
+
+const SlottedIconTemplate = ({ label, labelTwo, iconName, ...args }) => (
+  <sbb-toggle {...args}>
+    <sbb-toggle-option value="Value 1">
+      <sbb-icon slot="icon" name={iconName}></sbb-icon>
+      {label}
+    </sbb-toggle-option>
+
+    <sbb-toggle-option value="Value 2">
+      <sbb-icon slot="icon" name={iconName && 'arrows-right-left-small'}></sbb-icon>
```
jeripeierSBB, line 108: what's the sense of {iconName && 'arrows-right-left-small'}?

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+@Component({
+  shadow: true,
+  styleUrl: 'sbb-toggle.scss',
+  tag: 'sbb-toggle',
+})
+export class SbbToggle implements ComponentInterface {
+  /**
+   * Id of the toggle element.
+   */
+  @Prop() public sbbToggleId = `sbb-toggle-${++nextId}`;
```
jeripeierSBB, line 33: Seems not to be used at all -> remove it :)

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+   */
+  @Prop({ reflect: true }) public size?: InterfaceSbbToggleAttributes['size'] = 'm';
+
+  /**
+   * The value of the toggle.
+   */
+  @Prop({ mutable: true }) public value: any | null;
+
+  /**
+   * Whether the animation is enabled.
```
jeripeierSBB, line 56: Can you document, why it's necessary to have it mutable: true?

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+  private _setCheckedPillPosition(disableAnimation: boolean): void {
+    const options = this._options;
+    const checked = this._checked;
+
+    if (!checked) {
+      return;
+    }
+
+    if (!disableAnimation) {
+      this._element.removeAttribute('disable-animation');
```
jeripeierSBB, line 127: Is this part really needed?

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  /**
+   * The value of the toggle.
+   */
+  @Prop({ mutable: true }) public value: any | null;
+
+  /**
+   * Whether the animation is enabled.
+   */
+  @Prop({ reflect: true, mutable: true }) public disableAnimation = false;
```
jeripeierSBB, line 58: setting disableAnimation to true does not work

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -0,0 +1,202 @@
...
+      evt.preventDefault();
+    }
+  }
+
+  public render(): JSX.Element {
+    return (
+      <Host role="radiogroup">
+        <div
+          class="sbb-toggle"
+          // tabIndex={this.disabled ? -1 : 0}
```
jeripeierSBB, line 193: remove the line?

jeripeierSBB (changes requested):

src/components/sbb-toggle-option/sbb-toggle-option.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    this.didSelect.emit(this.value);
+  }
+
+  public render(): JSX.Element {
+    return (
+      <Host
+        // eslint-disable-next-line jsx-a11y/aria-proptypes
+        aria-checked={`${this.checked}`}
+        aria-disabled={this.disabled}
```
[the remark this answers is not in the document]
dauriamarco: `aria-checked` needs a string to work properly, I can also change it to `aria-checked={this.checked.toString()}` if you like?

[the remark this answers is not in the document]
jeripeierSBB: no it' ok. I'm just wondering why aria-disabled is not the same. maybe add a comment?

src/components/sbb-toggle/sbb-toggle.tsx
dauriamarco: This is needed otherwise the first click will not be animated.

src/components/sbb-toggle/sbb-toggle.tsx
```diff
@@ -188,12 +188,7 @@ export class SbbToggle implements ComponentInterface {
   public render(): JSX.Element {
     return (
       <Host role="radiogroup">
-        <div
-          class="sbb-toggle"
-          // tabIndex={this.disabled ? -1 : 0}
-          tabindex="-1"
-          ref={(toggle) => (this._toggleElement = toggle)}
-        >
+        <div class="sbb-toggle" tabindex="-1" ref={(toggle) => (this._toggleElement = toggle)}>
```
jeripeierSBB, line 183: Do you really need the reference _toggleELement as e.g. in connectedCallback you use directly "this._element.firstElementChild"?. Or would a getter be an option?

src/components/sbb-toggle-option/sbb-toggle-option.tsx
```diff
@@ -61,6 +68,8 @@ export class SbbToggleOption implements ComponentInterface {
 
   @Element() private _element!: HTMLElement;
 
+  private _toggleOptionAttributeObserver = new MutationObserver(() => this._onToggleOptionChange());
```
jeripeierSBB, line 71: Why is in this case a "@Watch" on the checked attribute not possible? It would be good, if we document the reason of the approach somehow.

src/components/sbb-toggle/sbb-toggle.scss
```diff
@@ -35,6 +35,11 @@
   --sbb-toggle-selected-option-border-color: var(--sbb-color-graphite-default);
 }
 
+:host([disabled='false']) {
```
jeripeierSBB, line 38: does this ever happen?

src/components/sbb-toggle/sbb-toggle.e2e.ts
```diff
@@ -1,13 +1,101 @@
-import { newE2EPage } from '@stencil/core/testing';
+import { E2EElement, E2EPage, newE2EPage } from '@stencil/core/testing';
```
jeripeierSBB, line 1: What about testing the new "checked" handling?

jeripeierSBB (changes requested):

jeripeierSBB (commented):
Another question regarding accessibility, should the option not have the accessibilityLabel property if you like to specify another text for screen readers instead of the label?

src/components/sbb-toggle-option/sbb-toggle-option.tsx
dauriamarco: I tried to use a `Watch('checked')` decorator but doing so would trigger a loop and break the page, with the MutationObserver I can check whether or not a `checked` attribute has simply been added to the DOM and call the `select()` method accordingly.

jeripeierSBB: Ok, I thought there must be something of this stuff ;) Just please comment in the code! thanks

src/components/sbb-toggle/sbb-toggle.scss
dauriamarco: It doesn't happen in our examples, but it could happen if the consumer sets the `disabled` attribute to 'false' (string). With this rule it would actually evaluate as `false`, otherwise it would still be as it were `true`. It's up to us to decide whether to cover or not this case, what do you think?

dauriamarco:
> Another question regarding accessibility, should the option not have the accessibilityLabel property if you like to specify another text for screen readers instead of the label?

I'm not sure, normally a radio would have an implicit label based on the visual content in order to avoid discrepancies between what the user sees and what the screen-reader reads. But we can consider this, perhaps we can also ask for other opinions?

src/components/sbb-toggle/sbb-toggle.tsx
dauriamarco: They're actually two different elements, `this._element.firstElementChild` should return the first toggle option, while the `this._toggleElement` is just the toggle element of course. I observe the `firstElementChild` (i.e. the option) in the connectedCallback in order to have more stability when setting the checked pill position (to solve that visual bug we noticed last time if you remember).

jeripeierSBB: I don't get how firstElementChild can be the first slotted option :D. I have to to get deeper later.

jeripeierSBB:
> > Another question regarding accessibility, should the option not have the accessibilityLabel property if you like to specify another text for screen readers instead of the label?
> 
> I'm not sure, normally a radio would have an implicit label based on the visual content in order to avoid discrepancies between what the user sees and what the screen-reader reads. But we can consider this, perhaps we can also ask for other opinions?

@kyubisation Can you also give your opinion here? Would probably also have an impact for radio button

src/components/sbb-toggle/sbb-toggle.scss
jeripeierSBB: I think we should not support this case. We assume, that a boolean property should be set as defined HTML standards. @kyubisation can you give your decision here? ;)
We also don't support it anywhere else (So far I know)

dauriamarco:
> > > Another question regarding accessibility, should the option not have the accessibilityLabel property if you like to specify another text for screen readers instead of the label?
> > 
> > 
> > I'm not sure, normally a radio would have an implicit label based on the visual content in order to avoid discrepancies between what the user sees and what the screen-reader reads. But we can consider this, perhaps we can also ask for other opinions?
> 
> @kyubisation Can you also give your opinion here? Would probably also have an impact for radio button

I found a paragraph on MDN that might help us as a reference:

_"Each radio element is labelled by its content, has a visible label referenced by aria-labelledby, or has a label specified with aria-label. The containing radiogroup element should either have a visible label referenced by aria-labelledby or a label specified with aria-label."_

[https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role](url)

jeripeierSBB:
> > > > Another question regarding accessibility, should the option not have the accessibilityLabel property if you like to specify another text for screen readers instead of the label?
> > > 
> > > 
> > > I'm not sure, normally a radio would have an implicit label based on the visual content in order to avoid discrepancies between what the user sees and what the screen-reader reads. But we can consider this, perhaps we can also ask for other opinions?
> > 
> > 
> > @kyubisation Can you also give your opinion here? Would probably also have an impact for radio button
> 
> I found a paragraph on MDN that might help us as a reference:
> 
> _"Each radio element is labelled by its content, has a visible label referenced by aria-labelledby, or has a label specified with aria-label. The containing radiogroup element should either have a visible label referenced by aria-labelledby or a label specified with aria-label."_
> 
> [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role](url)

So this would mean, adding a possibility to specify an aria-label?

dauriamarco:
> > I found a paragraph on MDN that might help us as a reference:
> >
> > _"Each radio element is labelled by its content, has a visible label referenced by aria-labelledby, or has a label specified with aria-label. The containing radiogroup element should either have a visible label referenced by aria-labelledby or a label specified with aria-label."_
> >
> > [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role](url)
> 
> So this would mean, adding a possibility to specify an aria-label?

Yes, it might be. But I think we can still interpret it in different ways and decide: it's either, "if there's a visual content, then the radio it's labelled by it, otherwise, provide aria-label (but we always have a visual content)*" or "it can be labelled by its content or by using aria-label (no matter if there's is already a visual label or not)".

*You could have radios without labels in case like the following image, but we don't expect to have similar scenarios:
image

src/components/sbb-toggle/sbb-toggle.tsx
kyubisation: @jeripeierSBB Do you mean `this._element.shadowRoot.firstElementChild`?

jeripeierSBB: Yes, we marco pointed me to it as well in teams :-)

jeripeierSBB (approved):
LGTM, thanks!

kyubisation (approved):
LGTM 👍

PR #1489, opened 2022-12-07
refactor: rename all mixins to kebab-case

jeripeierSBB:
BREAKING CHANGE:
In order to standardize mixin names, all mixin names were renamed to kebab-case style. E.g. `ifForcedColors` became `if-forced-colors`.


kyubisation (approved):
LGTM 👍 
Thanks! 😃

PR #1522, opened 2023-01-04
fix(sbb-title): fix spacing between sbb-title and paragraph

jeripeierSBB:


src/global/styles/typography.scss
```diff
@@ -47,3 +47,7 @@ html {
 .sbb-step-list {
   @include lists.step-list;
 }
+
+.sbb-title + p {
```
kyubisation, line 51: Can you add a comment about why we do this?

kyubisation (changes requested):

src/global/styles/typography.scss
```diff
@@ -51,3 +51,10 @@ html {
 .sbb-legend {
   @include typo.legend;
 }
+
+// In smaller title font-sizes, the space after the title is smaller than the default paragraph space before.
+// Due to margin collapsing, the paragraph space wrongly wins.
+// To prevent the too large gap, we reset the paragraph space.
```
kyubisation, line 57:
```suggestion
// Due to margin collapsing, the wrong paragraph space wins.
// To prevent the mistakenly large gap, we reset the paragraph space.
```

PR #1527, opened 2023-01-09
Feat/sbb time input

### LINKED ISSUE #1488: story(sbb-time-input): initial implementation
### Feature Description

We want to provide a time input. This does not include `<input type="time">`, which is a separate story.

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=16087%3A487158&t=GPgMuAnvkWt2Vhcx-1)
[See Angular implementation](https://github.com/sbb-design-systems/sbb-angular/blob/main/src/angular/time-input/time-input.ts)
[See usage for the angular library](https://angular.app.sbb.ch/angular/components/time-input/examples)

**Design Spec**
- IMPORTANT! The design in figma includes the form field. The implementation should not. (But it should support being placed inside a `sbb-form-field`.)
- The size inside an `sbb-fom-field` should stretch like a normal input to the size of a form field.
  - Add property @Prop({reflect:true}) width: 'default' | 'collapse'  = 'default'; to `sbb-form-field`. Default is the current behavior, collapse means that the width adapts to the inner content, which means just not adding a width or min-width.
- Time format is 00:00 to 24:00 (no am/pm allowed)
- Wrong inputs are "allowed" (large numbers, letters, etc.)

**Technical Spec**
- Add properties
  - `placeholder` = `HH:MM`
  - `value`
  - `valueAsDate` which returns a date object with applied time (e.g. 1.1.1970 22:00)
  - `required`
  - `disabled`
  - `readonly`
- User inputs should be validated like in Angular implementation (See link above) and content should be updated accordingly. e.g. input of `123` should be displayed as `01:23`. This should trigger a change event.

```html
<sbb-time-input value="value"></sbb-time-input>
```

```html
<sbb-form-field width="collapse">
  <sbb-time-input value="value"></sbb-time-input>
</sbb-form-field>
```

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

DavideMininni-Fincons:
## Preflight Checklist

## Issue

This PR Closes #1488 

## Pull request checklist

## Changes

Changes in this pull request:

- add sbb-time-input component
- minor improvement in sbb-form-field

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [ ] Safari Desktop
- [ ] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

- [ ] JAWS Firefox Desktop
- [ ] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [ ] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/sbb-time-input/sbb-time-input.tsx
```diff
@@ -0,0 +1,157 @@
...
+    } else if (value) {
+      return value.match(REGEX_GROUPS_WITH_COLON);
+    } else {
+      return null;
+    }
+  }
+
+  /** Parse the hours with two chars. */
+  private _parseHour(regGroupHours: string): string {
+    return regGroupHours.length > 1 ? regGroupHours : '0' + regGroupHours;
```
kyubisation, line 124: I believe this could be achieved by `regGroupHours.padStart(2, '0')`

src/components/sbb-time-input/sbb-time-input.tsx
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  /** Parse the minutes with two chars. */
+  private _parseMinute(regGroupMin: string): string {
+    regGroupMin = regGroupMin || '00';
```
kyubisation, line 129: I believe this could be achieved by `(regGroupMin || '').padStart(2, '0')`

src/components/sbb-time-input/sbb-time-input.tsx
```diff
@@ -0,0 +1,157 @@
...
+  shadow: true,
+  styleUrl: 'sbb-time-input.scss',
+  tag: 'sbb-time-input',
+})
+export class SbbTimeInput implements ComponentInterface, AccessibilityProperties {
+  /** Value for the inner HTMLInputElement. */
+  @Prop({ mutable: true }) public value?: string = '';
+
+  /** Date value with the given time for the inner HTMLInputElement. */
+  @Prop({ mutable: true }) public valueAsDate?: Date;
```
kyubisation, line 31: I'd like to have parity with the native time input and have `null` as the default value

src/components/sbb-time-input/sbb-time-input.tsx
```diff
@@ -0,0 +1,157 @@
...
+const REGEX_GROUPS_WO_COLON = /([0-9]{1,2})([0-9]{2})/;
+
+@Component({
+  shadow: true,
+  styleUrl: 'sbb-time-input.scss',
+  tag: 'sbb-time-input',
+})
+export class SbbTimeInput implements ComponentInterface, AccessibilityProperties {
+  /** Value for the inner HTMLInputElement. */
+  @Prop({ mutable: true }) public value?: string = '';
```
kyubisation, line 28: Have a `@Watch` that checks if the value is valid and reset to empty string, if it is not

src/components/sbb-time-input/sbb-time-input.tsx
kyubisation, line 31: Have a `@Watch` that checks if the value is valid and reset to null, if it is not

kyubisation (changes requested):
Nice work! 👍

src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -98,6 +99,12 @@
   --sbb-form-field-background-color: var(--sbb-color-white-default);
 }
 
+:host([width='collapse']) {
+  max-width: min(#{sbb.px-to-rem-build(300)}, 100%);
```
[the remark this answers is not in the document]
DavideMininni-Fincons: Yes; since the default width rule at line 31 is overridden to `width: auto`, without max-width rule the input width can stretch over that value.

src/components/sbb-time-input/sbb-time-input.tsx
DavideMininni-Fincons: "Valid" means that passes the validation in the `_validateInput` function, or should there be other logic?

src/components/sbb-form-field/sbb-form-field.scss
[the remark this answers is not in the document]
kyubisation: I mean whether it is necessary to define a max-width at all. At least I don't think so.
Or is there a use case we want to match?

src/components/sbb-time-input/sbb-time-input.tsx
kyubisation: Yes, according to `_validateInput` 👍

src/components/sbb-time-input/sbb-time-input.scss
```diff
@@ -0,0 +1,24 @@
+@use '../../global/styles' as sbb;
+
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin
+@include sbb.host-component-properties;
+
+:host {
+  --sbb-time-input-max-width: #{sbb.px-to-rem-build(42)};
```
kyubisation, line 8: @mcilurzo changed the dimension, as the placeholder was too wide for `HH:MM`
It is now 58px and 65px

DavideMininni-Fincons:
Note: the Storybook's date control is broken, no way of make it works; if time is changed, dateAsValue always stays 'null' in the 'render' function. :|

kyubisation:
Ok, that's fine. Just remove it from the storybook.
I just noticed, the documentation is slightly off due to my change requests.

kyubisation (approved):
LGTM 👍 
Thank you very much!

src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -97,6 +98,11 @@
   --sbb-form-field-background-color: var(--sbb-color-white-default);
 }
 
+:host([width='collapse']) {
```
jeripeierSBB, line 101: I'm not sure whether we should change the notation to the following. It would mean that we never override a definition and it's maybe more consumer friendly. what do you think @kyubisation?
```suggestion
:host(:where(:not([width='collapse']))) {
  min-width: #{sbb.px-to-rem-build(150)};
  width: min(#{sbb.px-to-rem-build(300)}, 100%);
}
```

src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -119,6 +125,7 @@
 ::slotted(*[slot='suffix']) {
   min-width: var(--sbb-icon-svg-width);
   display: flex;
+  color: var(--sbb-form-field-prefix-color);
 }
```
jeripeierSBB, line 129: It's strange for me that a "prefix" color is applied to the "suffix" and not anymore to the "prefix"?

jeripeierSBB: Also, comparing to figma, this seems to introduce a bug.

src/components/sbb-form-field/readme.md
```diff
@@ -53,6 +54,12 @@ the `prefix` and `suffix` slots.
 </sbb-form-field>
 ```
 
+### Width
+
+By default, the component has a defined width and min-width. However, this behavior can be overridden by setting 
+the `width` property to `collapse`: in this way the component adapts its width to the inner slotted input component.
+This is useful, for example, for the `sbb-time-input` component.
```
jeripeierSBB, line 61:
```suggestion
This is useful, for example, for the `sbb-time-input` component. 
However, as the width-styles are exposed to the host, it's possible to apply any desired width by setting just the `width` and `min-width` css properties.
```

src/components/sbb-time-input/sbb-time-input.scss
```diff
@@ -0,0 +1,24 @@
...
+  display: inline-block;
+}
+
+input {
+  @include sbb.input-reset;
+
+  max-width: var(--sbb-time-input-max-width);
+  font-size: var(--sbb-font-size-text-m);
+  line-height: var(--sbb-typo-line-height-body-text);
+  color: var(--sbb-form-field-text-color);
```
jeripeierSBB, line 23: This only works if used with sbb-form-field, but maybe also the desired behavior. If so, I suggest to just add a comment.

jeripeierSBB (changes requested):
Some little open points / questions in my eyes, but generally looking nice! :)

src/components/sbb-time-input/sbb-time-input.scss
DavideMininni-Fincons: Yes, the color rule doesn't work, e.g. the disabled state is not shown with gray text if the component is NOT used within a sbb-form-field. https://63ca6f1f40bf1c0084a928a7--lyne-components-storybook.netlify.app/?path=/story/components-form-elements-sbb-time-input--sbb-time-input&args=disabled:true

@kyubisation Are there use cases with the time-input without form-field? If yes as i think, maybe I have to add here the rules for the color handling..

src/components/sbb-form-field/sbb-form-field.scss
kyubisation: Discussed with Teams. I agree with the proposal.

src/components/sbb-time-input/sbb-time-input.scss
kyubisation: We can ignore the "standalone" variant for now. Might make sense to also remove the story for it.

jeripeierSBB (approved):
LGTM

PR #1531, opened 2023-01-10
fix: boolean properties handling

jeripeierSBB:
Previously only setting an attribute (e.g. "disabled") on an element was working. Setting `disabled="false"` which is not removing the entire attribute, had no effect. The HTML Standard says that the previous behavior was correct, but Stencil also allows setting `disabled="false"`. Therefore we enabled support also for setting a property by a specific boolean value.


kyubisation (approved):
LGTM 👍

dauriamarco (approved):
LGTM 👍

PR #1574, opened 2023-01-27
refactor: improve overlay components styles and state handling

dauriamarco:
## Preflight Checklist

## Issue

This PR Closes #

## Pull request checklist

## Changes

Changes in this pull request:

- improve `sbb-dialog` styling
- improve `sbb-menu` styling and state handling
- improve `sbb-tooltip` styling and state handling
- improve `sbb-navigation` styling and state handling
- improve `sbb-navigation-section` styling and state handling
- handle `display` property on not yet hydrated components
- add `role` attribute to `sub-button` and `sbb-tooltip-trigger`
- add aria attributes to trigger elements

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [ ] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [ ] JAWS Firefox Desktop
- [ ] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [ ] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

- [ ] Yes
- [x] No

### DIFF
--- a/src/components/sbb-dialog/sbb-dialog.scss
+++ b/src/components/sbb-dialog/sbb-dialog.scss
@@ -27,6 +27,9 @@
   --sbb-scrollbar-color: var(--sbb-color-black-alpha-30);
   --sbb-scrollbar-color-hover: var(--sbb-color-black-alpha-60);
 
+  position: absolute;
+  inset: 0 auto auto 0;
+
   @include sbb.mq($from: micro) {
     --sbb-dialog-padding-inline: var(--sbb-spacing-fixed-6x);
   }
--- a/src/components/sbb-dialog/sbb-dialog.scss
+++ b/src/components/sbb-dialog/sbb-dialog.scss
@@ -27,6 +27,9 @@
   --sbb-scrollbar-color: var(--sbb-color-black-alpha-30);
   --sbb-scrollbar-color-hover: var(--sbb-color-black-alpha-60);
 
+  // We use this rule to make the inner container element to appear as if it were a
+  // direct child of the host's parent element. This is useful because the host
+  // should be ignored when using CSS grid or similar layout techniques.
   display: contents;
 
   @include sbb.mq($from: micro) {
--- a/src/components/sbb-dialog/sbb-dialog.scss
+++ b/src/components/sbb-dialog/sbb-dialog.scss
@@ -13,37 +13,19 @@
   --sbb-dialog-max-width: 100%;
   --sbb-dialog-max-width-default: #{sbb.px-to-rem-build(892)};
   --sbb-dialog-max-height: 100%;
+  --sbb-dialog-inset: 0 auto auto 0;
   --sbb-dialog-border-radius: var(--sbb-border-radius-8x);
   --sbb-dialog-padding-block: var(--sbb-spacing-responsive-s);
   --sbb-dialog-animation-duration: var(--sbb-animation-duration-6x);
   --sbb-dialog-animation-easing: ease;
+  --sbb-dialog-pointer-events: none;
+  --sbb-dialog-backdrop-visibility: hidden;
+  --sbb-dialog-backdrop-pointer-events: none;
   --sbb-dialog-backdrop-color: transparent;
   --sbb-dialog-header-padding-block: var(--sbb-spacing-responsive-s) 0;
   --sbb-dialog-footer-border: var(--sbb-border-width-1x) solid var(--sbb-color-cloud-default);
   --sbb-scrollbar-color: var(--sbb-color-black-alpha-30);
   --sbb-scrollbar-color-hover: var(--sbb-color-black-alpha-60);
-
-  position: fixed;
-  pointer-events: none;
-  display: flex;
-  align-items: center;
-  inset: 0;
-  z-index: 10;
-
-  // Backdrop
-  &::before {
-    content: '';
-    visibility: hidden;
-    pointer-events: none;
-    position: fixed;
-    inset: 0;
-    background-color: var(--sbb-dialog-backdrop-color);
-    transition: {
-      duration: var(--sbb-dialog-animation-duration);
-      timing-function: var(--sbb-dialog-animation-easing);
-      property: background-color, visibility;
-    }
-  }
 
   @include sbb.mq($from: micro) {
     --sbb-dialog-padding-inline: var(--sbb-spacing-fixed-6x);
@@ -64,15 +46,12 @@
 }
 
 :host(:is([data-state='opened'], [data-state='opening'])) {
-  pointer-events: all;
+  --sbb-dialog-pointer-events: all;
 
   @include sbb.mq($from: medium) {
+    --sbb-dialog-backdrop-visibility: visible;
+    --sbb-dialog-backdrop-pointer-events: all;
     --sbb-dialog-backdrop-color: var(--sbb-color-milk-default);
-
-    &::before {
-      visibility: visible;
-      pointer-events: all;
-    }
   }
 }
 
@@ -105,6 +84,34 @@
 
 :host([data-overflows]:not([data-fullscreen], [negative]:not([negative='false']))) {
   --sbb-dialog-footer-border: none;
+}
+
+:host(:not([data-state='closed'])) {
+  --sbb-dialog-inset: 0;
+}
+
+.sbb-dialog__container {
+  position: fixed;
+  pointer-events: var(--sbb-dialog-pointer-events);
+  display: flex;
+  align-items: center;
+  inset: var(--sbb-dialog-inset);
+  z-index: 10;
+
+  // Dialog backdrop (not visible on mobile)
+  &::before {
+    content: '';
+    visibility: var(--sbb-dialog-backdrop-visibility);
+    pointer-events: var(--sbb-dialog-backdrop-pointer-events);
+    position: fixed;
+    inset: var(--sbb-dialog-inset);
+    background-color: var(--sbb-dialog-backdrop-color);
+    transition: {
+      duration: var(--sbb-dialog-animation-duration);
+      timing-function: var(--sbb-dialog-animation-easing);
+      property: background-color, visibility;
+    }
+  }
 }
 
 .sbb-dialog {
--- a/src/components/sbb-dialog/sbb-dialog.scss
+++ b/src/components/sbb-dialog/sbb-dialog.scss
@@ -27,8 +27,7 @@
   --sbb-scrollbar-color: var(--sbb-color-black-alpha-30);
   --sbb-scrollbar-color-hover: var(--sbb-color-black-alpha-60);
 
-  position: absolute;
-  inset: 0 auto auto 0;
+  display: contents;
 
   @include sbb.mq($from: micro) {
     --sbb-dialog-padding-inline: var(--sbb-spacing-fixed-6x);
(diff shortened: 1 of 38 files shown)


src/components/sbb-menu/sbb-menu.scss
```diff
@@ -118,16 +125,16 @@
...
 
   @include sbb.mq($from: medium) {
-    top: 0;
-    left: 0;
-    right: unset;
-    bottom: unset;
+    inset-block-start: 0;
+    inset-inline-start: 0;
+    inset-inline-end: unset;
+    inset-block-end: unset;
```
kyubisation, line 131: Nitpick; Would this be equal?
```suggestion
    inset: 0 unset unset 0;
```

src/components/sbb-dialog/sbb-dialog.scss
```diff
@@ -13,37 +13,22 @@
...
-    position: fixed;
-    inset: 0;
-    background-color: var(--sbb-dialog-backdrop-color);
-    transition: {
-      duration: var(--sbb-dialog-animation-duration);
-      timing-function: var(--sbb-dialog-animation-easing);
-      property: background-color, visibility;
-    }
-  }
+  position: absolute;
```
jeripeierSBB, line 30: does this really work in case e.g. some outer element in the DOM has position: relative on it?
Maybe it helps if we just don't apply any display property to the host, so it has also no space in view. This comment would affects also all other dialog components.

src/components/sbb-navigation-section/sbb-navigation-section.scss
```diff
@@ -17,8 +17,8 @@
   --sbb-focus-outline-color: var(--sbb-focus-outline-color-dark);
   --sbb-navigation-section-width: 100%;
 
-  // We need to handle here the placement within the grid, since the host is exposed to the parent's grid (sbb-navigation)
-
+  // We need to handle here the placement and styling within the grid, since the host
```
jeripeierSBB, line 20: Did you test like display: unset or not including sbb.host-component-properties at all? Then maybe the grid styles also work if applied to an inner element.

src/components/sbb-navigation/sbb-navigation.stories.js
```diff
@@ -15,18 +15,17 @@ const playStory = async ({ canvasElement }) => {
   );
 
   const button = canvas.getByTestId('navigation-trigger');
-  await userEvent.click(button);
-  await waitFor(() =>
-    expect(
-      canvas.getByTestId('navigation').classList.contains('sbb-navigation--opened')
-    ).toBeTruthy()
-  );
+  userEvent.click(button);
```
jeripeierSBB, line 18: I'd suggest to leave the check in this method as it's like a unit test asserting that navigation was really opened.

src/components/sbb-navigation/sbb-navigation.stories.js
```diff
@@ -221,7 +220,7 @@ const WithNavigationSectionTemplate = (args) => [
 export const Default = DefaultTemplate.bind({});
 Default.argTypes = basicArgTypes;
 Default.args = { ...basicArgs };
-Default.play = isChromatic() && playStory;
+Default.play = playStory;
```
jeripeierSBB, line 223: Don't forget to re-add isChromatic() &&

src/components/sbb-navigation/sbb-navigation.scss
```diff
@@ -75,27 +55,22 @@
...
 
-:host(.sbb-navigation--opened) {
-  --sbb-navigation-animation-easing: ease-out;
+:host(:not([data-state='closed'])) {
+  --sbb-navigation-inset: 0;
+  --sbb-navgation-transform: translateX(0);
 }
 
-:host(.sbb-navigation--has-navigation-section) {
+:host([data-has-navigation-section]) {
```
jeripeierSBB, line 73: I think the way you implemented it on tsx file, it should also handle the false case here? (:not([data-has-navigation-section="false"])

src/components/sbb-menu/sbb-menu.scss
dauriamarco: Unfortunately, this shorthand would not work because apparently 'unset' is not supported there. Where possible I replaced it with something like `inset: 0 auto auto 0`, so I can try using this but I'm not sure if it has the same effect.

src/components/sbb-menu/sbb-menu.scss
```diff
@@ -86,10 +93,10 @@
...
   text-align: start;
   position: var(--sbb-menu-position);
-  left: 0;
-  bottom: 0;
-  top: unset;
-  right: unset;
+  inset-inline-start: 0;
+  inset-block-start: unset;
+  inset-block-end: 0;
+  inset-inline-end: unset;
```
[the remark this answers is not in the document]
dauriamarco: Unfortunately, this shorthand would not work because apparently 'unset' is not supported there. Where possible I replaced it with something like `inset: 0 auto auto 0`, so I can try using this but I'm not sure if it has the same effect.

src/components/sbb-navigation/sbb-navigation.scss
dauriamarco: This shouldn't actually happen, as this attribute is added/removed from within the component and should not be exposed to the consumer. However, we can add the rule anyway if you prefer.

src/components/sbb-menu/sbb-menu.scss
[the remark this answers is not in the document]
kyubisation: No, then it is fine as it is. I was not aware of that limitation. 😞

src/components/sbb-navigation/sbb-navigation.scss
jeripeierSBB: I'm surprised that it is removed like it's implemented :D

jeripeierSBB: but yes when it's working internally, no other case should be checked

src/global/interfaces/link-button-properties.spec.ts
```diff
@@ -169,6 +169,9 @@ describe('getLinkRenderVariables', () => {
         rel: 'external noopener nofollow',
         tabIndex: '-1',
       },
+      hostAttributes: {
+        role: 'link',
```
jeripeierSBB, line 173: I'm insecure whether there is any case of assigning role=link?

jeripeierSBB (commented):
- What about setting role=button only when it's needed (from overlay component via triggerElement) and if not, should then not every action element has the hostAttributes?
- I think there are more elements which a role make sense, e.g. the header action where the header menu is triggered

src/components/sbb-tooltip/sbb-tooltip.tsx
```diff
@@ -65,7 +70,12 @@ export class SbbTooltip implements ComponentInterface {
   /**
    * The state of the tooltip.
    */
-  @State() private _state: 'closed' | 'opening' | 'opened' | 'closing' = 'closed';
+  private set _state(state: SbbTooltipState) {
```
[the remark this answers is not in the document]
dauriamarco: Yes, I still use this approach in this case because otherwise it would break the appearance of the component by incorrectly positioning the tooltip arrow.

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -191,6 +205,13 @@ export class SbbMenu implements ComponentInterface {
       return;
     }
 
+    this._triggerElement.setAttribute('aria-haspopup', 'menu');
+    this._triggerElement.setAttribute('aria-controls', this._element.id || this._menuId);
```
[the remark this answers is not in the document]
dauriamarco: True, but I need this logic here because in the `connectedCallback` I still don't have access to the potential auto-generated id assigned in the ref function.

dauriamarco:
> * What about setting role=button only when it's needed (from overlay component via triggerElement) and if not, should then not every action element has the hostAttributes?
> * I think there are more elements which a role make sense, e.g. the header action where the header menu is triggered

Yes, I agree that every action component should have `hostAttributes` (or `role="button"` where only a button could be rendered). If you agree, I'm going to add it the following components:
- `sbb-header-action`
- `sbb-navigation-action`
- `sbb-menu-action`

Let me know if I'm forgetting other action components.

src/components/sbb-menu/sbb-menu.tsx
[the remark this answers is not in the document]
jeripeierSBB: I see I see ;)

jeripeierSBB (approved):
LGTM, thanks!

src/components/sbb-menu/sbb-menu.scss
```diff
@@ -13,53 +13,42 @@
...
 
   // 85vh is not an exact value but looks optimized for mobile view.
   --sbb-menu-max-height: calc(85vh - var(--sbb-spacing-fixed-8x));
   --sbb-menu-min-height: #{sbb.px-to-rem-build(120)};
   --sbb-menu-border-radius: var(--sbb-border-radius-4x);
+  --sbb-menu-visibility: hidden;
+  --sbb-menu-pointer-events: none;
   --sbb-menu-backdrop-color: transparent;
 
+  display: contents;
```
kyubisation, line 26: Maybe document why this is necessary?

src/components/sbb-navigation/sbb-navigation.tsx
```diff
@@ -177,6 +188,12 @@ export class SbbNavigation implements ComponentInterface, AccessibilityPropertie
 
   // Check if the trigger is valid and attach click event listeners.
   private _configure(trigger: string | HTMLElement): void {
+    if (this._triggerElement) {
```
kyubisation, line 191: Maybe extract to a re-usable function?

src/components/sbb-navigation/sbb-navigation.tsx
```diff
@@ -193,6 +210,13 @@ export class SbbNavigation implements ComponentInterface, AccessibilityPropertie
       return;
     }
 
+    this._triggerElement.setAttribute('aria-haspopup', 'menu');
```
kyubisation, line 213: Maybe extract to a re-usable function?

kyubisation (approved):
LGTM 👍 
Nice work!
I added a few comments about potential improvements.

kyubisation (approved):
LGTM 👍

src/global/helpers/overlay-trigger-attributes.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Add meaningful aria attributes to trigger elements that are connected to
+ * popup elements such as menus and dialogs.
+ */
+export function setAriaOverlayTriggerAttributes(
+  trigger: HTMLElement,
+  popupType: 'menu' | 'dialog',
+  popupId: string,
```
jeripeierSBB, line 8: overlayId?

src/global/helpers/overlay-trigger-attributes.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Add meaningful aria attributes to trigger elements that are connected to
+ * popup elements such as menus and dialogs.
```
jeripeierSBB, line 3: *overlay elements

PR #1577, opened 2023-01-30
fix(sbb-navigation-section): place button always in new row

### LINKED ISSUE #1575: Wrong positioning of the button in the navigation
## Preflight Checklist

- [x] I have searched the issue tracker for an issue that matches the one I want to file, without success.

## Which package(s) are you using?

- [x] `lyne-components`
- [ ] `non package issue`

## Environment

> Operating system (e.g. Windows 10)
mac OS

> Browser / browser version (e.g. Mozilla Firefox 74.0 (64-bit))
All

## Detailed description

> Describe in detail the issue you are having.
The button in the navigation is positioned incorrectly. The button should always appear after the menu items and not be part of the grid.

> What did you expect to happen? What happened instead? What would you like to
> see changed?

> Which [version](https://lyne-documentation.netlify.app/deployments/) of _Lyne Components_ are you using?
master, 0.23.11

> Is the issue related to a specific component? If so please provide the components Storybook link of the corresponding version.
navigation-component

## Steps to reproduce the issue

1. Step: Open Navigation
2. Step: Select a Top-Level-Item
3. Step: Remove a sbb-navigation-list to see the wrong positioning. the bug is only visible if there are not 3 or 6 sbb-navigation-lists in the second level.

## Additional information
bug-navigation

jeripeierSBB:
This fix only works if there is exactly one button and if it is an "sbb-button".

Closes #1575

### DIFF
--- a/src/components/sbb-navigation-section/sbb-navigation-section.stories.js
+++ b/src/components/sbb-navigation-section/sbb-navigation-section.stories.js
@@ -135,7 +135,6 @@ const DefaultTemplate = (args) => [
 
       {navigationList('Label')}
       {navigationList('Label')}
-      {navigationList('Label')}
 
       <sbb-button size="m" variant="secondary" style="width: fit-content" sbb-navigation-close>
         Close navigation
--- a/src/components/sbb-navigation-section/sbb-navigation-section.scss
+++ b/src/components/sbb-navigation-section/sbb-navigation-section.scss
@@ -32,6 +32,11 @@
     padding-inline-start: var(--sbb-navigation-section-content-padding-inline-start);
   }
 
+  // Always place the sbb-button on a new row
+  ::slotted(sbb-button) {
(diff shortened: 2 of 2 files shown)


src/components/sbb-navigation-section/sbb-navigation-section.stories.js
```diff
@@ -133,7 +133,6 @@ const DefaultTemplate = (args) => [
       {navigationList('Label')}
       {navigationList('Label')}
 
-      {navigationList('Label')}
```
dauriamarco, line 136: What happens if we add some columns after the button? Does it render as expected in the image below, for example? Or should they simply not be added in this way?
image

dauriamarco (commented):
Thank you! I just have a question about a possible side case:

dauriamarco (approved):
LGTM 👍

PR #1583, opened 2023-01-31
fix(typography): fix name of `sbb-text--bold` CSS class

jeripeierSBB:
BREAKING CHANGE:
Renamed CSS class `sbb-text-bold` to `sbb-text--bold` to match BEM naming.


DavideMininni-Fincons (approved):
Looks good, thanks!

PR #1588, opened 2023-02-01
fix(sbb-card): fix main slot not being rendered and animation on windows

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
Maybe add comment about why the JSX change is necessary

Issue #1368, opened 2022-09-02
story(sbb-action-group): initial implementation


kyubisation:
### Feature Description

We want to provide the sbb-action-group component.

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=7965%3A296702)

**Design Spec**
- The `<sbb-action-group>` should support both `<sbb-button>` and `<sbb-link>`

**Technical Spec**
- Use CSS Flexbox (with `gap`, etc.)
- Provide property `orientation: 'horizontal' | 'vertical' = 'horizontal'`
- Provide property `horizontalFrom?: 'zero' | 'micro' | 'small' | 'medium' | 'large' | 'wide' | 'ultra'`, which overrides the `orientation` property
- Provide property `align: 'start' | 'center' | 'end' = 'start'` (To discuss; Add 'stretch' as a differentiation to 'center')
- Support attribute `align-self="start"` `align-self="end"` on `<sbb-button>` or `<sbb-link>` to allow a single instance to be moved contrary to the group alignment

```html
<sbb-action-group>
  <sbb-button variant="secondary">Action 1</sbb-button>
  <sbb-button>Action 2</sbb-button>
</sbb-action-group>
```

```html
<sbb-action-group align="end">
  <sbb-button align-self="start" variant="secondary">Action 1</sbb-button>
  <sbb-button variant="secondary">Action 2</sbb-button>
  <sbb-button>Action 3</sbb-button>
</sbb-action-group>
```

```html
<sbb-action-group orientation="vertical">
  <sbb-button variant="secondary">Action 1</sbb-button>
  <sbb-button variant="secondary">Action 2</sbb-button>
  <sbb-button>Action 3</sbb-button>
</sbb-action-group>
```

### Definition of Done

 - [ ]  Component is implemented
 - [ ]  Documentation is provided
 - [ ]  Tests are implemented
 - [ ]  E2E Tests are implemented
 - [ ]  Storybook stories are implemented
 - [ ]  Navigation via keyboard is tested
 - [ ]  Screen reader output is tested
 - [ ]  High-contrast is tested
 - [ ]  Remaining accessibility is tested
 - [ ]  UX approved

### Definition of Done

- [ ] Component is implemented
- [ ] Documentation is provided
- [ ] Tests are implemented
- [ ] E2E Tests are implemented
- [ ] Storybook stories are implemented
- [ ] Navigation via keyboard is tested
[...]

PR #1606, opened 2023-02-08
fix: provide CSS variables to set `z-index` for overlay components

dauriamarco:
Provide the possibility to manually set the `z-index` property for all dialog based components. This is necessary since we use the `show()` method instead of the `showModal()` method (see more here #1368), which would otherwise place the dialog on top of everything else in the web page, regardless of the z-index of other elements.

### DIFF
--- a/src/components/sbb-navigation/sbb-navigation.scss
+++ b/src/components/sbb-navigation/sbb-navigation.scss
@@ -97,7 +97,7 @@
   inset: var(--sbb-navigation-inset);
   pointer-events: none;
   overflow: hidden;
-  z-index: var(--sbb-navigation-z-index, 1000);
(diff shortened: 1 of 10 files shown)


src/components/sbb-navigation/sbb-navigation.scss
```diff
@@ -181,7 +181,7 @@
   width: var(--sbb-navigation-expanded-width);
   pointer-events: none;
   padding: var(--sbb-spacing-responsive-xs);
-  z-index: 11;
+  z-index: calc(var(--sbb-navigation-z-index, 1000) + 1);
```
jeripeierSBB, line 184: nice! :)

jeripeierSBB (changes requested):
We should create a global css var

jeripeierSBB (approved):
LGTM

PR #1618, opened 2023-02-13
fix: fix hydrated loading state hiding

jeripeierSBB:


src/global/styles/typography.scss
```diff
@@ -16,6 +16,11 @@ html {
   color: var(--sbb-color-metal-default);
 }
 
+// Hide components as long as they are instantiated
```
dauriamarco, line 19: I think it would be more like "Hiding components until they are instantiated" (?) if I understand correctly. Great btw 😃👍

dauriamarco (approved):
LGTM 👍

src/global/styles/typography.scss
jeripeierSBB, line 19:
```suggestion
// Hiding components until they are instantiated
```

kyubisation (approved):
LGTM 👍

PR #1619, opened 2023-02-13
fix(sbb-header): fix hiding transition

jeripeierSBB:


src/components/sbb-header/sbb-header.scss
```diff
@@ -34,8 +33,13 @@
   --sbb-header-inset-block-start: calc(var(--sbb-header-height) * -1);
 }
 
+:host([hide-on-scroll][data-fixed][data-animated]) {
```
kyubisation, line 36: Nitpick; It is not immediatelly clear to me _when_ this is applied? Could this maybe be documented?

kyubisation (approved):
LGTM 👍

PR #1630, opened 2023-02-20
feat(sbb-datepicker): component implementation

### LINKED ISSUE #1244: story(sbb-datepicker): initial implementation
### Feature Description

We want to provide a datepicker

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=140%3A10)

**References**

- 12427
- https://angular.app.sbb.ch/angular/components/datepicker/overview
- https://material.angular.io/components/datepicker/overview
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

**Design Spec**

- The sbb-calendar can optionally be displayed with two months, however on zero size it is always only one month

**Technical Spec**

- Extract date handling functions into own class (see e.g. https://github.com/sbb-design-systems/sbb-angular/blob/main/src/angular/core/datetime/native-date-adapter.ts)

- `sbb-calendar`
  - Use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat for date formats/names
  - `@Prop() wide: boolean = false` to indicate whether the datepicker should be displayed `wide` with two months
  - Should support properties `min` and `max` which should support any form of date input (ISO string, Unix Timestamp)
  - Provide `@Prop() dateFilter: (Date | null) => boolean` to define which dates are available (true) or disabled (false)
  - Should emit an event `date-selected` when a date is selected
  - Should `sbb-calendar` support ranges? If so, how? @mcilurzo 
  - Should have internal buttons to switch months
  - Should support navigation via arrow keys (How should behavior be for two months? @mcilurzo )
    - The arrow keys should be usable to switch between dates with wrap-around when at the end of the week and switching months if at start/end of month
  - The calendar should be rendered as a native table and each day should be a native button due to accessibility
  - Compare aria attributes with existing implementations

- `sbb-datepicker`
  - Should internally render an `sbb-calendar` inside an `sbb-tooltip`
  - Should have property `input`, which accepts a string id or an input HTMLElement
    - If it is not set, resolve the input by looking for `sbb-form-field` parent and checking its children
  - Should have properties `min`/`max` which accept `string` | `number` | `Date`, which is forwarded to the `sbb-calender`
    - The datepicker should read `min`/`max` values from input, if `min`/`max` is not declared on `sbb-datepicker`
  - The month buttons should become disabled if min and max date are set and clicking them would navigate to before/after min/max date
  - Provide `@Prop() dateFilter: (Date | null) => boolean` and forward to `sbb-calendar`
  - The datepicker should open when clicking on the toggle

- `sbb-datepicker-previous-day`/`sbb-datepicker-next-day`/`sbb-datepicker-toggle`
  - Should have property `datepicker`, which accepts a string id or an `sbb-datepicker` HTMLElement
    - If it is not set, resolve the `sbb-datepicker` by looking for `sbb-form-field` parent and checking its children
  
 Open questions from refinement:
 - Attachment rules of calendar/tooltip to toggle, or trigger
 - When should it be closed? blur, scroll, sticky, move with. trigger? backdrop click => tooltip behavior

```html
<sbb-calendar min="..." max="..."></sbb-calendar>
```

```html
<sbb-form-field label="Picker">
  <sbb-datepicker-previous-day></sbb-datepicker-previous-day>
  <sbb-datepicker-toggle></sbb-datepicker-toggle>
  <input>
  <sbb-datepicker-next-day></sbb-datepicker-next-day>
  <sbb-datepicker></sbb-datepicker>
</sbb-form-field>

```

### Definition of Done
 - [ ]  Component is implemented
 - [ ]  Documentation is provided
 - [ ]  Tests are implemented
 - [ ]  UX approved

MarioCastigliano:
## Preflight Checklist

## Issue

This PR Closes #1244

## Pull request checklist

## Changes

Changes in this pull request:
The following components were implemented
- sbb-calendar
- sbb-datepicker
- sbb-datepicker-toggle
- sbb-datepicker-previous-day
- sbb-datepicker-next-day

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [x] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/sbb-calendar/sbb-calendar.tsx
```diff
@@ -0,0 +1,420 @@
...
+    ) as HTMLButtonElement[];
+  }
+
+  /** Initializes the component. */
+  private _init(): void {
+    this._setCalendarWidth(this.wide);
+    this._setWeekdays();
+    this._weeks = this._createWeekRows(this._activeDate.getMonth(), this._activeDate.getFullYear());
+    this._nextMonthWeeks = [[]];
+    if (this._wide) {
```
jeripeierSBB, line 140: would this not also be relevant when changing wide property?

src/components/sbb-calendar/sbb-calendar.tsx
```diff
@@ -0,0 +1,420 @@
...
+        </td>
+      );
+    });
+  }
+
+  /** Creates the month label. */
+  private _createMonthLabel(d: Date): JSXElement {
+    return (
+      <span class="sbb-calendar__controls-month-label">
+        {this._dateAdapter.getMonthNames('long')[this._dateAdapter.getMonth(d)]}{' '}
```
jeripeierSBB, line 287: instead of the fancy  {' '} we could concatenate to string with template string?

src/components/sbb-calendar/sbb-calendar.tsx
```diff
@@ -0,0 +1,420 @@
...
+
+  private _handleKeyboardEvent(event, day: Day): void {
+    if (isArrowKeyOrPageKeysPressed(event)) {
+      event.preventDefault();
+    }
+    const index = this._days.findIndex((e: HTMLButtonElement) => e === event.target);
+    handleKeyboardEvent(event, index, this._days, day)?.focus();
+  }
+
+  /** Goes to the previous month. */
```
jeripeierSBB, line 324:
```suggestion
  /** Goes to month. */
```

src/components/sbb-calendar/sbb-calendar.tsx
```diff
@@ -0,0 +1,420 @@
...
+  }
+
+  private _getFirstFocusable(): HTMLButtonElement {
+    let firstFocusable =
+      this._element.shadowRoot.querySelector('.sbb-datepicker__day-selected') ??
+      this._element.shadowRoot.querySelector('.sbb-datepicker__day-today');
+    if (!firstFocusable || (firstFocusable as HTMLButtonElement)?.disabled) {
+      firstFocusable = Array.from(
+        this._element.shadowRoot.querySelectorAll('.sbb-datepicker__day')
+      ).find((e) => !(e as HTMLButtonElement).disabled);
```
jeripeierSBB, line 370: I guess it can be shortened?
```suggestion
        this._element.shadowRoot.querySelector('.sbb-datepicker__day:not([disabled]')
      ));
```

jeripeierSBB (commented):
Partial review of sbb-calendar

src/components/sbb-calendar/sbb-calendar.scss
```diff
@@ -0,0 +1,111 @@
...
+  gap: var(--sbb-spacing-fixed-10x);
+}
+
+.sbb-calendar__table {
+  width: 100%;
+  border-collapse: collapse;
+  height: max-content;
+}
+
+th {
```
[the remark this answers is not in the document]
MarioCastigliano: I agree, will do right away

src/components/sbb-calendar/sbb-calendar.tsx
```diff
@@ -0,0 +1,420 @@
...
+  public connectedCallback(): void {
+    window.addEventListener('resize', () => this._init(), { passive: true });
+    this.convertMinDate(this.min);
+    this.convertMaxDate(this.max);
+    this._setDates();
+    this._init();
+  }
+
+  public componentDidRender(): void {
+    this._setTabIndex();
```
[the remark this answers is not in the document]
MarioCastigliano: The tab indexes need to be re-calculated every the user selects a new date, changes month and any time the component loads from scratch. These are also the main cases in which the components renders, that's why I used the didRender hook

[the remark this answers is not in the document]
jeripeierSBB: ok should be ok. It was just that we normally never use componentDidRender (except for two other cases ;) ). It's only important to not retrigger another change. But with tabindex this is not the case. Can you add your explanation as a comment in the code?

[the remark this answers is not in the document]
MarioCastigliano: Sure

src/components/sbb-calendar/sbb-calendar.stories.js
```diff
@@ -0,0 +1,200 @@
...
+CalendarFilterFunction.args = { ...defaultArgs, dateFilter: dateFilter.options[2] };
+
+export default {
+  decorators: [
+    (Story) => (
+      <div>
+        <Story />
+      </div>
+    ),
+  ],
```
jeripeierSBB, line 187: This decorator is probably not needed?
```suggestion
```

src/global/helpers/native-date-adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /**
+   * Creates an array of strings with length = 12, filled with the name of the months starting from January, in the document language.
+   * @param style See `month` in `DateTimeFormatOptions`: `long` for full name, `short` for short name, `narrow` for single letter,
+   * E.g., with January in en-gb: `long` returns "January", `short` returns "Jan", `narrow` returns "J".
+   */
+  public getMonthNames(style: 'long' | 'short' | 'narrow'): string[] {
+    const formatter = new Intl.DateTimeFormat(documentLanguage(), { month: style });
+    return this._range(12, (i) => formatter.format(new Date(2017, i, 1)));
```
jeripeierSBB, line 81: Little bit hacky, but I don't know a better solution :D

src/global/helpers/native-date-adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+export class NativeDateAdapter {
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  public static readonly DAYS_PER_WEEK = 7;
+
+  /**
+   * Creates an array with the given length and fills it by mapping with the provided function.
+   * @param length The length of the array to be created.
+   * @param valueFunction The function of array's index used to fill the array.
+   */
+  private _range<T>(length: number, valueFunction: (index: number) => T): T[] {
```
jeripeierSBB, line 12: I'd place private methods at the end, but this is not a hard rule ;)

src/global/helpers/native-date-adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  /* Defines which is the first day of the week (0: sunday; 1: monday; etc.). */
+  public getFirstDayOfWeek(): number {
+    return 1;
+  }
+
+  /** Calculates the number of days in a month given the year and the month. */
+  public getNumDaysInMonth(year: number, month: number): number {
+    return new Date(year, month + 1, 0, 0, 0, 0, 0).getDate();
```
jeripeierSBB, line 107: surprised that this works

src/global/helpers/native-date-adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+    const result = this._createDateWithOverflow(year, month, date);
+    // Check that the date wasn't above the upper bound for the month, causing the month to overflow
+    if (result.getMonth() !== month) {
+      throw Error(`Invalid date "${date}" for month with index "${month}".`);
+    }
+
+    return result;
+  }
+
+  /** Checks whether the 2 dates are non-null and are in the same month of the same year. */
```
jeripeierSBB, line 135:
```suggestion
  /** Checks whether the two dates are non-null and are in the same month of the same year. */
```

src/global/helpers/native-date-adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+    return null;
+  }
+
+  /** Returns the right format for the `valueAsDate` property. */
+  public formatValueAsDate(value: string): Date {
+    const values = value?.split('.');
+    if (!values || values.length <= 2) {
+      return undefined;
+    }
+    return new Date(+values[2], +values[1] - 1, +values[0]);
```
jeripeierSBB, line 222: should the time not be set to 0 ? otherwise it depends when a user is calling formatValueAsDate(). but maybe I'm wrong

src/components/sbb-datepicker/sbb-datepicker.tsx
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  private _preventCharInsert(event): void {
+    const match = event.target.value.match(REGEX_PATTERN);
+    event.target.value = match?.[0] ?? null;
+  }
+
+  public render(): JSX.Element {
+    if (this._inputElement) {
+      this._inputElement.placeholder = i18nDatePickerPlaceholder[this._currentLanguage];
```
jeripeierSBB, line 208: I'd only overwrite if there is no placeholder on input (did we discuss this? :D)

src/components/sbb-datepicker/sbb-datepicker.tsx
```diff
@@ -0,0 +1,213 @@
...
+    }
+
```
jeripeierSBB, line 210: The place of this statement is a little bit weird for me. Is it really necessary to be in the render method?

src/components/sbb-datepicker/sbb-datepicker.tsx
```diff
@@ -0,0 +1,213 @@
...
+      return value;
+    }
+
+    const day: string = match[1].padStart(2, '0');
+    const month: string = match[2].padStart(2, '0');
+    let year: number = +match[3];
+    if (year < 100 && year >= 0) {
+      year += 1900;
+    }
+    return `${day}.${month}.${year}`;
```
jeripeierSBB, line 159: should we not use DateTimeFormat with options { year: 'numeric', month: '2-digit', day: '2-digit' }? maybe not, I don't really know :D

jeripeierSBB (changes requested):
Finally finished for round 1 :)

src/global/helpers/native-date-adapter.ts
MarioCastigliano: It's already 0:00 by default if you don't specify a time, but I can make it explicit

src/components/sbb-form-field/sbb-form-field.tsx
```diff
@@ -163,15 +163,17 @@ export class SbbFormField implements ComponentInterface {
...
+    value && this._input?.setAttribute('aria-describedby', value);
   }
 
   public render(): JSX.Element {
     return (
       <div class="sbb-form-field__space-wrapper">
         {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */}
         <div onClick={(): void => this._input?.focus()} class="sbb-form-field__wrapper">
-          <slot name="prefix"></slot>
+          <div class="sbb-form-field__slot">
```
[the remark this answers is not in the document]
MarioCastigliano: The outer div (sbb-form-field__wrapper), has display:flex, and a gap, which was applied also to the slot's content (because by default the slot has display:contents) causing unwanted spacing between the arrow and the toggle. The slot doesn't allow classes unfortunately

kyubisation (changes requested):
Review still in progress
Great work so far!

src/components/sbb-datepicker-next-day/sbb-datepicker-next-day.tsx
```diff
@@ -0,0 +1,159 @@
...
+        datepicker.dateFilter,
+        this._dateAdapter,
+        this._max
+      );
+      this._disabled = this._dateAdapter.compareDate(nextDate, pickerValueAsDate) === 0;
+    }
+  }
+
+  private async _handleClick(event: Event): Promise<void> {
+    event.stopImmediatePropagation();
```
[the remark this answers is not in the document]
MarioCastigliano: The form-field refocuses non the input every time the user clicks on anything that's inside the form-field (label, icons but also date-picker controls). Without stopping the event the user experience becomes pretty wonky and the keyboard navigation breaks completely

src/components/sbb-datepicker-next-day/sbb-datepicker-next-day.tsx
```diff
@@ -0,0 +1,159 @@
...
+} from '../sbb-datepicker/sbb-datepicker.helper';
+
+@Component({
+  shadow: true,
+  styleUrl: 'sbb-datepicker-next-day.scss',
+  tag: 'sbb-datepicker-next-day',
+})
+export class SbbDatepickerNextDay implements ComponentInterface {
+  /** Datepicker reference. */
+  @Prop() public datePicker?: string | HTMLElement;
```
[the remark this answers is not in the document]
MarioCastigliano: Apart from being complicated, I'd say it's also not safe. The idea behind the current implementation is that the datepicker itself serves as a communication layer between the input and the controls, acting not only as an interface but also as a 'single source of truth'.

src/global/i18n.ts
```diff
@@ -418,6 +418,48 @@ export const i18nTomorrow: Record<string, string> = {
   it: 'Domani',
 };
 
+export const i18nNextMonth: Record<string, string> = {
```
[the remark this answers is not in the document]
MarioCastigliano: Got those from the angular sbb-datepicker

src/components/sbb-datepicker/sbb-datepicker.tsx
MarioCastigliano: Currently we're formatting the date as dd.mm.yyyy, and DateTimeFormat doesn't really let you choose the separator. In this way with lang="en" we would have mm/dd/yyyy

src/global/helpers/native-date-adapter.ts
MarioCastigliano: Well, while being a little counterintuitive it's a fairly solid logic, since that the day 0 of a month is always the last day of the previous month, which corresponds to its length 😅

src/components/sbb-datepicker/sbb-datepicker.tsx
```diff
@@ -0,0 +1,213 @@
...
+  /(^0?[1-9]?|[12]?[0-9]?|3?[01]?)[.,\\/\-\s](0?[1-9]?|1?[0-2]?)?[.,\\/\-\s](\d{1,4}$)?/;
+
+@Component({
+  shadow: true,
+  styleUrl: 'sbb-datepicker.scss',
+  tag: 'sbb-datepicker',
+})
+export class SbbDatepicker implements ComponentInterface {
+  /** If set to true, two months are displayed */
+  @Prop() public wide = false;
```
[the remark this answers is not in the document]
MarioCastigliano: The idea was to handle every datepicker-related aspect on the datepicker itself, but to me it's the same. Let's see what @kyubisation has to say about this 😃

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
```diff
@@ -8,26 +9,30 @@ import { Component, ComponentInterface, h, Host, JSX, Prop } from '@stencil/core
...
+  @Prop() public accessibilityLabel: string | undefined;
+
   public render(): JSX.Element {
     return (
-      <Host role="button" aria-disabled={this.disabled}>
-        <button class="sbb-tooltip-trigger" disabled={this.disabled}>
-          <slot>{this.iconName && <sbb-icon name={this.iconName} />}</slot>
-        </button>
-      </Host>
+      <button
```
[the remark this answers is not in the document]
MarioCastigliano: Accessibility was off, should I just revert this and let Lukas take care of it?

src/global/helpers/native-date-adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  /** Gets the day of the input date. */
+  public getDate(date: Date): number {
+    return date.getDate();
+  }
+
+  /** Gets the day of the week of the input date. */
+  public getDayOfWeek(date: Date): number {
+    return date.getDay();
+  }
```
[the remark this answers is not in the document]
MarioCastigliano: They were inherited from the angular version of the native date adapter, but since we're not using a "DateAdapter" interface in this case, and we're just dealing with native Date objects, I think we can get rid of them.

src/global/helpers/native-date-adapter.ts
MarioCastigliano: I kinda re-managed the component to call this method only on first load and on language change, but for the method itself I can't see how to improve it 😕

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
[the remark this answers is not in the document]
kyubisation: I am going to refactor this in my PR, you can revert it for now 👍

src/components/sbb-datepicker/sbb-datepicker.tsx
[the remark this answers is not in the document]
kyubisation: For me it is better on the sbb-datepicker, as it is primarily related to the datepicker, not the toggle.

src/components/sbb-datepicker-next-day/sbb-datepicker-next-day.tsx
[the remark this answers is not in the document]
kyubisation: Ah, right, I forgot about that. Maybe we need to check in the form field, if the click event originated from a button or [role="button"] and only focus if that is not the case.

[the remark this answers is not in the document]
MarioCastigliano: Okay, makes sense. How should we handle the datepicker toggle? Currently it has no role="button" and whether you click on the tooltip trigger or the tooltip, the event target is the datepicker toggle which contains both. (Composed path? Would that be ok performance-wise too?)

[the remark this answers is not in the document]
kyubisation: Hm, I think it is fine for now to leave a comment on this consideration in the code so we can revisit this in the future.

src/components/sbb-form-field/sbb-form-field.tsx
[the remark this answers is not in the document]
jeripeierSBB: If you are sure that this does not have side effects, it's ok I think.

But I guess it's possible to apply classes, you just have to change display to flex of the <slot> element. not? But at the end it would be the same only with one element less in the DOM :)

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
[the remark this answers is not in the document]
jeripeierSBB: I think we will soon merge the refactor PR, you can then adapt it.

src/global/i18n.ts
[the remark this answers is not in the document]
jeripeierSBB: ok. should be ok then. And at least in Italian you would have recognized wrong words :P

src/global/helpers/native-date-adapter.ts
jeripeierSBB: I did also not found a better solution, so let's consider it as best solution :D

src/components/sbb-datepicker/sbb-datepicker.tsx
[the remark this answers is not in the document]
jeripeierSBB: On the other hand, the toggle has the calendar in it :) I can live also with this solution, I think it can be a little bit confusing. but also the other way :D

src/components/sbb-datepicker-next-day/sbb-datepicker-next-day.tsx
```diff
@@ -0,0 +1,145 @@
...
+  private _datePickerElement: HTMLSbbDatepickerElement;
+
+  private _dateAdapter: NativeDateAdapter = new NativeDateAdapter();
+
+  private _datePickerController: AbortController;
+
+  @Watch('datePicker')
+  public findDatePicker(newValue: string | HTMLElement, oldValue: string | HTMLElement): void {
+    if (newValue !== oldValue) {
+      this._init(this.datePicker);
```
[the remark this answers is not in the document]
kyubisation: I don't think this was resolved? My recommendation would be to call abort in _init before creating a new AbortController.

kyubisation (changes requested):
Some issues, but other than that looks good to me 👍

src/components/sbb-datepicker/sbb-datepicker.tsx
```diff
@@ -0,0 +1,235 @@
...
+        return;
+      }
+
+      this._inputObserver?.disconnect();
+      this._inputObserver.observe(this._inputElement, {
+        attributeFilter: ['disabled', 'readonly', 'min', 'max'],
+      });
+
+      this._inputElement.type = 'text';
+      this._inputElement.role = 'status';
```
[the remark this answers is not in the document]
MarioCastigliano: Okay, I'll test if `aria-live="polite"` works in the same way for this scenario

jeripeierSBB (changes requested):
Partial review, will continue tomorrow

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
```diff
@@ -46,7 +50,7 @@ export class SbbTooltipTrigger implements ComponentInterface, ButtonProperties {
 
     return (
       <Host {...hostAttributes}>
-        <span class="sbb-tooltip-trigger">
+        <span class="sbb-tooltip-trigger" data-icon-small={this._isInFormField}>
```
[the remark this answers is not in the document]
MarioCastigliano: The icon-button mixin wouldn't work if I did that because it uses :host() selectors. To separate the cases where the tooltip trigger is inside a form field from the cases where it isn't, I should use something like :host([data-icon-small]) { @include sbb.icon-button }, using essentially a host selector inside another host selector, causing the mixin to have no effect.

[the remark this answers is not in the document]
jeripeierSBB: I see the problem which let me have a deeper look at the mixin. In order to support also the mixin being included in the sbb-button, I think we should refactor it the way we often do it:
- split up the mixins into xy--variables where there are the default variables contained
- create mixins with only variables for every state (hover, active, focus, ..?) without containing any selectors. focus visible will probably need one mixin more.
- define the selectors only in the consuming component

src/components/sbb-calendar/sbb-calendar.stories.js
```diff
@@ -0,0 +1,208 @@
...
+  'data-now': dataNow,
+};
+
+const defaultArgs = {
+  wide: false,
+  selectedDate: new Date(2023, 0, 20),
+  dateFilter: dateFilter.options[0],
+};
+
+if (isChromatic()) {
```
jeripeierSBB, line 157: move to args object

src/components/sbb-calendar/sbb-calendar.spec.ts
```diff
@@ -0,0 +1,267 @@
+import { SbbCalendar } from './sbb-calendar';
+import { newSpecPage } from '@stencil/core/testing';
+
+describe('sbb-calendar', () => {
+  it('renders', async () => {
+    const page = await newSpecPage({
+      components: [SbbCalendar],
+      html: "<sbb-calendar selected-date='2023-01-20T00:00:00'/>",
```
jeripeierSBB, line 8: maybe use data-now to also check for classes set

src/components/sbb-calendar/sbb-calendar.scss
```diff
@@ -0,0 +1,144 @@
...
+
+.sbb-calendar__table-data {
+  padding: var(--sbb-calendar-day-padding);
+  height: var(--sbb-calendar-day-size);
+  width: var(--sbb-calendar-day-size);
+  transition-duration: var(--sbb-calendar-day-transition-duration);
+  transition-timing-function: var(--sbb-calendar-day-transition-easing-function);
+}
+
+.sbb-calendar__table-data:has(.sbb-calendar__day:not(.sbb-calendar__day-selected):active) {
```
jeripeierSBB, line 81: please find alternative for :has .

src/components/sbb-calendar/sbb-calendar.scss
```diff
@@ -0,0 +1,144 @@
+@use '../../global/styles' as sbb;
```
jeripeierSBB, line 1: Please check high contrast mode for every component

src/global/styles/mixins/buttons.scss
```diff
@@ -63,6 +63,10 @@
   pointer-events: var(--sbb-button-pointer-events);
   background-color: transparent;
   color: var(--sbb-button-color-default-text);
+  outline-offset: var(--sbb-focus-outline-offset);
```
jeripeierSBB, line 66: Is this line really needed?

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.scss
```diff
@@ -64,14 +64,14 @@
...
 
-  :host([data-icon-small]:is(:active, [data-active])) & {
-    @include sbb.icon-button-active;
-  }
-
   :host([data-icon-small]:hover) & {
     @include sbb.icon-button-hover;
   }
 
+  :host([data-icon-small]:is(:active, [data-active])) & {
```
jeripeierSBB, line 71: If order matters, we should commented it

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.scss
```diff
@@ -37,23 +37,51 @@
   pointer-events: none;
 }
 
-.sbb-tooltip-trigger:not([data-icon-small]) {
-  @include sbb.text-inherit;
+.sbb-tooltip-trigger {
+  :host([data-icon-small]) & {
```
jeripeierSBB, line 41: What if using a tooltip trigger in a form field on a text and not on an icon? I feel that the design is not currently not very clear. There are somehow different kinds of tooltip trigger. Maybe a more general discussion. @kyubisation

src/global/styles/mixins/buttons.scss
```diff
@@ -15,3 +19,102 @@
...
+
+@mixin icon-button {
+  @include icon-button-base('sbb-icon');
+
+  :host(:focus-visible) & {
+    @include a11y.focus-outline;
+  }
+
+  :host([data-active]) &,
+  :host(:active) & {
```
jeripeierSBB, line 31:
```suggestion
  :host(:is([data-active], :active)) & {
```

src/global/styles/mixins/buttons.scss
```diff
@@ -15,3 +19,102 @@
   -webkit-tap-highlight-color: transparent;
   color: inherit;
 }
+
```
jeripeierSBB, line 22: I'm not sure about the design of the mixins, but let's discuss it when refactoring it anyways for the sbb-button usage :)

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.scss
kyubisation: Should be fine for now. Re-evaluate when we refactor tooltip.

src/global/styles/mixins/buttons.scss
MarioCastigliano: It kinda is to make sure the animation is okay, but I refactored the style anyway in a more correct way

src/global/styles/mixins/buttons.scss
MarioCastigliano: I totally agree, I sense that I'll have to re-shape the whole thing anyway

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
```diff
@@ -37,7 +37,8 @@ export class SbbTooltipTrigger implements ComponentInterface, ButtonProperties {
     this._handlerRepository.connect();
     this._isInFormField = !!(
       hostContext('sbb-form-field', this._element) ??
-      hostContext('[data-form-field]', this._element)
+      hostContext('[data-form-field]', this._element) ??
+      this._element.dataset.iconSmall === ''
```
jeripeierSBB, line 41: Is this not kind of circular?

src/components/sbb-calendar/sbb-calendar.scss
```diff
@@ -71,25 +71,29 @@
 }
 
 .sbb-calendar__table-data {
-  position: relative;
+  padding: var(--sbb-calendar-day-padding);
```
jeripeierSBB, line 74: seems to introduce a new bug on placing the strike through on filtered out dates.
Bildschirmfoto 2023-04-12 um 08 02 58

src/components/sbb-form-field/sbb-form-field.tsx
```diff
@@ -261,14 +261,20 @@ export class SbbFormField implements ComponentInterface {
...
+
   public render(): JSX.Element {
     return (
       <div class="sbb-form-field__space-wrapper">
         {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */}
         <div onClick={(event) => this._handleWrapperClick(event)} class="sbb-form-field__wrapper">
-          <div class="sbb-form-field__slot-container">
-            <slot name="prefix"></slot>
-          </div>
+          {this._slotHasContent('prefix') && (
```
jeripeierSBB, line 273: We have for that the the named slot handler which also takes care of changing events (see examples for full code snippets).
`  @State() private _namedSlots = createNamedSlotState('prefix');`
`  private _handlerRepository = new HandlerRepository(
    this._element,
    namedSlotChangeHandlerAspect((m) => (this._namedSlots = m(this._namedSlots)))
  ); `

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
```diff
@@ -31,15 +31,15 @@ export class SbbTooltipTrigger implements ComponentInterface, ButtonProperties {
...
     this._handlerRepository.connect();
-    this._isInFormField = !!(
+    if (
       hostContext('sbb-form-field', this._element) ??
-      hostContext('[data-form-field]', this._element) ??
-      this._element.dataset.iconSmall === ''
-    );
+      hostContext('[data-form-field]', this._element)
+    ) {
+      this._element.dataset.iconSmall = '';
```
jeripeierSBB, line 41: I like the solution now :) Although it's the same, I'd prefer to use toggleDatasetEntry helper here.

``

jeripeierSBB (changes requested):

src/components/sbb-tooltip-trigger/sbb-tooltip-trigger.tsx
MarioCastigliano: But toggleDatasetEntry removes the attribute if the condition is not true, we just want to set it in case it's true, unless we want to include the condition `this._element.dataset.iconSmall === ''` I don't think this utility is what we need  for this case

src/components/sbb-datepicker/sbb-datepicker.tsx
```diff
@@ -159,9 +160,19 @@ export class SbbDatepicker implements ComponentInterface {
     );
   }
 
+  @Listen('datepicker-control-registered')
+  public onInputPropertiesChange(): void {
```
jeripeierSBB, line 164: I think method does not have to be public?

jeripeierSBB (approved):
Congratulations to this great work you did! Thank you for bringing the huge effort to make it work and look perfectly :)

src/components/sbb-datepicker/sbb-datepicker.tsx
MarioCastigliano: You're right, I just assumed it would throw errors when using Listen with private methods, turns out it does not 😄. Fixing this right away

kyubisation (approved):
LGTM 👍 
Thanks for your excellent work!

PR #1643, opened 2023-02-23
fix(sbb-toggle-check): scroll into view when focused via tab key

### LINKED ISSUE #1639: [Bug] (sbb-dialog): scroll behavior with tab
currently the sbb-dialog doesn't scroll when we navigate with tab

dauriamarco:
Closes #1639

### DIFF
--- a/src/components/sbb-toggle-check/sbb-toggle-check.e2e.ts
+++ b/src/components/sbb-toggle-check/sbb-toggle-check.e2e.ts
@@ -46,6 +46,9 @@ describe('sbb-toggle-check', () => {
 
       // Although the inner native button receives the focus, the active element is the host
       expect(await page.evaluate(() => document.activeElement.id)).toBe('outer-id');
+      expect(
+        await page.evaluate(() => document.activeElement.shadowRoot.activeElement.tagName)
+      ).toBe('INPUT');
     });
   });
 });
--- a/src/components/sbb-toggle-check/sbb-toggle-check.e2e.ts
+++ b/src/components/sbb-toggle-check/sbb-toggle-check.e2e.ts
@@ -46,9 +46,6 @@ describe('sbb-toggle-check', () => {
 
       // Although the inner native button receives the focus, the active element is the host
       expect(await page.evaluate(() => document.activeElement.id)).toBe('outer-id');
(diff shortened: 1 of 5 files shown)


jeripeierSBB (approved):
LGTM, but should we also check other components?

src/components/sbb-toggle-check/sbb-toggle-check.tsx
```diff
@@ -87,19 +87,19 @@ export class SbbToggleCheck implements ComponentInterface, AccessibilityProperti
...
-          required={this.required}
-          checked={this.checked}
-          value={this.value}
-          onChange={(event: Event): void => this.checkedChanged(event)}
-          aria-label={this.accessibilityLabel}
-        />
-        <label class="sbb-toggle-check" htmlFor="sbb-toggle-check-input">
+        <label class="sbb-toggle-check">
+          <input
+            id="sbb-toggle-check-input"
```
jeripeierSBB, line 92: Do we still need the id?

dauriamarco:
> LGTM, but should we also check other components?

I've checked and I think the only other case where we don't have the input nested in the label is with form fields, but I've tested them and they seem to already work correctly.

src/components/sbb-toggle-check/sbb-toggle-check.tsx
dauriamarco: Probably not 👍

kyubisation (approved):
LGTM 👍

src/components/sbb-toggle-check/sbb-toggle-check.e2e.ts
```diff
@@ -46,9 +46,6 @@ describe('sbb-toggle-check', () => {
 
       // Although the inner native button receives the focus, the active element is the host
       expect(await page.evaluate(() => document.activeElement.id)).toBe('outer-id');
-      expect(await page.evaluate(() => document.activeElement.shadowRoot.activeElement.id)).toBe(
```
jeripeierSBB, line 49: maybe check for tagName here?

jeripeierSBB (changes requested):

jeripeierSBB (approved):
LGTM

PR #1645, opened 2023-02-23
fix(sbb-icon): remove unwanted white space

dauriamarco:
Get rid of unwanted white space under the `sbb-icon`.


kyubisation (approved):
LGTM 👍 
Nice work! 😃

dauriamarco:
We will investigate for a better solution that has less impact on the other components where the icon is used.

PR #1646, opened 2023-02-23
fix(sbb-navigation): set correct dimension on mobile

dauriamarco:
Use dynamic viewport units to calculate the correct dimension for the `sbb-navigation` and the `sbb-navigation-section`. The bug is visible when the nav or the section goes in overflow on mobile.

### DIFF
--- a/src/components/sbb-navigation-section/sbb-navigation-section.scss
+++ b/src/components/sbb-navigation-section/sbb-navigation-section.scss
@@ -16,8 +16,8 @@
   --sbb-navigation-section-transform: translateX(100%);
   --sbb-navigation-section-content-padding-inline-start: var(--sbb-spacing-fixed-12x);
   --sbb-focus-outline-color: var(--sbb-focus-outline-color-dark);
-  --sbb-navigation-section-width: 100vw;
-  --sbb-navigation-section-height: 100vh;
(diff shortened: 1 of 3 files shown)


jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍

src/components/sbb-navigation-section/sbb-navigation-section.scss
```diff
@@ -19,6 +19,14 @@
   --sbb-navigation-section-width: 100vw;
   --sbb-navigation-section-height: 100vh;
 
+  @supports (width: 100dvw) {
```
jeripeierSBB, line 22: Should we add this fallback also for menu?

src/components/sbb-navigation/sbb-navigation.scss
```diff
@@ -76,7 +81,7 @@
   --sbb-navigation-content-transform: translateX(-100%);
 
   @include sbb.mq($from: 'large') {
-    --sbb-navigation-expanded-width: 100vw;
+    --sbb-navigation-expanded-width: 100dvw;
```
jeripeierSBB, line 84: supports check also needed here?

jeripeierSBB (approved):
LGTM

PR #1650, opened 2023-02-27
fix(sbb-dialog): observe dialog content resizing

dauriamarco:
Handle the resizing of the dialog content and apply the correct style to the dialog accordingly.


jeripeierSBB (approved):
LGTM thanks

PR #1654, opened 2023-02-28
fix: improve focus trap class mechanism

dauriamarco:
Improve the focus trap class in order to make it work also with slotted components with a different shadowRoot (e.g. `sbb-datepicker` inside the tooltip). iIx `sbb-form-field` focus handling.


src/global/helpers/focus.ts
```diff
@@ -1,6 +1,6 @@
 export type SbbFocusOrigin = 'touch' | 'mouse' | 'keyboard' | 'program' | null;
 
-export const IS_FOCUSABLE_QUERY = `:is(button, [href], input, select, textarea, details, summary:not(:disabled), [tabindex], sbb-button, sbb-link, sbb-menu-action, sbb-navigation-action):not([disabled]):not([tabindex="-1"]):not([static])`;
+export const IS_FOCUSABLE_QUERY = `:is(button, [href], input, select, textarea, details, summary:not(:disabled), [tabindex], sbb-button, sbb-link, sbb-menu-action, sbb-navigation-action):not([disabled]:not([disabled='false'])):not([tabindex="-1"]):not([static])`;
```
jeripeierSBB, line 3: nice catch, thanks! :)

src/components/sbb-form-field/sbb-form-field.tsx
```diff
@@ -142,11 +142,26 @@ export class SbbFormField implements ComponentInterface {
     this._namedSlots = queryNamedSlotState(this._element, this._namedSlots, event.detail);
   }
 
+  @Listen('will-open')
```
jeripeierSBB, line 145: I'm still not very happy about the coupling to the tooltip, but you probably did not found "better" solutions?

jeripeierSBB: detection via trigger is maybe a solution, but would need mutationobserver :/ and finding all possible triggers can be expensive. :has() selector is still not default available in firefox, but would be a nice option, e.g. something like `:host(:has([aria-expanded]))`, but maybe we can add a todo, as. during summer :has() should be available

jeripeierSBB (approved):
LGTM, with same changes in mind for the future

PR #1680, opened 2023-03-21
fix(sbb-form-field): always use label to link to input

kyubisation:
BREAKING CHANGE: It is now required to use a `<label>` to label an input via slot/element in a `<sbb-form-field>`.


jeripeierSBB (approved):
Cool! LGTM

PR #1681, opened 2023-03-21
fix(sbb-icon): empty icon on Safari

dauriamarco:
Fix incorrect rendering on Safari when an `sbb-icon` is provided without a specified name property.

### DIFF
--- a/src/components/sbb-icon/sbb-icon.scss
+++ b/src/components/sbb-icon/sbb-icon.scss
@@ -10,6 +10,10 @@
   -webkit-tap-highlight-color: transparent;
 }
 
+:host([data-empty]) {
+  --sbb-icon-default-dimension: 0;
+}
+
 :host([data-empty][data-namespace='default'][name$='-small']) {
   --sbb-icon-default-dimension: var(--sbb-size-icon-ui-small);
 }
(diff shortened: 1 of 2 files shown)


src/components/sbb-icon/sbb-icon.tsx
```diff
@@ -114,10 +114,12 @@ export class SbbIcon implements ComponentInterface {
         {this._svgIcon ? (
           <span class="sbb-icon-inner" innerHTML={this._svgIcon}></span>
         ) : (
-          <span class="sbb-icon-inner">
-            {/* To reserve space, we need an empty svg to apply dimension to. */}
-            <svg width="0" height="0"></svg>
-          </span>
+          this.name && (
```
jeripeierSBB, line 117: In my opinion I would keep the existing solution so that an empty name could have a meaning (to have a placeholder icon). So I'd suggest to fix it in the tab group story. What do you think?

dauriamarco: Ok, I can fix it from the CSS without changing the logic 👍

jeripeierSBB (approved):
I think it should be ok like that. thanks :)

PR #1684, opened 2023-03-23
fix: stabilize positioning function

dauriamarco:
Make sure that the positioned element is overflowing before repositioning it to where there is more space available.


jeripeierSBB (approved):
LGTM, but what about a test for the new logic?

PR #1685, opened 2023-03-27
fix(sbb-menu-action): assign pointer-events to none when disabled

kyubisation:
This PR applies `pointer-events: none` for `<sbb-menu-action>` when disabled. This is done similarly in e.g. `<sbb-button>` and `<sbb-link>`.

### DIFF
--- a/src/components/sbb-menu-action/sbb-menu-action.scss
+++ b/src/components/sbb-menu-action/sbb-menu-action.scss
@@ -29,6 +29,9 @@
   --sbb-menu-action-color: var(--sbb-color-graphite-default);
   --sbb-menu-action-forced-color-border-color: GrayText;
 
+  pointer-events: none;
+  cursor: default;
+
   @include sbb.if-forced-colors {
     --sbb-menu-action-color: GrayText;
   }


jeripeierSBB (approved):
LGTM

src/components/sbb-menu-action/sbb-menu-action.e2e.ts
```diff
@@ -6,42 +6,31 @@ describe('sbb-menu-action', () => {
...
-      const menuAction = await page.find('sbb-menu-action');
       const changeSpy = await page.spyOnEvent('click');
 
-      await menuAction.click();
+      await element.click();
       expect(changeSpy).toHaveReceivedEventTimes(1);
     });
 
+    /*
+    TODO: Figure out why this test fails, when sbb-button test works
```
dauriamarco, line 22:
```suggestion
    it('should not dispatch event on click if disabled', async () => {
      element.setAttribute('disabled', true);
      await page.waitForChanges();
      
      const clickSpy = await page.spyOnEvent('click');
      
      element.triggerEvent('click');
      await page.waitForChanges();
      
      expect(clickSpy).not.toHaveReceivedEvent();
    });
```

I suggest to try this approach as the method `triggerEvent()` turns out to be more reliable in cases like this.

dauriamarco (approved):
LGTM, I just left a comment about a test 👍

src/components/sbb-menu-action/sbb-menu-action.e2e.ts
kyubisation: Thanks for the hint 😃 👍 
I now use `element.triggerEvent('click', { bubbles: true, cancelable: true, composed: true });` and that seems to work.

PR #1692, opened 2023-03-30
refactor: adapt language change and named slot change to new repository pattern

kyubisation:


jeripeierSBB (approved):
LGTM, Not 100% sure because it seems to generate more code

PR #1710, opened 2023-04-12
fix(sbb-button): fix icon only detection if whitespaces in slot

kyubisation:


jeripeierSBB (approved):
Can textContent be undefined or not a string? And maybe unit / e2e tests for this.

jeripeierSBB (approved):
LGTM ;)

PR #1713, opened 2023-04-13
chore: stabilize e2e tests

jeripeierSBB:


kyubisation (approved):
I'm not quite sure whether this is the best solution, but if it makes our builds more stable then it is certainly an improvement. 👍

PR #1727, opened 2023-05-01
fix(sbb-tooltip): fix Safari outline on trigger after closing

jeripeierSBB:


jeripeierSBB:
@dauriamarco Hey Marco, I had to revert some recently introduced changes of you, as it broke the logic. Now it's working again, but maybe there would be a nicer solution :D

dauriamarco (approved):
Hi Jeri, that's ok. Thank you for fixing it 👍

PR #1732, opened 2023-05-03
fix(sbb-card, sbb-card-badge): fix accessibility of card

### LINKED ISSUE #1242: refactoring: sbb-card-badge
### Feature Description
We want to review/refactor the sbb-card-badge implementation.

[Figma Spec] (https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=1341%3A636046)

- Skew separate element on the left instead of whole element (avoids text unskew)
- Discuss removal of isDiscount, text and price with @mcilurzo (should be left to user)
- Discuss spacings between texts with @mcilurzo
- Is title attribute intended? @mcilurzo
- Replace accessibilityLabel with aria-label, aria-labelledby and aria-describedby forwardings
- Remove size
- Replace appearance with :host-context selectors
- Always render slots (remove check in componentWillLoad)

**Technical Spec**

- 

### Definition of Done

 - [ ]  Tests are implemented
 - [ ]  UX approved

### LINKED ISSUE #1833: feat: sbb-card / sbb-group refactoring
With this refactoring the sbb-group and sbb-card should be merged into one component. To make the card interactive, a <sbb-card-action> element can be placed inside the card. This solves accessibility issues and enables also slotting other actions or interactive content into a card.

Moreover, the new dashed variant should be included.

To be synchronized with #1242 

Remaining Todos:
 - [x] Nested buttons in card examples / home stories should be static @jeripeierSBB 
 - [x] Naming of dashed property align with Figma
 - [x] Clean up card mixin
 - [x] Align timetable row with card refactoring and document it @dauriamarco 
 - [ ] Testing with screen readers and browser combination
 - [x] Check high contrast mode @dauriamarco 
 - [x] Documentation @jeripeierSBB 
 - [ ] Migration docs
 - [x] UX check (Manuel)
 - [x] Acesssibility Check (Mürsel)

jeripeierSBB:
This structure change removes the `<sbb-group>` and refactors `<sbb-card>` & `<sbb-card-badge>` in a breaking way. 

With this change we provide a much better accessibility experience. Now the card is not an action element itself anymore, but the action functionality is delegated to the new `<sbb-card-action>` component which can be placed inside an `<sbb-card>`. Due to this, the semantic meaning of the  `<sbb-card>` content is preserved (as semantic meaning is lost in a button/link), and now completely accessible for all kind of other elements/roles (headings, interactive elements, tables, etc.). With `<sbb-card-action>` it is required to place a descriptive label/content in it (e.g. `<sbb-card-action>Buy ticket for Bern to Münsingen</sbb-card-action>`), as this is used as the description for the action (button/link).

Moreover, the `<sbb-card-badge>` was simplified and counts now as a "finished" component.

Last but not least, we introduced the new color variant `transparent-bordered-dashed` and the new size `xxxl` of the `<sbb-card>`.

Closes #1833 
Closes #1242 

BREAKING CHANGE:

- `sbb-group` was removed and replaced by `sbb-card`.

   **How to migrate**:
    - Rename `sbb-group` to `sbb-card`.
    - If `color` property is set to `transparent`, change value from `transparent` to `transparent-bordered`.
    - Replace property `padding` with property `size` and update the values like following:
      - Padding `3x-xxs` -> size `xs`
      - Padding `xxxs-xxs` -> size `s`
      - Padding `xxxs-s` -> size `m`
      - Padding `4x-xxs` -> size `l`
      - Padding `xxs-xxs` -> size `xl`
      - Padding `s-s` -> size `xxl`

----------------------------------------------

- `sbb-card` is not interactive anymore and replaces the `sbb-group` in a static way. To create an action, add `<sbb-card-action>` component as content to `sbb-card`. All the interactive properties of the former `<sbb-card>` were transferred to the new `<sbb-card-action>` (`active`, `download`, `form`, `href`, `name`, `rel`, `target`, `type`, `value`).

   Before, interactive content in the unnamed slot of the `<sbb-card>` was automatically made static as it is prohibited to have an action nested in an action. However, as the content is not nested anymore in the interactive element, nested interactive elements stay interactive. This needs either adding `is-static` to the `<sbb-button>` or `<sbb-link>` or adding a real action to it (other action than the card, but it's not recommended!).

   **Windows High Contrast notes**

   In high contrast mode, all the content of a link or a button receives a specific color which overrides every other color. 
However, as the content of the card is not directly inside the button or link, this does not happen 
when the slotted content has a specific color set.
To improve coloring, it's needed to manually define styles for high contrast mode (setting `LinkText` or `ButtonText`).

   **How to migrate**:
   1. Create a `<sbb-card-action>` element and add it as content of the `<sbb-card>`.
   2. Move all action related properties from `<sbb-card>` to `<sbb-card-action>` (`active`, `download`, `form`, `href`, `name`, `rel`, `target`, `type`, `value`).
   3. **Create a meaningful label which describes the action.** This label is important for search engines and screen readers but is not visible.
   4. If you had other interactive elements inside the `<sbb-card>`, as e.g. the often seen usage of a `<sbb-button>`, add `is-static` attribute to the element. 
   5. If needed, style content of card for windows high contrast mode with `LinkText` or `ButtonText`.
 
   before:
   ```
   <sbb-card href="https://www.sbb.ch" target="_blank">
     Tickets
     <sbb-button>Buy a ticket</sbb-button>
   </sbb-card>
   ```

   after:
   ```
   <sbb-card>
     <sbb-card-action href="https://www.sbb.ch" target="_blank">Buy a ticket</sbb-card-action>
     Tickets
     <sbb-button **is-static**>Buy a ticket</sbb-button>
   </sbb-card>
    ```

----------------------------------------------

-  All properties of `<sbb-card-badge>` were either removed and replaced by the unnamed slot or renamed. 
    - The property `size` was removed, as it was not used at all.
    - The property `appearance` was renamed to `color`. Its old value `primary` becomes `charcoal` and its old value `primary-negative` becomes `white`.
    - The properties `isDiscount`, `price`, `text` were removed.
    - The attribute `slot="badge"` is assigned automatically and therefore doesn't have to be added by the consumer anymore.

   **How to migrate**:
    - The property `appearance` should be renamed to `color`. Its old value `primary` becomes `charcoal` and its old value `primary-negative` becomes `white`.
    - The properties `isDiscount`, `price`, `text` should be replaced with `<span>` elements in the unnamed slot (see example below). If the `isDiscount` property was set, it should be replaced by `<span>%</span>`.
    
    before:
   ```
   <sbb-card-badge 
     aria-label="Super saver sales ticket price starts at CHF 92.50 Black Friday Special"
     appearance="primary"
     is-discount
     price="92.50"
     text="from CHF"
     slot="badge"
   >
     <div slot="generic"><time datetime="2021-11-25">Black Friday</time> Special</div>
   </sbb-card-badge>
   ```

   after:
   ```
    <sbb-card-badge 
      aria-label="Super saver sales ticket price starts at CHF 92.50 Black Friday Special"
      color="charcoal"
    >
      <span>%</span> <!-- Replaces the logic of isDiscount -->
      <span>from CHF</span>
      <span>92.50</span>
      <span><time datetime="2021-11-25">Black Friday</time></span>
    </sbb-card-badge>
    ```


src/global/styles/typography.scss
```diff
@@ -60,6 +60,10 @@ sbb-breadcrumb-group:not(.hydrated) {
   overflow: hidden;
 }
 
+[data-is-card-content-focusable] {
```
kyubisation, line 63: Should we scope this to `sbb-card`?

kyubisation (changes requested):
Nice work so far! 😃

src/components/sbb-card/readme.md
```diff
@@ -1,49 +1,78 @@
-The `sbb-card` component is a generic content container as an action element; its task is to contain HTML elements related to a single subject.
+The `sbb-card` component is a generic content container; its task is to contain content related to a single subject. 
+There are various sizes (affecting paddings) and colors available.
 
-Internally, it is either rendered as an anchor (`<a>`, when used with a `href` attribute) or as a button (when used without a `href` attribute).
-When rendered as a button, it is possible to listen to the click event to add behavior.
+The `sbb-card-badge` component can be used via slot to display a badge in the upper right corner.
```
kyubisation, line 4: `via slot` can maybe be removed or rephrased, as the slotting is done internally and is therefore just an implementation detail the consumer does not need to know.

src/components/sbb-card/readme.md
```diff
@@ -1,49 +1,78 @@
...
+The badge is hidden with card sizes `xs` or `s`.
 
-There are six size variants available (from `xs` to `xxl`), which can be configured via the `size` attribute value,
-and two background variants, white by default and milk when the `color` attribute is set to `milk`.
-Set the `active` attribute to display the card in an active state, which is bordered and has no hover effect.
+## Action
+
+To add an action to a card, add a `<sbb-card-action>` to the main slot. With the `<sbb-card-action>` 
+all the card area becomes clickable. 
+Consult [API docs of `sbb-card-action`](/docs/components-sbb-card-sbb-card-action--docs) for detailed information.
```
kyubisation, line 11: There is not a lot of information in the action documentation? Is that intentional?
If so, the phrasing could maybe be adapted?

kyubisation (approved):
LGTM 👍 
Amazing work!

src/components/sbb-card/readme.md
jeripeierSBB: What I meant is that the api documentation of props is there. Do you have a suggestion how to write it best?

PR #1738, opened 2023-05-10
refactor: migrate to new English named pictograms

jeripeierSBB:
BREAKING CHANGE:
The `picto:`-namespace is now pointing to the new English named icons. If the old german named icons should be used, you can access them by using the namespace `picto-legacy:` (deprecated).


kyubisation (approved):
LGTM 👍

PR #1744, opened 2023-05-15
refactor(sbb-tab-group): remove sbb-tab-amount, provide amount and iconName properties

jeripeierSBB:
BREAKING CHANGE:
removed `sbb-tab-amount` component which can be replaced by property `amount` or slot `amount` of `sbb-tab-title` component.
e.g. `<sbb-tab-amount>123</sbb-tab-amount>` becomes `<span slot="amount">123</span>` or alternatively `<sbb-tab-title amount="123">Label</sbb-tab-title>`.


kyubisation (approved):
LGTM 👍 
Thanks! 🙇

dauriamarco (approved):
LGTM, thank you! 👍 
Just the animation seems a bit slower, surely slower than the tags. But maybe it's my impression or it should be that way.

PR #1748, opened 2023-05-17
fix: avoid reading all overlay content for screen readers

### LINKED ISSUE #1595: [A11y] Navigation confusing for Screenreaders
## Detailed description

According to our first accessibility review by Mürsel Genc, the whole is always read fully (one after the other) when opening the navigation. Instead, only the focused element should be output and the user should navigate the elements using focus.

image

See also: https://flow.sbb.ch/browse/SBBCH-15183

Note by @kyubisation: Evaluate whether we really should use the `<dialog>` element for `<sbb-navigation>`, as this might be what is causing this behavior.

jeripeierSBB:
Closes #1595


src/components/sbb-navigation/sbb-navigation.tsx
```diff
@@ -377,6 +377,7 @@ export class SbbNavigation implements ComponentInterface {
             aria-label={this.accessibilityLabel}
             onAnimationEnd={(event: AnimationEvent) => this._onAnimationEnd(event)}
             class="sbb-navigation"
+            role="group"
```
kyubisation, line 380: Is this `group` due to `aria-label`?

kyubisation (approved):
LGTM 👍

src/components/sbb-navigation/sbb-navigation.tsx
jeripeierSBB: Yes, as it reads the aria label, which it doesn't when it don't have a specific role. as group is very neutral, we decided to go with it

PR #1752, opened 2023-05-17
fix: chromatic and force-open tests

dauriamarco:
Fix incorrect chromatic snapshots and force-open failing tests.


jeripeierSBB (approved):
LGTM

PR #1758, opened 2023-05-22
fix: set most recent input modality initial value

### LINKED ISSUE #1757: bug: the tooltip trigger shows the focus ring on tooltip close if no mouse event has happened on the page yet

dauriamarco:
Closes #1757.

### DIFF
--- a/src/global/helpers/eventing/input-modality-detector.ts
+++ b/src/global/helpers/eventing/input-modality-detector.ts
@@ -76,7 +76,7 @@ class SbbInputModalityDetector {
   public get mostRecentModality(): SbbInputModality | null {
     return this._mostRecentModality;
   }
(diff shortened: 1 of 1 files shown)


src/global/helpers/eventing/input-modality-detector.ts
```diff
@@ -76,7 +76,7 @@ class SbbInputModalityDetector {
   public get mostRecentModality(): SbbInputModality | null {
     return this._mostRecentModality;
   }
-  private _mostRecentModality: SbbInputModality | null = null;
+  private _mostRecentModality: SbbInputModality | null = 'mouse';
```
jeripeierSBB, line 79: should we add a test case somehow and make a comment here why we use mouse as a default?

jeripeierSBB (approved):
LGTM

PR #1767, opened 2023-05-23
fix(sbb-selection-panel): allow nested checkbox and radio groups

### LINKED ISSUE #1768: Bug: support radio group in content of selection panel
Current problem: radio group of selection panel manages also radio buttons of inner radio group, but they should be excluded.

jeripeierSBB:
Closes #1768


kyubisation (approved):
LGTM 👍

PR #1772, opened 2023-05-24
feat(sbb-autocomplete): favourites PoC

TomMenga:
## Changes

Implemented a POC for the autocomplete with favourites component

## Known a11y problems
The ARIA pattern we are using is not broadly used and presents some problems:

- VoiceOver + Chrome => Reads the content of the option only if the previous one has finished speaking (it is needed to navigate slowly)
- VoiceOver on Safari and iOS is not supported
- VoiceOver + Chrome => Groups labels are not read
- JAWS + Firefox => Groups labels are not read
- Android + TalkBack => Groups labels are not read


jeripeierSBB:
Continued in #2512

PR #1787, opened 2023-05-31
fix(sbb-train): improve accessibility by replacing `span` with heading tag

### LINKED ISSUE #1771: [Bug]: SbbTrainFormation - direction ignored by SR
### Preflight Checklist

### Bug type

Accessibility

Components

0.33.10

### Operating system

Windows 10

### Browser / Browser version

Chrome

### Input mode

Mouse/Keyboard

### Additional settings

_No response_

### What happened?

Screenreaders ignores the direction of the train (Tested with nvda and voiceover)

1. Navigate trought the train-formation with the arrow keys

### Relevant log output

_No response_

jeripeierSBB:
Closes #1771

### DIFF
--- a/src/components/sbb-train/sbb-train.spec.ts
+++ b/src/components/sbb-train/sbb-train.spec.ts
@@ -12,7 +12,7 @@ describe('sbb-train', () => {
       <sbb-train direction-label="Driving direction" station="Bern" direction="left">
         <mock:shadow-root>
           <div class="sbb-train">
-            <sbb-title visually-hidden level="6">Train, Driving direction Bern.</sbb-title>
+            <h6 class="sbb-train__direction-label-sr">Train, Driving direction Bern.</h6>
             <ul class="sbb-train__wagons" aria-label="Coaches of the train"></ul>
             <span hidden>
               <slot />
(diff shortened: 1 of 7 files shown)


src/components/sbb-train/sbb-train.tsx
```diff
@@ -30,6 +31,9 @@ export class SbbTrain implements ComponentInterface {
   /** General label for "driving direction". */
   @Prop() public directionLabel!: string;
 
+  /** Heading level of the direction label, used for screenreaders. */
+  @Prop() public directionLabelLevel: InterfaceTitleAttributes['level'] = '3';
```
jeripeierSBB, line 35:
```suggestion
  @Prop() public directionLabelLevel: InterfaceTitleAttributes['level'] = '6';
```

src/components/sbb-train/sbb-train.spec.ts
```diff
@@ -12,7 +12,7 @@ describe('sbb-train', () => {
       <sbb-train direction-label="Driving direction" station="Bern" direction="left">
         <mock:shadow-root>
           <div class="sbb-train">
-            <span class="sbb-screenreaderonly">Train, Driving direction Bern.</span>
+            <sbb-title visually-hidden level="6">Train, Driving direction Bern.</sbb-title>
```
kyubisation, line 15: Is it necessary to use an `sbb-title` here or would a h1-6 be enough?

kyubisation (approved):
LGTM 👍 
One question about title usage

src/components/sbb-train/sbb-train.spec.ts
jeripeierSBB: @kyubisation No it's not necessary, my thought was the chance that the title component is anyways loaded in a page context is very high and so there is less payload than all the rules of the screenreader only mixin. But I can easily change it if you want?

PR #1792, opened 2023-06-02
refactor: convert storybook stories to TypeScript

kyubisation:


convenience/generate-component/boilerplate/component.stories.tsx
```diff
@@ -1,24 +1,27 @@
...
-Story1.args = {
-  'some-prop': 'opt1',
+export const Story1: StoryObj = {
+  render: Template,
+  args: {
+    'some-prop': 'opt1',
+  },
 };
 
 export default {
```
jeripeierSBB, line 17: With default export we lose the type information from Meta

jeripeierSBB (changes requested):
thanks for the hard work :) one little thing to address in my opinion

jeripeierSBB (approved):
LGTM

PR #1804, opened 2023-06-08
feat(sbb-datepicker): provide `setValue` and `getValue` on `sbb-datepicker`

jeripeierSBB:


src/components/sbb-datepicker/readme.md
```diff
@@ -14,7 +14,10 @@ date, the `sbb-invalid` class is added to the input. The component also listens
  `wide` and `dateFilter`, and emits a `datePickerUpdated` event when changed.
 
 Consumers can listen to the native `change` event on the `sbb-datepicker` component to intercept the date change `event`;
-the current value can be read from the async method `event.target.getValueAsDate()`.
+the current value can be read from the async method `event.target.getValueAsDate()` or `event.target.getValue()`.
+In order to set the value programmatically it's recommended to use the `setValue()` or `setValueAsString()`
+method of the `<sbb-datepicker>` as setting it only on the `<input>` does not trigger a change event
+and so interpreting of the date can be missed.
```
MarioCastigliano, line 20:
```suggestion
method of the `<sbb-datepicker>` as setting it only on the `<input>` does not trigger a change event. 
```

I'm not sure if this is correct because the mutation observer calls the _getValidValue method whenever it detects a value change. Has this ever happened before?

jeripeierSBB: Hmm yes I saw this problem yesterday debugging with Oliver from Draconis in react context. But the onchange on the input does never fire when changed programmatically, but you mean maybe the change on the datepicker does?

MarioCastigliano: It's normal that changing the value on the input doesn't trigger the onchange (as per native behavior). Oliver himself said that he couldn't use the setValueAsDate method because that would've triggered a loop. In this case I'm referring specifically to the bit where you said "interpreting of the date can be missed", because there's a specific logic to ensure the date is interpreted every time the value is changed programmatically (even without the onchange firing)

MarioCastigliano (approved):
LGTM

kyubisation (approved):
LGTM 👍

PR #1805, opened 2023-06-08
fix(sbb-form-field): fix floating label support for popup components

jeripeierSBB:


src/components/sbb-form-field/sbb-form-field.scss
```diff
@@ -190,7 +190,17 @@
...
   }
 
-  :host([floating-label]:not([floating-label='false'], :focus-within)[data-input-empty]) & {
+  // If floating-label is activated and there is no focus in it (except for select) and no popup is open
+  // and input is empty
+  // then apply the label transition.
+  :host(
+      [floating-label]:not(
+          [floating-label='false'],
+          :focus-within:not([data-input-type='sbb-select']),
```
TomMenga, line 199: Why the exception on the select?
I expected that the `data-has-popup-open` would suffice :|

TomMenga (approved):
LGTM! Thanks 👍

PR #1807, opened 2023-06-12
fix: enable pressing enter on action elements in Firefox

kyubisation:
Previously it was not possible in Firefox for an action element with an internal anchor element (e.g. `<sbb-button href="...">` to be able to "click" it with an Enter press. This PR fixes this by switching from a `PointerEvent` to a `MouseEvent`.


dauriamarco (approved):
LGTM 👍

PR #1810, opened 2023-06-13
feat(sbb-toast): component implementation

### LINKED ISSUE #1026: story(toast): initial implementation

TomMenga:
## Preflight Checklist

## Issue

This PR Closes #1026 

## Pull request checklist

## Changes

The following components were implemented

* sbb-toast

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [x] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


kyubisation (commented):
Nice work so far! 😃

src/components/sbb-toast/sbb-toast.custom.d.ts
```diff
@@ -0,0 +1,11 @@
+export type ToastPosition =
```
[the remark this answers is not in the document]
TomMenga: I like it :D
Didn't know you could do that though

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -0,0 +1,260 @@
...
+      this._state = 'closed';
+      this.didClose.emit();
+    }
+  }
+
+  /**
+   * Since we do not stack toasts, we force the closing on other existing opened toasts
+   */
+  private async _closeOtherToasts(): Promise<void> {
+    document.querySelectorAll('sbb-toast').forEach(async (t) => {
```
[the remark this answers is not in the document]
TomMenga: Initially, I thought if we should support toasts in shadow DOM but, in the end, I couldn't find a realistic case where this would be needed. But, since you pointed it out, it doesn't hurt to handle those cases too. I will apply your solution :)

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -0,0 +1,260 @@
...
+      if ((await t.getState()) === 'opened') {
```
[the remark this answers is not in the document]
TomMenga: True. Could a method to check for status be useful for consumers? Since they don't have an "official" way to check if the toast is open or closed.

src/components/sbb-toast/sbb-toast.tsx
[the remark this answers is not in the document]
TomMenga: Hello @kyubisation, 
unfortunately, the "WeakRef" class has been added in ES2021.
Since we are targeting the ES2017, we need to add a reference to a part of the new lib.
I don't fully understand the consequences of this change, do you see a problem that I don't?

image

src/components/sbb-toast/sbb-toast.scss
```diff
@@ -0,0 +1,161 @@
...
+
+  :host([position='top-right']) &,
+  :host([position='top-end']) & {
+    justify-content: flex-end;
+    align-items: flex-start;
+  }
+}
+
+.sbb-toast {
+  z-index: var(--sbb-toast-z-index);
```
jeripeierSBB, line 70: Allow multilevel redefinition
```suggestion
  z-index: var(--sbb-toast-z-index, var(--sbb-overlay-z-index));
```

src/components/sbb-toast/sbb-toast.scss
```diff
@@ -0,0 +1,161 @@
...
+.sbb-toast__content {
+  @include sbb.title-6($exclude-spacing: true);
+
+  flex: 1 1 0;
+}
+
+.sbb-toast__action {
+  position: relative;
+
+  // Since the action button has an absolute position, we have to "reserve" space
```
jeripeierSBB, line 116: I think the size is not fixed for all screen sizes

src/components/sbb-toast/sbb-toast.scss
```diff
@@ -0,0 +1,161 @@
...
+  min-width: #{sbb.px-to-rem-build(40)};
+
+  :host(:not([data-has-action])) & {
+    display: none;
+  }
+}
+
+::slotted(sbb-button),
```
jeripeierSBB, line 124: Should we really target with ::slotted or not just targeting "span" or whatever which is around?

src/components/sbb-toast/sbb-toast.scss
```diff
@@ -0,0 +1,161 @@
...
+.sbb-toast__action-button {
+  margin-inline: var(--sbb-spacing-fixed-2x);
+  position: absolute;
```
jeripeierSBB, line 127: why positioning absoute?

src/components/sbb-toast/sbb-toast.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+  <Fragment>
+    <sbb-button onClick={() => document.querySelector('sbb-toast').open()}>Show toast</sbb-button>
+    <sbb-toast
+      position={`${args.verticalPosition}-${args.horizontalPosition}` as any}
+      dismissible={args.dismissible}
+      timeout={args.timeout}
+      politeness={args.politeness}
+      icon-name={args.iconName}
+      disable-animation={args.disableAnimation}
+      data-testid="sbb-toast"
```
jeripeierSBB, line 101: would it not more stable to just ust {...args} ? for the position you could still concatenate it. But I'm also thinking if it is only one propery (position), it's better to also only provide one control for it with all the options in it. or otherwise we should make two props in the component

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -0,0 +1,260 @@
...
+ * @slot unnamed - Use this to document a slot.
+ */
+@Component({
+  shadow: true,
+  styleUrl: 'sbb-toast.scss',
+  tag: 'sbb-toast',
+})
+export class SbbToast implements ComponentInterface {
+  /** The length of time in milliseconds to wait before automatically dismissing the toast. */
+  @Prop() public timeout = 6000;
```
jeripeierSBB, line 36: I think it should be documented, that setting to "0" you can opt out of a timeout

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  /**
+   * The icon name we want to use, choose from the small icon variants from the ui-icons category from here
+   * https://lyne.sbb.ch/tokens/icons/.
+   */
```
jeripeierSBB, line 41:
```suggestion
  /**
   * The name of the icon, choose from the small icon variants
   * from the ui-icons category from here
   * https://icons.app.sbb.ch.
   */
```

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -0,0 +1,260 @@
...
+            <div class="sbb-toast__action">
+              <slot name="action" onSlotchange={(event) => this._onActionSlotChange(event)}>
+                {this.dismissible && (
+                  <sbb-button
+                    class="sbb-toast__action-button"
+                    icon-name="cross-small"
+                    variant="transparent"
+                    negative={true}
+                    size="m"
+                    aria-label={i18nCloseDialog[this._currentLanguage]}
```
jeripeierSBB, line 249: not sure if i18nCloseDialog describes it best, but it's for sure not wrong :)

src/components/sbb-toast/sbb-toast.tsx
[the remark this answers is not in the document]
TomMenga: For now remove the api

src/components/sbb-toast/sbb-toast.tsx
TomMenga: The choice was between i18nCloseDialog and i18nCloseAlert. Or maybe create a new one :/
What do you think?

src/components/sbb-toast/sbb-toast.scss
TomMenga: I've tested all the resolutions and the "needed" space is constant. 
Mainly because the padding changes with screen resolution and compensates for the change in size of the button

src/components/sbb-toast/sbb-toast.scss
TomMenga: That style is specifically meant for slotted sbb-button.
The alternative would be to check for slotted content type and add a class or a data-attribute to target the container only when a button is contained. Not sure it's worth :/
Or maybe there is another way to do that I do not know?

src/components/sbb-toast/sbb-toast.stories.tsx
TomMenga: I thought to have two different controls on storybook because it seemed more convenient to use them.
On the other hand, a single input is more convenient from a code perspective.

But you might be right, it might be misleading and maybe we should have a single control in storybook too. What do you think?

src/components/sbb-toast/sbb-toast.scss
TomMenga: Because the sbb-button doesn't fit in the toast unless it ignores the container spacing. 
It was a suggestion that came out in an initial call with Lukas.
Now that you're pointing it out, maybe it is better to work with negative margins to handle this case.
I'll try to implement it and see which one seems less ugly :D

jeripeierSBB: Ok I understand, please also check the alert component as there is the same problem. Maybe we can align the implementation :)

src/components/sbb-toast/sbb-toast.stories.tsx
jeripeierSBB: I think I‘d prefer s single control.

src/components/sbb-toast/sbb-toast.scss
jeripeierSBB: Ok I understand. So you can keep it as it is :-)

src/components/sbb-toast/sbb-toast.tsx
jeripeierSBB: I agree :-)

src/components/sbb-toast/sbb-toast.scss
jeripeierSBB: Ok, sounds magic 😅

src/components/sbb-toast/sbb-toast.stories.tsx
TomMenga: Got it. This way we can go back using {...args} too

src/components/sbb-toast/sbb-toast.scss
TomMenga: Already checked it :)
In that case, it is cleaner to use negative margins since there's no impediment that we, unfortunately, have on the toast.
The impediment has to do with negative bottom-margin ( if you'd like to deep down on it we can have a call )

src/components/sbb-toast/sbb-toast.e2e.ts
```diff
@@ -82,6 +77,37 @@ describe('sbb-toast', () => {
     expect(element.getAttribute('data-state')).toEqual('closed');
   });
 
+  it('closes by marked action element', async () => {
+    page = await newE2EPage({
+      html: `
+      <sbb-toast>
+        <sbb-button slot="action" sbb-toast-close/>
```
jeripeierSBB, line 84: the sbb-toast-close attribute should be described in docs, not?

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -150,17 +158,40 @@ export class SbbToast implements ComponentInterface {
     this.willClose.emit();
   }
 
-  @Method() public async getState(): Promise<SbbOverlayState> {
-    return this._state;
+  @Listen('click') public async onClick(ev: Event): Promise<void> {
+    if ((ev.target as HTMLElement).hasAttribute('sbb-toast-close')) {
```
jeripeierSBB, line 162: I think it would be a little bit more safe to do the same approach as in tooltip by reading the composedPath. What do you think?

```
event
      .composedPath()
      .find(
        (e) => e instanceof window.HTMLElement && e.hasAttribute('sbb-toast-close')
      ) as HTMLElement;
```

src/components/sbb-toast/sbb-toast.tsx
```diff
@@ -150,17 +158,40 @@ export class SbbToast implements ComponentInterface {
     this.willClose.emit();
   }
 
-  @Method() public async getState(): Promise<SbbOverlayState> {
-    return this._state;
+  @Listen('click') public async onClick(ev: Event): Promise<void> {
```
jeripeierSBB, line 161: completely nitpicky comment, but so far we used `evt` and `event`. `ev` would be new (not completely, it's in autocomplete as well :P . But I let it up to you :D
```suggestion
  @Listen('click') public async onClick(event: Event): Promise<void> {
```

TomMenga: I'll change those, no problem :)

src/components/sbb-toast/sbb-toast.e2e.ts
TomMenga: It is described D: (I'll improve it a bit)
image

src/components/sbb-toast/sbb-toast.tsx
jeripeierSBB: It would include also shadow dom cases. e.g. in sbb-tooltip we applied "sbb-tooltip-close" also to the internal close button (could be done here as well). the parents should better not be included as you said. Lukas and I had the idea of creating a helper function doing this. something like `composedPathHasAttribute(event, attributeName, endElement)`. This function would filter out all elements after/above endElement and then check for an attribute. And as you mentioned it would also include setting sbb-tooltip-close on in between elements, like on a container div containing several actions.

TomMenga: Thanks for the explanation :D
I'll align the implementation to the tooltip

jeripeierSBB: I think we had a little misunderstanding here, can you create a helper function for both the tooltip and toast as following: `composedPathHasAttribute(event, attributeName, endElement)` (or similar). The function should exclude all parents (starting from endElement which would be sbb-tooltip, or sbb-toast). This fixes a rare case issue where parent elements are searched too and helps avoiding code duplication. Thanks!

TomMenga: My bad :( 
I thought that the helper method was a future refactoring. I'll fix this asap

src/global/helpers/composed-path-has-attribute.ts
```diff
@@ -10,7 +10,7 @@ export function composedPathHasAttribute(
   event: Event,
   attribute: string,
   endElement = null
-): EventTarget {
+): HTMLElement {
```
jeripeierSBB, line 13: Sorry, hopefully last comment, but it can also be null which should be reflected in types

jeripeierSBB (approved):
LGTM, thanks! Let's wait for Manuels approval

jeripeierSBB (approved):
Thanks for the improvements! LGTM, waiting for reply of Mürsel

PR #1811, opened 2023-06-13
refactor: adapt chromatic check imports

kyubisation:
With the new patch release of the `chromatic` package, the imports from `chromatic/isChromatic` no longer work and need to be changed to `chromatic`, due to the `exports` configuration in the `chromatic` package.


TomMenga (approved):
LGTM! 👍

PR #1850, opened 2023-06-29
fix(sbb-button, sbb-link): rename `static` attribute to `is-static`

jeripeierSBB:
BREAKING CHANGE:
 - In order to fully support typings, `static` attribute of `sbb-link` and `sbb-button` have been renamed to `is-static` (property `isStatic`).


DavideMininni-Fincons (approved):
LGTM, thank you

kyubisation (approved):
LGTM 👍

PR #1876, opened 2023-07-14
refactor: re-structure helpers directory

kyubisation:
`global/helpers` has been split into `global/a11y`, `global/datetime`, `global/dom`, `global/eventing`, `global/observers`, `global/overlay`, `global/testing` and `global/timetable`.
Renamed imports of `AgnosticIntersectionObserver`, `AgnosticMutationObserver` and `AgnosticResizeObserver` have been changed to directly use the Symbol with `Agnostic*`, as it is less error prone when accidentally removing the import (as no error would be thrown when using e.g. MutationObserver, as it is a valid symbol).


src/global/a11y/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './arrow-navigation';
```
jeripeierSBB, line 1: for me it's not really clear why those files belong to a11y. focus and arrow navigation is a standard for me and nothing special a11y related. and assign-id is often used to help reference something, but also for me for itself has nothing to do with a11y. hmm :D

jeripeierSBB (commented):
This task was on my todo list for after the card refactoring, so thanks for doing my tasks :D

I am not sure about some categories, maybe we can discuss this via Teams:
 - a11y
 - dom
 - eventing
 - interfaces

Those categories are somehow not self describing. you can e.g. have events for datetime or timetable. etc... maybe we find categories which are more logic and therefore it would be easier to decide where to put future "helpers".

jeripeierSBB:
I forgot to mention that I really like this refactoring :)

jeripeierSBB (approved):
LGTM

PR #1893, opened 2023-07-20
fix(sbb-time-input): fix accessibility with structure refactoring

### LINKED ISSUE #1888: [Bug]: time-input - aria-label not read by screenreader
### Preflight Checklist

### Bug type

Accessibility

Components

0.35.5

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

Mouse/Keyboard

### Additional settings

_No response_

### What happened?

The aria-label of the time-input does not get read by screenreaders. The placeholder (HH:MM) is read instead.

1. Set aria-label property of time-input
2. Tab to it and listen to the screenreader

### Relevant log output

_No response_

jeripeierSBB:
Closes #1888

To fix accessibility, we had to extract the input from the shadow DOM into the light DOM. Now consumers have to provide their own native `input` element and link it to the `sbb-time-input`. With this new structure, the `sbb-time-input` is also aligned to the `sbb-datepicker` which has the same structure.

BREAKING CHANGE:
 - Consumers have to provide their own native `input` element and link it to the `sbb-time-input` (e.g. ```<input id="input-id"><sbb-time-input input="input-id"/>```. If it's used inside an `sbb-form-field`, the connection is made automatically (no id reference necessary).
 - The following properties of `sbb-time-input` were removed: `value`, `valueAsDate`, `form`, `readonly`, `disabled` and `required`. They can directly be used on the `input` except `valueAsDate` which was replaced by `getValueAsDate()` method on `sbb-time-input`.


src/global/dom/input-element.ts
```diff
@@ -1,16 +1,23 @@
+import { findReferencedElement } from './find-referenced-element';
+
 /**
  * Resolves the input element inside the shadow DOM of the given element.
  */
-export function inputElement(element: HTMLElement): HTMLInputElement | null {
+export function shadowInputElement(element: HTMLElement): HTMLInputElement | null {
```
kyubisation, line 6: Should this be named `findShadowInput` to be consistent with the other function?

kyubisation (commented):
Nice work! 😃 
(Documentation is missing, which I assume you are still working on.)

WalkingOS:
Hey, nice work so far :)

The only thing I noticed:

-> if you have an invalid input and you type something wrong again, the SR would read "Time changed to ..."

jeripeierSBB:
> Hey, nice work so far :)
> 
> The only thing I noticed:
> 
> -> if you have an invalid input and you type something wrong again, the SR would read "Time changed to ..."

Great catch, pushed a fix

PR #1910, opened 2023-08-02
fix(sbb-dialog): close stacked dialogs one by one on ESC press

### LINKED ISSUE #1904: bug(sbb-dialog): if dialogs are stacked, pressing escape key closes all dialogs instead of only the one on top
`sbb-dialog` should register stack context and only close the top `sbb-dialog` instance.
Suggestion: When opening a `sbb-dialog` add it to a global array of opened `sbb-dialog` instances and when it is closed, it should be removed from the array. When ESC is pressed, only the last added dialog should be closed.

dauriamarco:
Closes #1904

### DIFF
--- a/src/components/sbb-dialog/sbb-dialog.tsx
+++ b/src/components/sbb-dialog/sbb-dialog.tsx
@@ -29,6 +29,8 @@ import {
 } from '../../global/eventing';
 import { AgnosticResizeObserver } from '../../global/observers';
 
+// A global collection of existing dialogs
+const dialogRefs: HTMLSbbDialogElement[] = [];
 let nextId = 0;
 
 /**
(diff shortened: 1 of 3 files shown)


src/components/sbb-dialog/sbb-dialog.tsx
```diff
@@ -313,8 +322,9 @@ export class SbbDialog implements ComponentInterface {
       this._windowEventsController?.abort();
       this._focusTrap.disconnect();
       this._dialogContentResizeObserver.disconnect();
-      // Enable scrolling for content below the dialog
-      this._scrollHandler.enableScroll();
+      this._removeInstanceFromGlobalCollection();
+      // Enable scrolling for content below the dialog if no dialog is open
+      !dialogRefs.length && this._scrollHandler.enableScroll();
```
jeripeierSBB, line 327: For what scenario is this check needed?

jeripeierSBB (changes requested):
Could you add some tests testing stacked dialogs? Thanks :)

src/components/sbb-dialog/sbb-dialog.tsx
dauriamarco: Without the check, after you close a stacked dialog the scroll on the page will be enabled again even if there is another dialog or dialogs still open.

jeripeierSBB: Ah I see. Was misreading once more the "&&" which calls a function and is not just a check...

jeripeierSBB (approved):
LGTM

PR #1922, opened 2023-08-08
fix(sbb-form-field): fix accessibility when slotting form error

jeripeierSBB:
Previously, aria-describedby of the input was not correctly updated when removing or first adding a sbb-form-error. Additionally, to support removing and adding sbb-form-error multiple times for screen readers, we had to move the aria-live region to a single element, although according to documentation the aria-live region should be on a container.

### DIFF
--- a/src/components/sbb-form-field/sbb-form-field.tsx
+++ b/src/components/sbb-form-field/sbb-form-field.tsx
@@ -24,6 +24,7 @@ import {
 } from '../../global/eventing';
 
 let nextId = 0;
+let nextFormFieldErrorId = 0;
 
 const supportedPopupTagNames = ['SBB-TOOLTIP', 'SBB-AUTOCOMPLETE', 'SBB-SELECT'];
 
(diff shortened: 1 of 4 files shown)


src/components/sbb-form-field/sbb-form-field.tsx
```diff
@@ -367,16 +368,38 @@ export class SbbFormField implements ComponentInterface {
...
 
   private _applyAriaDescribedby(): void {
-    const value = this._errorElements.length
-      ? this._errorElements.map((e) => e.id).join(' ')
-      : this._originalInputAriaDescribedby;
-    if (value) {
-      this._input?.setAttribute('aria-describedby', value);
+    const ids = [];
+
+    if (this._originalInputAriaDescribedby) {
```
DavideMininni-Fincons, line 390: Just a quick question: 
in the original code the _originalInputAriaDescribedby OR the joined id are used, while now they can both be used. 
It is correct?

DavideMininni-Fincons (commented):
Looks good to me, a single comment added

src/components/sbb-form-field/sbb-form-field.tsx
jeripeierSBB: Yes I compared to Angular implementation and it make sense in my mind as well that whatever original aria-describedby is never removed

jeripeierSBB: Do you have concerns?

DavideMininni-Fincons: No, not at all, i just saw the difference so I asked :) can be merged for me

DavideMininni-Fincons (approved):

dauriamarco (approved):
LGTM 👍

PR #1928, opened 2023-08-09
fix(sbb-notification): handle stories disableAnimation correctly

### LINKED ISSUE #1926: bug(sbb-notification): disabling animation in storybook broken (maybe also in real?)
transitionend event does not fire with 0s timing, we should check if this has implications in sending didOpen or didClose event.

dauriamarco:
Closes #1926


jeripeierSBB (approved):
LGTM :)

PR #1954, opened 2023-08-30
refactor(sbb-message): rename `sbb-no-results` to `sbb-message`

jeripeierSBB:
BREAKING CHANGE:
- `sbb-no-results` was renamed to `sbb-message`


kyubisation (approved):
LGTM 👍

PR #1956, opened 2023-08-30
build: increase test execution memory

kyubisation:
Increase memory from 4GB to 6GB for test execution

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -52,7 +52,8 @@ jobs:
       - run: yarn install --frozen-lockfile --non-interactive
 
       - name: Run tests
-        run: yarn test:prod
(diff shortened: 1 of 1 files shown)


jeripeierSBB:
Seemed not to work :(

.github/workflows/continuous-integration.yml
```diff
@@ -47,8 +52,9 @@ jobs:
       - run: yarn install --frozen-lockfile --non-interactive
 
       - name: Run tests
-        # Try yarn test:prod up to three times, if it fails and exit with the exit code from the last execution
```
jeripeierSBB, line 50: I today triggered several re runs and no one had a "operation cancelled" exit. So it could be that your fix helps. But I'd suggest to re add our retry mechanism, as we still have some failing builds from time to time

jeripeierSBB (approved):
LGTM. please merge :P

PR #1965, opened 2023-09-04
fix(sbb-navigation-section): use divider instead of separation border

### LINKED ISSUE #1934: story(sbb-navigation): use divider to divide navigation from section which will have some color changes

jeripeierSBB:
Closes #1934 

Has a one pixel shift of whole navigation section, but seems like designed in Figma


dauriamarco (approved):
LGTM 👍

PR #1988, opened 2023-09-18
build: implement tool for partially automated Lit migration of components

### LINKED ISSUE #1831: story(proof of concept): Lit implementation
### Feature Description

We would like, as proof of concept, to implement some components with lit-element.

e.g. integrate the `time input` component with Lit

### References

https://lit.dev/

### Tasks

- [x] Create new repository
- [x] Implement a few components with lit-element
  - [x] time input component as a native form component (without slotting an input element)
  - [x] a group of components that are tightly coupled (maybe sbb-radio-button and sbb-radio-button-group)
- [ ] Topics to evaluate
  - [x] Getter/Setter
  - [x] Native form support [Example](https://www.dannymoerkerke.com/blog/web-components-can-now-be-native-form-elements/)
  - [x] Lifecycle methods/hooks
  - [x] JSX => html tagged template function
  - [x] Inheritence/Mixins
  - [ ] Hydration (e.g. our .hydrate css rules and SSR)
  - [ ] Support Css in tests ( need to implement a custom plugin for it )
  - [ ] React integration
  - [ ] Distribution format
- [x] Present and discuss findings with the team

### LINKED ISSUE #1949: story(lit): Migration script
### Feature Description

We need to implement a script to migrate components from stencil.js to lit automatically.
Following, the task list divided by file:

# [Component migration checklist](#component-anchor)

## Imports
- [x] Remove @stencil imports
- [x] Add common lit imports
  - [x] `import { LitElement, PropertyValues, html, nothing } from 'lit';`
  - [x]  `import { customElement, property, state, ... } from 'lit/decorators.js';`
- [x] Import scss file => `import Style from './${componentName}.scss';`
- [ ] Replace all occurrences of `HTMLSbb*Element` with `Sbb*` and import `import { type Sbb* } from '../*';`

## Class declaration
- [x] `@Component({ ... , tag: ${compTag} })` => `@customElement('${compTag}')`
- [x] Ignore the other `@Component` properties
- [x] `implements ComponentInterface` => `extends LitElement`
- [x] Right after the class declaration, add `static override styles = Style;`

## Properties
- [x] `@Prop()` => `@property()`
- [ ] Prop options
  - [x] Ignore `mutable`. Properties are always mutable in lit
  - [x] `reflect` and `attribute` remains the same
  - [x] If the property name is camelCase => add `@property({ ... , attribute: ${ propName.toKebabCase() } })`
  - [ ] If the property type is !== string => add `@property({ ... , type: ${ propType } } })` ( [Supported types](https://lit.dev/docs/components/properties/#conversion-type) )
- [x] `@State()` => `@state()`

## `@Element()`
- [x] Remove the `@Element() private _element!: HTMLElement;`
- [x] Replace `this._element` => `this`

## `@Event`
 _TODO: confirm the EventEmitter signature_
- [x] If `eventName` is not specified in `{ props }`
  - [x] `@Event({ props }) public customEvent` => `public customEvent = new EventEmitter( '${ customEvent }', {...props})`
- [x] Else
  - [x] `@Event({ eventName: 'change', props }) public customEvent` => `public customEvent = new EventEmitter( '${ props.eventName }', {...props})`. 

## Method
- [x] Remove `@Method`
- [x] Remove `async` and the `Promise` return type
- [ ] _Advanced:_ We could probably remove 99% of the `awaits` 

## `@Listen`
There is no equivalent in Lit. Luckily, we only have basic usages in our components.
- [x] Remove the `@Listen('event-name', {props})`
- [x] Inside the constructor, add `this.addEventListener('event-name', (ev) => this.method(ev), {props});`
  - [ ] _Advanced:_ We could retrieve the event type from the method signature
  - [x] _TODO Discussion:_ We could add the listener in the constructor or the connectedCallback. The difference is that the connectedCallback needs to remove the listeners. 

## LifeCycle callbacks
- [x] Add call to `super()` in the constructor and the connected/disconnected Callback
- [x] For each callback **add the call to the super**
- [x] `componentWillLoad()` has no equivalent. It might be converted to `connectedCallback` or `firstUpdated` but needs the human intervention
- [x] `componentDidRender()` and `componentDidUpdate()` => `protected override updated(changedProperties: PropertyValues<this>)`
- [x] `componentDidLoad()` => `protected override firstUpdated(changedProperties: PropertyValues<this>)`

## Watch
There's no perfect equivalent to the `@Watch()` decorator. The best effort conversion is to use the `WillUpdate()`.  
The main difference is that the `willUpdate` will be called on initialization, while the `@Watch` is not.
- [x] Remove `@Watch('propName')`
- [x] Inside the `willUpdate()` callback, add:
```ts
willUpdate(changedProperties: PropertyValues<this>)
  ...
  if (changedProperties.has('propName')) {
    this.${watchedMethod}(this.${propName});
  }
}
```
- [x] _Optional_ Leave a comment to warn the dev of the different behavior
_Note:_ In some cases we have double `@Watch(...) @Watch(...)` (e.g. sbb-slider)

## Miscellaneous
- [x] At the end of the file, add: 
```ts
declare global {
  interface HTMLElementTagNameMap {
    '${componentTag}': ${componentClass};
// e.g.: 'sbb-time-input': SbbTimeInput;
  }
}
```

## Render
Stencil uses JSX whereas Lit uses string literals to describe the template.

### Template
- [x] `public render(): JSX.Element {` => `override render() {`
- [x] `return ( ... )` => `return html' ... ' `
- [x] Conditional template **This will be hard to migrate**
  - [x] `{ condition && template }` => `${ condition ? html' template ' : nothing }`
  - [x] `{ condition ? template_true : template_false }` => `${ condition ? html' template_true ' : html' template_false' }`

### Host
Unfortunately, there is no equivalent for the `<Host>` in Lit
- [x] Each attribute => `setAttribute(this, 'attributeName', value);` _TODO Discuss the custom setAttribute approach_
- [x] Event => look at `@Listen()` conversion.
- [x] Remove the <Host> tag
- [x] AssignId on Host `<Host ref={assignId( ... )} >`
  - [x] In the `render()` function add `assignId( ... )(this);`

### [Attributes](#component-attributes)
The main attribute difference between Stencil and Lit is the handling of null values. 
- [ ] If the expression { ... } is boolean than => ?attribute=${ ... }
- [x] attribute={ ... } => attribute=${ ... }
  - [ ] _Advanced:_ To match the exact behavior, the migration should be one of the following:
    - [ ] attribute={ ... } => attribute=${ ... ?? nothing } **equivalent to stencil**
    - [ ] attribute={ ... } => attribute=${ ifDefined( ... ) }
    - [ ] attribute={ ... } => attribute=${ ... || nothing } **This also checks for truthy values**

### Events
- [x] `onEvent={ ... }` => `@event=${ ... }`

### Refs
- [x] Add import `import { ref } from 'lit/directives/ref.js';`
- [x] `ref={ ... }` => `${ref( ... )}`

# [Storybook migration check-list](#storybook-anchor)
- [x] `import type { ... } from '@storybook/html'` => `import type { ... } from '@storybook/web-components'`
- [x] Add the component import - `import { SbbRadioButton } from './sbb-radio-button';`
- [x] Add the creation of an instance - `const a = new SbbRadioButton();`
_Discussion:_ The component instance is necessary for the compiler. Otherwise, it'll be stripped by the build.

# [Spec test migration check-list](#spec-anchor)

The most common test structure is `describe => [it]*`.
Unfortunately, there are some strange cases like the `sbb-toggle-check.spec.ts` where the structure is not the expected one.
I think the fastest way to migrate the uncommon cases is to leave them as they are and only operate on the content of each `it( ... )`

## Import section
- [x] Remove everything from `@stencil`
- [x] Add common Lit imports
  - [x] `import { expect, fixture } from '@open-wc/testing';`
  - [x] `import { html } from 'lit/static-html.js';`
- [x] Add the creation of an instance - `const a = new SbbComponent();`
_Discussion:_ The component instance is necessary for the compiler. Otherwise, the component is stripped by the build.

## Setup
- [x] Template
```ts
const { root } = await newSpecPage({
  components: [SbbToast],
  html: ` ${ test_template } `,
});
```
Migrate to:
```ts
const root = await fixture(html` ${ test_template } `);
``` 
## Assertion
- [x] Assertion
```ts
expect(root).toEqualHtml(`
  <Light-dom ... >
    <mock:shadow-root>
      <Shadow-dom .... >
    </mock:shadow-root>  
  <Light-dom ... >
`);
```
Migrate to:
```ts
expect(root).dom.to.be.equal(
  `<Light-dom ... >`
)
expect(root).shadowDom.to.be.equal(
  `<Shadow-dom .... >`
)
```

# [E2e test migration check-list](#e2e-anchor)

It is the hardest thing to migrate automatically. Manual work will probably be needed :/
The average structure is `describe => beforeEach, [it]*` but, as expected, there are several different cases.

## Import section
- [x] Remove imports from `@stencil`
- [x] Add common Lit imports
  - [x] `import { assert, expect, fixture, oneEvent } from "@open-wc/testing";`
  - [x] `import { html } from 'lit/static-html.js';`
- [x] Import the component 
  - [x] `import { SbbTimeInput } from "./sbb-time-input"; `

## Setup
- [x] `let element: E2EElement, page: E2EPage;` => `let element: ${CompClass}` (there is no `page` equivalent)
  -  Note: This rule has not been implemented. Since these variables are semantically complex, the migration will add a comment advising on how to solve them.
- [x] Template 
```ts
  beforeEach(async () => {
    page = await newE2EPage();  // Not needed anymore 
    await page.setContent(`
       ${ test_template }
    `);
    element = await page.find('sbb-time-input');
    await page.waitForChanges();  // Not needed anymore 
  });

  // Migrates to

  beforeEach(async () => {
    element = await fixture(html`
        ${ test_template }
    `);
  });
  ```
- [x] is defined
```ts
  it('renders', async () => {
    expect(element).toHaveClass('hydrated');
  });

  // Migrates to

  // This is also needed to ensure the compiler does not strip the component class
  it('is defined', () => {
    assert.instanceOf(element, ${CompClass});
  });
  ```

## Assertion & Miscellaneous
Probably, it's impossible to do a full migration because some patterns are different.
These are the instructions that are easy to translate:
- [x] `await element.setProperty('propName', value);` => `element.propName = value;
- [x] `await element.getProperty('propName'))` => `element.propName`
- [x] `await page.waitForChanges();` => `await element.updateComplete`
- [x] `await page.setContent('template')` => `await fixture('template')`
- [x] `await page.find('query');` => `document.querySelector('query')`
- [x] `await page.find('element-tag >>> query');` => `element.shadowRoot.querySelector('query')`
- [x] `await element.press('key')` => `element.focus(); await sendKeys({press: 'key'});`
- [x] `await element.type('text')` => `element.focus(); await sendKeys({type: 'text'});`
- [x] `await page.keyboard.down('key')` => `await sendKeys({down: 'key'})`
- [x] `await page.setViewport({ ... });` => `await setViewport({ ... });`
- [x] `await page.evaluate(() => {code})` => `code` (page.evaluate is not needed anymore)
- [x]  `element.triggerEvent('event', data)` => `element.dispatchEvent(new CustomEvent('event', data))`
- [x] `expect(...).toEqual(value)` => `expect(...).to.be.equal(value)`
- [x] `expect(...).toHaveClass(value)` => `expect(...).to.have.class(value)`
- [x] `expect(...).toHaveAttribute(value)` => `expect(...).to.have.attribute(value)`

## EventsSpy
The pattern to assert for events is different from the one proposed by Stencil.
To facilitate the migration, I implemented a simple "[EventSpy](https://github.com/lyne-design-system/lit-elements-poc/blob/master/src/global/helpers/testing/event-spy.ts)" class that mimics the Stencil API.
_TODO_ We should discuss whether, after the migration, we should stop using the spy and adapt to the new way to test for events.
- [x] `const eventSpy = await page.spyOnEvent('event-name');` => `const eventSpy = new EventSpy('event-name');`
- [x] `expect(eventSpy).toHaveReceivedEvent();` => `expect(eventSpy.count).to.be.greaterThan(0);`
- [x] `expect(eventSpy).not.toHaveReceivedEvent();` => `expect(eventSpy.count).not.to.be.greaterThan(0);`
- [x] `expect(eventSpy).toHaveReceivedEventTimes(number);` => `expect(eventSpy.count).to.be.equal(number);`

kyubisation:
Closes #1831
Closes #1949


TomMenga (approved):
All good! :)

PR #2011, opened 2023-09-26
chore: mitigate phishing attacks

kyubisation:
This PR adds a banner to the Storybook preview page, if viewed directly in the iframe. This is only visible in deployed versions.

image

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -30,3 +30,46 @@
     }
   }
 </style>
+
+<script>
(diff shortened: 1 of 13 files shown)


TomMenga (approved):
It is unfortunate that we cannot use text controls anymore but maybe it's reasonable to solve the 'issue'.
LGTM 👍

kyubisation:
Yeah, but I think limiting the link hrefs should be fine, as that is seldomly an area we need to debug or demonstrate.

PR #2027, opened 2023-09-29
fix(sbb-dialog): open/close methods exit conditions

dauriamarco:
Fixes open/close method exit conditions and allows interactive elements under the header to be clicked.


kyubisation (approved):
LGTM 👍

PR #2035, opened 2023-10-02
fix(sbb-header): guarantee space between logo and other header content

### LINKED ISSUE #1984: bug(sbb-header): space between logo and item not guaranteed
### Preflight Checklist

### Bug type

Other

Components

0.40.8

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

If the space becomes tight, e.g. if the username is very long, then this should be truncated. The latter is probably not supplied by Lyne, but it lacks the minimum spacing S to the left of the logo.  

https://lyne-storybook.app.sbb.ch/?path=/story/components-sbb-header-sbb-header--basic

Figma:
https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?type=design&node-id=1466%3A976440&mode=design&t=1OdrZLWOuMLX2CDC-1

1. Step one
2. Step two
3. ...

### Relevant log output

_No response_

jeripeierSBB:
Closes #1984

With this change we define the spacing between the logo and the `sbb-header-action` in Lyne wich ensures an always correct protective room of the logo. Moreover we provide two new CSS classes which make consumption easier:
 - `sbb-header-spacer`: set this CSS class to your `div` (or other element) which is the spacer between the menu trigger and the other elements inside the `sbb-header`. It handles correct growing but consumers still need to decide themself, for which breakpoints they want to hide the element (normally below large breakpoint).
 - `sbb-header-shrinkable`: set this CSS class to any `sbb-header-action` to define that if there is not enough space, ellipsis to its context should be shown. This is the wanted behavior for the user name, which can be very long.

See our updated examples.

BREAKING CHANGE:
 - The spacing between the logo and the  `sbb-header-action`s is now in responsibility of Lyne. Therefore consumers need to remove any manually added spacing. Lyne examples are updated.


src/components/sbb-header/readme.md
```diff
@@ -51,18 +47,32 @@ Finally, the following custom CSS has been added*. The result can be seen in the
...
+@media screen and (min-width < 1024px) {
+  .sbb-header-spacer {
+    display: none;
   }
 }
 ```
 
+### Content overflow
+
+If a certain `sbb-header-action` should be shrunken (receive ellipsis) when there is too less space,
```
kyubisation, line 59:
```suggestion
If a certain `sbb-header-action` should be shrunken (receive ellipsis) when there is too little space, 
```

kyubisation (approved):
LGTM 👍

PR #2036, opened 2023-10-03
refactor(sbb-time-input): lit migration

TomMenga:


src/components/sbb-time-input/sbb-time-input.spec.ts
```diff
@@ -1,22 +1,26 @@
...
-    expect(doc.querySelector('span')).toEqualHtml(`
+    expect(root).dom.to.be.equal(`
       <span>
         <sbb-time-input input="id-1">
-          <mock:shadow-root>
-            <p role="status"></p>
-          </mock:shadow-root>
         </sbb-time-input>
-        <input id="id-1" placeholder="HH:MM" type="text" maxlength="5" data-sbb-time-input>
+        <input id="id-1" placeholder="HH:MM" type="text" maxlength="5" inputmode="numeric" data-sbb-time-input>
```
jeripeierSBB, line 19: do you know why before numeric input mode was not set?

TomMenga: In this case, the method that adds it is called in a setter (immediately on creation). While, before it had to run two cycles of rendering (and it was tested after the first probably).

In general, the `await element.updateComplete` has a more generous timing and also allows child components to be fully initialized before checking for dom differences

kyubisation (approved):
LGTM 👍

PR #2044, opened 2023-10-04
fix(sbb-calendar): fix wrong width when orientation changes in iOS

jeripeierSBB:
Due to a Safari iOS rendering bug we need to define min-width as well.
Otherwise, after orientation change, there is a wrong width if placed in an sbb-dialog.


MarioCastigliano (approved):
LGTM👍

PR #2059, opened 2023-10-09
refactor(sbb-clock, sbb-link-list, sbb-footer): migrate to lit

jeripeierSBB:


src/components/sbb-clock/sbb-clock.spec.ts
```diff
@@ -1,24 +1,25 @@
...
+import { html } from 'lit/static-html.js';
+import './sbb-clock';
 
 describe('sbb-clock', () => {
   it('renders', async () => {
-    const { root } = await newSpecPage({
-      components: [SbbClock],
-      html: '<sbb-clock />',
-    });
+    const root = await fixture(html`<sbb-clock />`);
```
TomMenga, line 7: In this specific case you won't have a problem, but better be consistent imho
```suggestion
    const root = await fixture(html`<sbb-clock></sbb-block>`);
```

src/components/sbb-clock/sbb-clock.e2e.ts
TomMenga, line 1: Should we move this to the spec file?

src/components/sbb-footer/sbb-footer.spec.ts
```diff
@@ -1,24 +1,27 @@
...
+import { expect, fixture } from '@open-wc/testing';
+import { html } from 'lit/static-html.js';
 
 describe('sbb-footer', () => {
   it('renders', async () => {
-    const { root } = await newSpecPage({
-      components: [SbbFooter],
-      html: '<sbb-footer accessibility-title="Footer" />',
-    });
+    const element: SbbFooter = await fixture(html`<sbb-footer accessibility-title="Footer" />`);
```
TomMenga, line 9:
```suggestion
    const element: SbbFooter = await fixture(html`<sbb-footer accessibility-title="Footer"></sbb-footer>`);
```

TomMenga (changes requested):
Good work 👍  Just some nitpicks and a topic to discuss

PR #2064, opened 2023-10-11
fix: resolve attributes function not handling states correctly

MarioCastigliano:
Improve hostProperties function, to avoid mixed states (`aria-disabled="true" tabindex="0"`)

### DIFF
--- a/src/global/interfaces/link-button-properties.ts
+++ b/src/global/interfaces/link-button-properties.ts
@@ -114,8 +114,8 @@ function hostProperties(role: string, disabled: boolean): Record<string, string>
   return Object.assign(
     { role, dir: getDocumentWritingMode() },
     disabled
-      ? { 'aria-disabled': 'true', tabIndex: '-1' }
-      : { 'aria-disabled': 'false', tabIndex: '0' },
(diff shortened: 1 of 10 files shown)


DavideMininni-Fincons (approved):
LGTM, thanks

src/global/interfaces/link-button-properties.ts
```diff
@@ -113,7 +113,9 @@ function getLinkAttributeList(linkProperties: LinkProperties): Record<string, st
 function hostProperties(role: string, disabled: boolean): Record<string, string> {
   return Object.assign(
     { role, dir: getDocumentWritingMode() },
-    disabled ? { 'aria-disabled': 'true' } : { tabIndex: '0' },
+    disabled
+      ? { 'aria-disabled': 'true', tabIndex: '-1' }
+      : { 'aria-disabled': 'false', tabIndex: '0' },
```
jeripeierSBB, line 118: Could you check if this output generates the same as before?
```suggestion
    disabled
      ? { 'aria-disabled': 'true', tabIndex: false }
      : { 'aria-disabled': false, tabIndex: '0' },
```

TomMenga (approved):
LGTM 👍

PR #2067, opened 2023-10-12
feat(sbb-tag-group): allow other slotted elements than `sbb-tag`s

jeripeierSBB:


src/components/sbb-tag-group/sbb-tag-group.tsx
```diff
@@ -117,17 +125,24 @@ export class SbbTagGroup implements ComponentInterface {
     return Array.from(this._element.querySelectorAll('sbb-tag')) as HTMLSbbTagElement[];
   }
 
+  private _readSlottedElements(): void {
+    this._slottedElements = Array.from(this._element.children).filter(
+      (e): e is HTMLElement => e instanceof window.HTMLElement,
+    );
+  }
+
```
TomMenga, line 133: Should it be a getter like `_tags`?
What if the slotted content changes?

DavideMininni-Fincons (commented):
It looks good to me; 
should the new case be tested? maybe just changing one sbb-tag in the actual 'renders' test with another sbb- component?

src/components/sbb-tag-group/sbb-tag-group.tsx
DavideMininni-Fincons: > What if the slotted content changes?

For content changes there should be the `this._readSlottedElements()` on line 156

src/components/sbb-tag-group/sbb-tag-group.e2e.ts
TomMenga, line 1: Side note: The tag group has already been migrated to lit, we need to cherry and manually merge the differences

src/components/sbb-tag-group/sbb-tag-group.tsx
TomMenga: you're right, soz :/

MarioCastigliano:
Also I think @DavideMininni-Fincons 's suggestion about the tests is a nice idea

src/components/sbb-tag-group/sbb-tag-group.e2e.ts
jeripeierSBB: I sadly know :(,but it's kind of a blocker for one consumer, so I decided to implement now

jeripeierSBB:
@MarioCastigliano @DavideMininni-Fincons, absolutely a good idea ;) Added a simple enhancement to the tests

PR #2080, opened 2023-10-16
build: inline web test runner vite plugin

kyubisation:


jeripeierSBB (approved):
LGTM, thanks! We need just (in another PR) to stabilize some tests :)

PR #2091, opened 2023-10-18
refactor(sbb-file-selector): lit migration

TomMenga:


dauriamarco (approved):
Thank you! LGTM 👍 
Just a couple of suggestions

PR #2095, opened 2023-10-20
fix: implement overlays inert mechanism

dauriamarco:
Implements a manual inert to every element except the overlay itself to improve accessibility.

### DIFF
--- a/src/global/overlay/overlay.tsx
+++ b/src/global/overlay/overlay.tsx
@@ -26,9 +26,6 @@ const getAncestors = (overlay: HTMLElement, root: HTMLElement): HTMLElement[] =>
   let el = overlay.parentElement;
   const ancestors: HTMLElement[] = [];
   while (el) {
-    if (el.matches(IS_OPEN_OVERLAY_QUERY)) {
-      el.dataset.sbbInert = `${+el.dataset.sbbInert + 1 || 0}`;
-    }
     ancestors.push(el);
     el = el !== root && el.parentElement;
   }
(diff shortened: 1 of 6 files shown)


src/global/overlay/overlay.tsx
```diff
@@ -16,3 +18,101 @@ export function overlayGapFixCorners(): JSX.Element {
...
 }
+
+/**
+ * Returns all the ancestors of the overlay.
+ */
+const getAncestors = (overlay: HTMLElement, root: HTMLElement): HTMLElement[] => {
+  let el = overlay.parentElement;
+  const ancestors: HTMLElement[] = [];
+  while (el) {
+    if (el.matches(IS_OPEN_OVERLAY_QUERY)) {
```
kyubisation, line 29: Is it intentional, that this logic also sets sbbInert? Because this way it seems to be done twice; Once here and once afterwards on line 96.

src/global/overlay/overlay.tsx
```diff
@@ -16,3 +18,101 @@ export function overlayGapFixCorners(): JSX.Element {
...
+ */
+export function applyInertMechanism(overlay: HTMLElement): void {
+  removeSbbInert(overlay);
+
+  const overlayRoot = overlay.closest('body > *') as HTMLElement;
+  const documentChilds = Array.from(document.querySelectorAll('body > *')).filter(
+    (el) => el !== overlayRoot,
+  ) as HTMLElement[];
+
+  documentChilds.forEach((el) => setSbbInert(el));
```
kyubisation, line 83: Nitpick
```suggestion
  const documentChildren = Array.from(document.querySelectorAll('body > *')).filter(
    (el) => el !== overlayRoot,
  ) as HTMLElement[];

  documentChildren .forEach((el) => setSbbInert(el));
```

src/global/overlay/overlay.tsx
kyubisation: Ah, I think I understand now, but it's still a bit confusing. Could this maybe be done in `applyInertMechanism`?

kyubisation (approved):

jeripeierSBB (commented):
I think on disconnect we need to call remove inert as well?

src/global/overlay/overlay.tsx
dauriamarco: Since it is an ancestor, the line 96 should not apply to it, this set only the custom attr to a partent that is an overlay. I needed that for the nested overlays case and to avoid a NaN on the line 106. But I forgot to remove it as for now we don't support the nested case, so I will remove it.

dauriamarco: Or should I leave it to at least partially cover the scenario (except the shadowRoot obvs)?

kyubisation (approved):
LGTM 👍 
Again, thank you very much 🙇

PR #2106, opened 2023-10-24
fix(sbb-dialog, sbb-menu, sbb-navigation, sbb-tooltip): fix reading order for VoiceOver with Safari

### LINKED ISSUE #1879: bug(a11y): sbb-navigation can be left with screenreader without closing A4A#77
The main navigation can be exited using the screen reader without closing it.
With screen readers you can use the arrow keys to reach the whole page when the navigation is open.

### LINKED ISSUE #1880: bug(a11y): language selection does not work completely reliably A4A#78
Note: It is not entirely clear whether this is directly a problem with our `sbb-menu` or the usage of sbb.ch, but we should still check.

The language selection does not work completely reliably. If the switch is pressed again with a screen reader in the open state, then it can no longer be closed. The same problem also exists with the dropdown context menu in the connection details.

The finding occurs in FF, regardless of whether Jaws or NVDA. It cannot be reproduced on Edge, i.e. sometimes it occurs and sometimes it doesn't.

Quick start Guide
The following is a quick guide, please use FF for this:
1. Go to the top of the page, arrow down, until the language selection comes. I'm not sure what you're testing with, but I suspect NVDA. When you click on the language selection, NVDA will read out something like: "German submenu, etc.".
2. The focus is set to the chosen language, which is great. Now you press the up arrow once and you land on the extended language switch (language button to open the menu), try to close it with space or enter, that won't work. If you click enter or the spacebar once, the escape key no longer works either.

https://dev-www.sbb.ch/en

### LINKED ISSUE #2015: [Bug]: Dialog is output with incorrect role
### Preflight Checklist

### Bug type

Accessibility

Components

0.40.14

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

The "Travelers" dialog is output with the wrong role as "group" because the element is marked with Role="group".

The Role="group" should be removed from the element.
Unbenannt1

1. Step one
2. Step two
3. ...

### Relevant log output

_No response_

jeripeierSBB:
Closes #2015 
Closes #1879
Closes #1880

BREAKING CHANGE:
- The property `accessibilityLabel` of the `sbb-navigation` was removed. As replacement, an `aria-label` can now directly be placed on the `sbb-navigation` element.


kyubisation (approved):
LGTM 👍

src/components/sbb-navigation-section/sbb-navigation-section.tsx
```diff
@@ -275,23 +266,14 @@ export class SbbNavigationSection implements ComponentInterface {
 
   // Set focus on the first focusable element.
   private _setNavigationSectionFocus(): void {
```
dauriamarco, line 268: By removing the "trick" of setting the focus on the wrapper when `mostRecentModality !== 'keyboard'`, if I open the nav by click and press `Tab` I'll land on the second focusable element and not the first. This happens for both the navigation and the navigation section. Is this wanted?

src/components/sbb-navigation/sbb-navigation.tsx
```diff
@@ -287,24 +276,11 @@ export class SbbNavigation implements ComponentInterface {
 
   // Set focus on the first focusable element.
   private _setNavigationFocus(): void {
```
dauriamarco, line 278: Same here as for the section.

dauriamarco (commented):
Looks good! 🙂 Just a comment:

src/components/sbb-navigation-section/sbb-navigation-section.tsx
```diff
@@ -275,23 +266,14 @@ export class SbbNavigationSection implements ComponentInterface {
...
-      // the focus-visible styles would be incorrectly applied
-      this._navigationSectionContainerElement.tabIndex = 0;
-      this._navigationSectionContainerElement.focus();
-
-      this._navigationSectionContainerElement.addEventListener(
-        'blur',
-        () => this._navigationSectionContainerElement.removeAttribute('tabindex'),
-        { once: true },
-      );
+    const firstFocusableElement = getFirstFocusableElement(
```
dauriamarco, line 269: Another thing I noticed is that previously on mobile the back button that appears was the first focusable to be focused but now it is the first label. Is this also ok?

dauriamarco: This happens on iOS with VoiceOver.

jeripeierSBB: I think it should be the back button, but I did not change this logic? :/ We would need to consider shadow dom as well here

dauriamarco: Yes, I was thinking the same thing. Because we didn't target the shadowDom before either, but somehow the back button was still focused by VoiceOver 🤔

dauriamarco (approved):
LGTM 👍

src/components/sbb-menu/sbb-menu.tsx
```diff
@@ -421,9 +391,9 @@ export class SbbMenu implements ComponentInterface {
     return (
       <Host data-state={this._state} ref={assignId(() => this._menuId)}>
         <div class="sbb-menu__container">
-          <dialog
+          <div
             onAnimationEnd={(event: AnimationEvent) => this._onMenuAnimationEnd(event)}
-            ref={(dialogRef) => (this._dialog = dialogRef)}
+            ref={(el) => (this._menu = el)}
             class="sbb-menu"
             role="presentation"
```
jeripeierSBB, line 398: role presentation can be removed

kyubisation (approved):
LGTM 👍

dauriamarco (approved):
LGTM 👍

PR #2109, opened 2023-10-24
docs(readme): restored readme auto-generation

TomMenga:
* [x] Missing `events` table (will fix itself once we write the events doc)
* [x] Missing `dependencies` table (maybe we can live with that)


kyubisation (approved):
LGTM 👍 
Thanks for the work 😃

PR #2121, opened 2023-10-27
chore: update lit to 3.0

TomMenga:
### Known issues
- [x] `@open-wc/lit-helpers` still has a dependency on lit 2.0 which causes the "Multiple version of lit..." warning


jeripeierSBB (commented):
LGTM. can we not remove this.requestUpdate in setters now? -> https://lit.dev/blog/2023-10-10-lit-3.0/ 

> We now call requestUpdate() automatically for @property and @state decorated accessors where previously that was the setter's responsibility.

TomMenga:
> LGTM. can we not remove this.requestUpdate in setters now? -> https://lit.dev/blog/2023-10-10-lit-3.0/
> 
> > We now call requestUpdate() automatically for @Property and @State decorated accessors where previously that was the setter's responsibility.

I thought that we would need to do that once we started using standard decorators but you might have a point. I might have misinterpreted it.
We also need to move the `@property` and `@state` decorators from the getters to the setters.

I'll test if it's true that the 'request update' is called automatically now

src/components/sbb-accordion/sbb-accordion.tsx
```diff
@@ -102,6 +99,10 @@ export class SbbAccordion extends LitElement {
     this.addEventListener('will-open', (e) => this._closePanels(e), { signal });
   }
 
+  public override willUpdate(_changedProperties: Map<PropertyKey, unknown>): void {
```
jeripeierSBB, line 102: probably left unintentionally

jeripeierSBB (changes requested):
A console log left and I suggest to use the following order: @property, setter, getter. I'ts more readable than getter, @property, setter.

PR #2122, opened 2023-10-27
refactor: improve linting

kyubisation:
This PR enables `lit-analyzer` with a limited ruleset.


src/components/sbb-alert-group/sbb-alert-group.tsx
```diff
@@ -5,15 +5,15 @@ import { customElement, property, state } from 'lit/decorators.js';
 import { EventEmitter, ConnectedAbortController } from '../../global/eventing';
 import { SbbAlert } from '../sbb-alert';
 import { setAttribute } from '../../global/dom';
-import Style from './sbb-alert-group.scss?lit&inline';
+import style from './sbb-alert-group.scss?lit&inline';
 
 /**
- * @slot unnamed - content slot, should be filled with `sbb-alert` items.
+ * @slot - content slot, should be filled with `sbb-alert` items.
```
jeripeierSBB, line 11: I'm not sure whether I like this notation without unnamed but still with "-". Is this a must?

src/components/sbb-skiplink-list/readme.md
```diff
@@ -37,7 +37,7 @@ to specify a custom stack order, the `z-index` can be changed by defining the CS
 
 ## Slots
 
-| Name      | Description                             |
-| --------- | --------------------------------------- |
-| `unnamed` | Use this to provide links for the list. |
+| Name | Description                             |
+| ---- | --------------------------------------- |
+|      | Use this to provide links for the list. |
```
jeripeierSBB, line 42: a little weird in this case because of using "this" but what is "this"? (I know, but with unnamed it was more clear)

jeripeierSBB (changes requested):
Very cool. with that we found several things to fix :D. I did not check, but maybe in some stories / tests imports could be removed again (e.g. the one to the icon) as it is already included in the component itself now.

src/components/sbb-alert-group/sbb-alert-group.tsx
kyubisation: As discussed via Teams, this is the recommendation. (Which is still vague, but exists.)

jeripeierSBB (approved):
LGTM! Skiplink is not done yet, so maybe my comment should be addressed later

PR #2124, opened 2023-10-27
fix: improve stabilty of random failing tests

TomMenga:


src/components/sbb-tooltip/sbb-tooltip.tsx
```diff
@@ -350,7 +350,7 @@ export class SbbTooltip extends LitElement {
         .filter((el) => el instanceof window.HTMLElement)
         .find((el) => (el as HTMLElement).matches(IS_FOCUSABLE_QUERY)) as HTMLElement;
       clearTimeout(this._closeTimeout);
-      await this.close();
+      this.close();
```
jeripeierSBB, line 353: If you remove that, function probably doesn't need to be async at all. Please leave a comment why await should not be used here (as e.b. I tend to always fixing not awaited calls ;) )

TomMenga: oh I missed it :/
Nice catch

src/components/sbb-tooltip/sbb-tooltip.tsx
```diff
@@ -343,14 +343,14 @@ export class SbbTooltip extends LitElement {
   };
 
   // Close tooltip on backdrop click.
-  private _closeOnBackdropClick = async (event: PointerEvent): Promise<void> => {
+  private _closeOnBackdropClick = (event: PointerEvent): Promise<void> => {
```
jeripeierSBB, line 346:
```suggestion
  private _closeOnBackdropClick = (event: PointerEvent): void => {
```

src/components/sbb-tooltip/sbb-tooltip.tsx
jeripeierSBB: Can you add a comment why it should not be awaited?

TomMenga: It was needed because stencil forced the async on the `close()` method.

jeripeierSBB: Ah sorry, the close method is not async anymore

jeripeierSBB (approved):
LGTM

PR #2139, opened 2023-11-01
build: use custom test reporter

kyubisation:
This PR adds a custom test reporter that prints each passed test as a dot and each failed or skipped test as a red x.
At the end, it reports amount of test suites, tests, passed, failed and skipped amount and additionally reports errors.


jeripeierSBB (approved):
Very cool!

PR #2151, opened 2023-11-06
test: unskip tests

TomMenga:


src/components/sbb-notification/sbb-notification.e2e.ts
```diff
@@ -73,22 +72,18 @@ describe('sbb-notification', () => {
...
     const didCloseEventSpy = new EventSpy(SbbNotification.events.didClose);
     const closeButton = element.shadowRoot.querySelector('.sbb-notification__close') as SbbButton;
 
     expect(element).not.to.be.null;
     expect(element).to.have.attribute('data-state', 'opened');
 
-    closeButton.focus();
-    await sendKeys({ down: 'Enter' });
-    await waitForLitRender(element);
+    closeButton.click();
```
jeripeierSBB, line 83: I think the previous test already does the same logic? but maybe this test does not give a lot of value anyways

TomMenga: That was my mistake. Initially, it tested the "close by keyboard".
Anyway, I agree with you. The difference between keyboard and click interaction on an `sbb-button` is not the responsibility of the `sbb-notification`.
I think I'll remove it since it has a strange timing interaction problem that doesn't let the button be focused

src/components/time-input/time-input.e2e.ts
```diff
@@ -104,7 +104,8 @@ describe('sbb-time-input', () => {
     expect(input).not.to.have.attribute('data-sbb-invalid');
   });
 
-  it('should interpret valid values', async () => {
+  it('should interpret valid values', async function () {
+    this.timeout(5000);
```
jeripeierSBB, line 108: what's the difference between this.timeout and aTimeout?

web-test-runner.config.js
```diff
@@ -12,9 +12,9 @@ const globalCss = sass.compile('./src/components/core/styles/global.scss', { loa
 
 const browsers = isCIEnvironment
   ? [
-      playwrightLauncher({ product: 'chromium' }),
-      playwrightLauncher({ product: 'firefox' }),
-      playwrightLauncher({ product: 'webkit' }),
+      playwrightLauncher({ product: 'chromium', concurrency: 1 }),
```
jeripeierSBB, line 15: Can you write a comment why we need the concurrency?

jeripeierSBB (approved):

src/components/time-input/time-input.e2e.ts
TomMenga: `aTimeout` waits for time
`this.timeout(5000)` sets the maximum time that the test can last before being forced to failure ( overrides the global config we have in `web-test-runner.config` )

web-test-runner.config.js
TomMenga: I'm not really sure why it has problems but I'll write a comment saying that it's necessary to stay like that

src/components/time-input/time-input.e2e.ts
jeripeierSBB: thanks for the hint

jeripeierSBB (approved):
LGTM, thanks!

PR #2160, opened 2023-11-07
fix: connect sbb-datepicker with related components when initialized

### LINKED ISSUE #2047: [Bug]: datepicker-toggle is not initialized
### What happened?

In certain cases it can be that the datepicker toggle cannot get itself initialized and is therefore not clickable

Create a form field and add first a datepicker toggle and in a second step the sbb-datepicker. The toggle does not get initialized.

### Solution ###
We need mechanism where the datepicker sends an event on which relevant components can react and initialize whenever needed.

DavideMininni-Fincons:
Closes #2047 

Code for testing (in toggle stories):

```
const fn = () : void => {
  const picker = document.createElement('sbb-datepicker');
  const container = document.querySelector('#container');
  container.append(picker);
};

const TestTemplate = ({negative, ...args}): JSX.Element => (
  <div>
    <sbb-form-field negative={negative} id='container'>
    <input />
    {StandaloneTemplate(null, args)}
    </sbb-form-field>
    <button onClick={() => fn()}>Click</button>
  </div>
);

export const Test: StoryObj = {
  render: TestTemplate,
  argTypes: defaultArgTypes,
  args: { ...defaultArgs },
};
```

Code for testing standalone case:

```
const fn = () : void => {
  const picker = document.createElement('sbb-datepicker');
  picker.setAttribute('id', 'datepicker');
  picker.setAttribute('input', 'datepicker-input');
  const container = document.querySelector('#container');
  container.append(picker);
};

const TestTemplate = ({negative, ...args}): JSX.Element => (
  <div id='container'>
    <input  id="datepicker-input" />
    {StandaloneTemplate('datepicker', args)}
    <button onClick={() => fn()}>Click</button>
  </div>
);

export const Test: StoryObj = {
  render: TestTemplate,
  argTypes: defaultArgTypes,
  args: { ...defaultArgs },
};
```

### DIFF
--- a/src/components/datepicker/datepicker-next-day/readme.md
+++ b/src/components/datepicker/datepicker-next-day/readme.md
@@ -28,7 +28,7 @@ if it is disabled, or if the selected date is equal to the input's `max` attribu
 </sbb-form-field>
 ```
 
-Since the component needs the `sbb-datepicker` to work properly, if they are dynamically added,
+NOTE: Since the component needs the `sbb-datepicker` to work properly,
 both standalone or within the `sbb-form-field`, they must have the same parent element to be correctly connected.
 
 <!-- Auto Generated Below -->
--- a/src/components/datepicker/datepicker-next-day/readme.md
+++ b/src/components/datepicker/datepicker-next-day/readme.md
@@ -28,6 +28,9 @@ if it is disabled, or if the selected date is equal to the input's `max` attribu
 </sbb-form-field>
 ```
 
+Since the component needs the `sbb-datepicker` to work properly, if they are dynamically added,
+both standalone or within the `sbb-form-field`, they must have the same parent element to be correctly connected.
+
 <!-- Auto Generated Below -->
 
 ## Properties
--- a/src/components/datepicker/datepicker-next-day/datepicker-next-day.ts
+++ b/src/components/datepicker/datepicker-next-day/datepicker-next-day.ts
@@ -90,6 +90,11 @@ export class SbbDatepickerNextDay extends LitElement {
     this._syncUpstreamProperties();
     if (!this.datePicker) {
       this._init();
+      // If the component is attached to the DOM before the datepicker, it has to listen for the datepicker init,
+      // assuming that the two components share the same parent element.
+      if (!this._datePickerElement) {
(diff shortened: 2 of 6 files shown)


MarioCastigliano (commented):
What happens if the input is actually updated (for example changing the `min` or `max` attributes)? Do both listeners call the _init method?

DavideMininni-Fincons:
> What happens if the input is actually updated (for example changing the `min` or `max` attributes)? Do both listeners call the _init method?

The listener is attached only if in the connectedCallback the picker is not found within the '_init' method. After the first inputUpdated event (which should happen in the datepicker's connectedCallback), the listener is disconnected due to 'once: true'.
In theory, no, it should no be called twice.

src/components/datepicker/datepicker-next-day/readme.md
```diff
@@ -28,6 +28,9 @@ if it is disabled, or if the selected date is equal to the input's `max` attribu
 </sbb-form-field>
 ```
 
+Since the component needs the `sbb-datepicker` to work properly, if they are dynamically added,
+both standalone or within the `sbb-form-field`, they must have the same parent element to be correctly connected.
+
```
TomMenga, line 33: Since the 'dynamically load' thing is hard to understand and explain, is it acceptable to inform the consumer that the date picker and the accessories components need to be children of the same element?

TomMenga (changes requested):
The fix should work now but I'd also like to hear what @MarioCastigliano thinks of it

src/components/datepicker/datepicker-next-day/readme.md
DavideMininni-Fincons: You mean remove the 'dynamically added' part, and just say 'when you use them, use them as siblings'?

TomMenga: Yep, I feel like it would simplify the communication

DavideMininni-Fincons: Done!

MarioCastigliano (commented):
👍👍

MarioCastigliano (approved):

PR #2170, opened 2023-11-08
docs(readme): adapt the doc generation to the new folder structure

TomMenga:


jeripeierSBB (approved):
LGTM, thanks for the fix

jeripeierSBB:
It seems that @kyubisation did a similiar fix in https://github.com/lyne-design-system/lyne-components/pull/2171?

TomMenga:
> It seems that @kyubisation did a similiar fix in #2171?

Yeah, it looks like it, I'll wait the daily to confirm it and in case we can close this pr

PR #2171, opened 2023-11-08
feat: implement React wrapper

kyubisation:
This PR adds the React wrapper to the library. The wrapper components are created temporarily during the build and removed after it.


src/react/.gitignore
```diff
@@ -0,0 +1 @@
+*/
```
jeripeierSBB, line 1: A potential new line missing=?

jeripeierSBB (approved):
Yeah!

jeripeierSBB (approved):
Yeah!

src/react/.gitignore
kyubisation: prettier does not seems to care 😅

PR #2186, opened 2023-11-14
fix(sbb-radio-button-group, sbb-checkbox-group): radio and checkbox group state update

dauriamarco:
Fix cherry-pick for `sbb-radio-button-group` and `sbb-checkbox-group`.


jeripeierSBB (approved):
LGTM. thanks

PR #2189, opened 2023-11-15
fix(sbb-selection-panel): disabled border color when checked

dauriamarco:
Fix border color when the selection panel has no content, is checked and is disabled.


jeripeierSBB (approved):
LGTM, thanks!

PR #2197, opened 2023-11-16
fix(sbb-icon): change sanitize prop to no-sanitize

dauriamarco:
Invert property default behaviour from `sanitize = true` to `noSanitize = false`.


jeripeierSBB (approved):
LGTM! Please don't forget to add breaking change message when merging.

PR #2198, opened 2023-11-16
refactor(stories): migrate jsx to lit template

TomMenga:


package.json
```diff
@@ -37,7 +37,7 @@
     "lint:js": "eslint \"**/*.js\"",
     "lint:yml": "eslint \"**/*.{yml,yaml}\"",
     "lint:styles": "stylelint \"**/*.scss\"",
-    "lint:lit": "yarn lit-analyzer \"src/components/**/*.ts\"",
```
jeripeierSBB, line 40: could we remove the react deps now?

jeripeierSBB (commented):
Does `yarn generate:chromatic-stories` still work?

package.json
jeripeierSBB: and jsx deps

TomMenga: I think yes.
We should also convert files from tsx to ts

kyubisation (approved):
LGTM 👍

PR #2208, opened 2023-11-21
fix(sbb-logo,sbb-signet): solve SSR issue with text assignment

kyubisation:
Previous attempts to fix SSR failed due to various reasons. Now we just use property assignment with `.textContent`, which forces any text given to be interpreted as text. This avoid exploitation and simplifies the code.


jeripeierSBB (approved):
Sometimes it's more simple :D

PR #2216, opened 2023-11-22
build: minify CSS in build bundle

kyubisation:
This PR configures Vite to minify CSS in our build bundle. This should reduce the payload both for our package and our consumers.


jeripeierSBB (approved):
LGTM

PR #2218, opened 2023-11-23
refactor: change EventEmitter.emit to return the result of dispatchEvent

kyubisation:
This change allows us to check if an event was prevented. Additionally this PR enables SSR usage.


jeripeierSBB (approved):
LGTM

PR #2219, opened 2023-11-23
refactor: update coding standards

### LINKED ISSUE #2210: update coding standards to lit

dauriamarco:
Closes #2210

### DIFF
--- a/CODING_STANDARDS.md
+++ b/CODING_STANDARDS.md
@@ -234,7 +234,7 @@ disabled: boolean = false;
 
 #### Properties initialization
 
-Boolean properties with a default value can be initialized with `false` but not with `true`. This is due to how Lit handles boolean attributes in the DOM: it evaluates, for example, `sanitize="false"` as `true` and not `false` as we would expect. Therefore the property shold be converted in something like `noSanitize = false`:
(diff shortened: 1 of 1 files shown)


kyubisation (approved):
LGTM 👍

CODING_STANDARDS.md
```diff
@@ -224,6 +232,28 @@ Boolean properties and return values should use "Whether..." as opposed to "True
 disabled: boolean = false;
 ```
 
+#### Properties initialization
+
+Boolean properties with a default value can be initialized with `false` but not with `true`. This is due to how Lit handles boolean attributes in the DOM: it evaluates, for example, `sanitize="false"` as `true` and not `false` as we would expect. Therefore the property shold be converted in something like `noSanitize = false`:
```
dauriamarco, line 237:
```suggestion
Boolean properties with a default value can be initialized with `false` but not with `true`. This is due to how Lit handles boolean attributes in the DOM: it evaluates, for example, `sanitize="false"` as `true` and not `false` as we would expect. Therefore the property should be converted in something like `noSanitize = false`:
```

jeripeierSBB (approved):
LGTM

jeripeierSBB (approved):
LGTM

PR #2231, opened 2023-11-24
refactor(sbb-dialog): introduce inner components and header animation

### LINKED ISSUE #2019: [Bug]: SBB-Dialog Content is not scrollable / accessible
### Preflight Checklist

### Bug type

Accessibility

Components

0.40.14

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

With the high font magnification (400%), the pages are not scrollable, so the application cannot be operated.
zoom

1. Step one
2. Step two
3. ...

### Relevant log output

_No response_

dauriamarco:
Closes #2019 

BREAKING CHANGE: The `sbb-dialog` component now supports the dedicated inner elements `sbb-dialog-title`, `sbb-dialog-content`, and `sbb-dialog-actions`. Use these components to respectively provide a title, a content and, optionally, a footer with an action group. Moreover, the full-screen variant (which occurred when no title was provided to the dialog) has been removed. To achieve a full-screen overlay, please use the `sbb-overlay` component.
This was the previous implementation:
```html
<sbb-dialog title-content="Title">
  <p>Dialog content.</p>
  <sbb-action-group slot="action-group">...</sbb-action-group>
</sbb-dialog>
```
This is the new implementation:
```html
<sbb-dialog>
  <sbb-dialog-title>Title</sbb-dialog-title>
  <sbb-dialog-content>Dialog content</sbb-dialog-content>
  <sbb-dialog-actions>...</sbb-dialog-actions>
</sbb-dialog>
```
Previously, a full-screen dialog was displayed if no title was provided to the dialog component:
```html
<sbb-dialog>
  <p>Dialog content.</p>
</sbb-dialog>
```
It is now possible to use an `sbb-overlay` component to display the same variant:
```html
<sbb-overlay>
  <p>Overlay content.</p>
</sbb-overlay>
```


jeripeierSBB (changes requested):
Thanks for addressing this!
Code looks ok so far, but testing it in real life, the initial scrolling looks buggy for me as it forces the user to scroll down at least the height of the header. Is this really intended?

Please also wait for approval of @mcilurzo

dauriamarco:
> Thanks for addressing this! Code looks ok so far, but testing it in real life, the initial scrolling looks buggy for me as it forces the user to scroll down at least the height of the header. Is this really intended?
> 
> Please also wait for approval of @mcilurzo

@jeripeierSBB Yes, I know what you mean, but I have already shown it to Manuel and it is fine with him. But if you want we can show it to him again and discuss it, although I can't think of any other possible alternatives at the moment.

jeripeierSBB:
We would like to discuss with @mcilurzo if there is the need of a property to opt out of the header hiding feature. potentially providing a breakpoint range from where it stops or starts... to be discussed.

kyubisation:
We discussed this with @mcilurzo.
We would like to have additional elements `sbb-dialog-title` (extends from `sbb-title`), `sbb-dialog-content` and `sbb-dialog-actions` (extends from `sbb-action-group`).
`sbb-dialog-title` should have a property `hideOnScroll: false | '' | 'zero' | 'micro' | ... = false;`, which indicates up to (including) which breakpoint to title should be hidden when scrolled (`''` loosely equals to 'ultra', in order for `<sbb-dialog-title hide-on-scroll>` to work).

src/components/dialog/dialog-content/readme.md
```diff
@@ -0,0 +1,15 @@
+Use the `sbb-dialog-content` in combination with the [sbb-dialog](/docs/components-sbb-dialog--docs) to display a content inside the dialog.
+
+```html
+<sbb-dialog-content>Dialog content.</sbb-dialog-content>
+```
```
DavideMininni-Fincons, line 5: Maybe add an example with the dialog?
```
<sbb-dialog params>
  <sbb-dialog-content>Dialog content</sbb-dialog-content>
</sbb-dialog>
```

src/components/dialog/dialog-title/readme.md
```diff
@@ -0,0 +1,46 @@
+The `sbb-dialog-title` component extends the [sbb-title](/docs/components-sbb-title--docs) component. Use it in combination with the [sbb-dialog](/docs/components-sbb-dialog--docs) to display a header in the dialog with a title, a close button and an optional back button.
+
+```html
+<sbb-dialog-title
```
DavideMininni-Fincons, line 4: See other similar comment to wrap it in a sbb-dialog tag

src/components/dialog/dialog-title/readme.md
```diff
@@ -0,0 +1,46 @@
...
+  accessibility-back-label="Back button"
+  >A describing title of the dialog</sbb-dialog-title
+>
+```
+
+It has an implicit slot name: `title`.
+
+## States
+
+The title can have a `negative` state which is automatically synchronised with the negative state of the dialog. In addition, the title can be hidden when scrolling down the content, to provide more space for reading the content itself; this can be done thanks to the `hide-on-scroll` property, which can determine whether to hide the title and up to which breakpoint.
```
DavideMininni-Fincons, line 17: Add examples (see title's readme as comparison)

src/components/dialog/dialog/readme.md
```diff
@@ -0,0 +1,109 @@
...
+emit a close event with an optional result as a payload.
+
+The component can also be dismissed by clicking on the close button, clicking on the backdrop, pressing the `Esc` key,
+or, if an element within the `sbb-dialog` has the `sbb-dialog-close` attribute, by clicking on it.
+
+You can also set the property `titleBackButton` on the `sbb-dialog-title` component to display the back button in the title section which will emit the event `requestBackAction` when clicked.
+
+## Style
+
+It's possible to display the component in `negative` variant using the self-named property.
```
DavideMininni-Fincons, line 71: Add an example here

src/components/dialog/dialog-actions/readme.md
```diff
@@ -0,0 +1,23 @@
+The `sbb-dialog-actions` component extends the [sbb-action-group](/docs/components-sbb-action-group--docs) component. Use it in combination with the [sbb-dialog](/docs/components-sbb-dialog--docs) to display a footer with an action group.
+
+```html
+<sbb-dialog-action>
```
DavideMininni-Fincons, line 4: See other comments about wrapping

DavideMininni-Fincons (changes requested):
Great work. I like the class extension (from title and action-group).

src/components/dialog/dialog/dialog.ts
```diff
@@ -215,8 +184,42 @@ export class SbbDialogElement extends LitElement {
...
+      // Scrolling up
+      this._setHideHeaderDataAttribute(false);
+    }
+    // `currentScroll` can be negative, e.g. on mobile; this is not allowed.
+    this._lastScroll = currentScroll <= 0 ? 0 : currentScroll;
+  }
+
   public override connectedCallback(): void {
     super.connectedCallback();
+    new NamedSlotStateController(this);
```
TomMenga, line 221: Not sure we need this

src/components/dialog/dialog/dialog.ts
```diff
@@ -133,49 +99,52 @@ export class SbbDialogElement extends LitElement {
...
       return;
     }
     this._lastFocusedElement = document.activeElement as HTMLElement;
 
+    // Initialize dialog elements
+    this._dialogTitleElement = this.querySelector('sbb-dialog-title')!;
+    this._dialogContentElement = this.querySelector('sbb-dialog-content')!.shadowRoot!
+      .firstElementChild as HTMLElement;
+    this._dialogActionsElement = this.querySelector('sbb-dialog-actions') || undefined;
+    this._ariaLiveRef = this.shadowRoot!.querySelector('.sbb-screen-reader-only')!;
```
TomMenga, line 136: While I understand the other inits, why do we need to fetch this element each open?

TomMenga (changes requested):
Good job! 👍 
Some minor discussion points and fixes

src/components/dialog/dialog-title/dialog-title.ts
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Visual level for the title. Optional, if not set, the value of level will be used.
+   */
+  @property({ attribute: 'visual-level', reflect: true }) public override visualLevel?: TitleLevel =
+    '3';
+
+  /**
+   * Whether a back button is displayed next to the title.
+   */
+  @property({ attribute: 'title-back-button', type: Boolean }) public titleBackButton = false;
```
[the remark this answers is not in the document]
jeripeierSBB: As the whole change is a breaking change, I think we could just rename it. A quick lookup in internal code database showed me that it anyways isn't used at all.

src/components/dialog/dialog/dialog.stories.ts
```diff
@@ -0,0 +1,483 @@
...
+  table: {
+    category: 'Title',
+  },
+};
+
+const hideOnScroll: InputType = {
+  control: {
+    type: 'select',
+  },
+  options: ['zero', 'micro', 'small', 'medium', 'large', 'wide', 'ultra'],
```
TomMenga, line 73:
```suggestion
  options: ['off', 'zero', 'micro', 'small', 'medium', 'large', 'wide', 'ultra'],
```
Should we add an 'off' value?
And in the story title template something like `hide-on-scroll=${hideOnScroll !== 'off' ? hideOnScroll : nothing}`

TomMenga (changes requested):
All good, just a nitpick in stories

src/components/dialog/dialog/dialog.stories.ts
dauriamarco: I see your point but I'm afraid it'd be confusing for the consumer who might think that `off` would be a possible value for the property.

TomMenga: true... can't see a good solution that makes everyone happy :/

TomMenga (approved):

src/components/dialog/dialog/dialog.ts
```diff
@@ -363,101 +402,61 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
 
   // Set focus on the first focusable element.
   private _setDialogFocus(): void {
-    const firstFocusable = this.shadowRoot!.querySelector(IS_FOCUSABLE_QUERY) as HTMLElement;
+    const firstFocusable = getFirstFocusableElement(
+      Array.from(this.children).filter((e): e is HTMLElement => e instanceof window.HTMLElement),
```
jeripeierSBB, line 406: Don't we exclude close button with that?

src/components/dialog/dialog/dialog.ts
```diff
@@ -348,7 +387,7 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
     // Take accessibility label or current string in title section
     const label =
       this.accessibilityLabel ||
-      (this.shadowRoot!.querySelector('.sbb-dialog__title') as HTMLElement)?.innerText.trim();
+      (this.querySelector('sbb-dialog-title') as HTMLElement)!.innerText.trim();
```
jeripeierSBB, line 390: we would have a field for sbbdialogtitle

jeripeierSBB (changes requested):
Thanks for the huge refactoring!
Some points to do and to discuss are left. The coupling of all the component is very tight (css var / data-attributes and logic), but probably it's okay because they are only meant to be used together.

- Do we need to include testA11yTreeSnapshot() in all specs? @TomMenga 
- [ ] Check HCM
-  [ ] Mention the breaking changes in PR description (including removed fullscreen support)

TomMenga:
> * Do we need to include testA11yTreeSnapshot() in all specs? @TomMenga

As a general rule, yes. At least one `testA11yTreeSnapshot` of the most meaningful scenario for each component.

For @dauriamarco, we have 3 possible ways to implement them:
- When the template is taken care of by a beforeEach (see `accordion.spec.ts`)
- When the template is elementary (see `clock.spec.ts`)
- Otherwise, we can convert a spec test to a describe with a beforeEach (see `expansion-panel.spec.ts`)

If you have any doubts let me know 😸

src/components/dialog/dialog/dialog.ts
dauriamarco: In this case not anymore, as we are now slotting the close button with the `sbb-dialog-title`.

src/components/dialog/dialog-title/dialog-title.scss
```diff
@@ -0,0 +1,58 @@
...
+    transition: {
+      property: box-shadow, transform;
+      duration: var(--sbb-dialog-animation-duration);
+      timing-function: var(--sbb-dialog-animation-easing);
+    }
+  }
+
+  :host([data-overflows][data-has-visible-focus-within]) &,
+  :host([data-overflows]:not([negative], [data-hide-header])) & {
+    @include sbb.shadow-level-9-soft;
```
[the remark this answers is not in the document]
dauriamarco: I asked Manuel and he confirmed no shadow on negative.

src/components/dialog/dialog/dialog.ts
```diff
@@ -233,10 +234,32 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
     }
   }
 
+  protected override firstUpdated(): void {
+    this._ariaLiveRef = this.shadowRoot!.querySelector('.sbb-screen-reader-only')!;
+    this._dialogTitleElement = this.querySelector('sbb-dialog-title')!;
```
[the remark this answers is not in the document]
dauriamarco: I added this because we would see a flash on the first open if it's negative, so I had to sync it earlier.

src/components/dialog/dialog/dialog.ts
```diff
@@ -233,10 +234,32 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
...
+    this._syncNegative();
```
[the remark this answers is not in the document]
dauriamarco: I added this because we would see a flash on the first open if it's negative, so I had to sync it earlier.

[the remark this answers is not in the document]
jeripeierSBB: ok. can you add a comment there why this is there?

src/components/dialog/dialog/dialog.ts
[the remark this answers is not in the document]
dauriamarco: I need this to sync the negative state before the first open. I could remove it from `open()`, even though the title might change on the next opens and keeping it can be more stable (?) You tell me 👍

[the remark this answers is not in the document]
jeripeierSBB: just keep it and add a comment :)

src/components/dialog/dialog/dialog.ts
jeripeierSBB: Ah I see because the focusable function goes into the shadow DOM of sub components.

src/components/dialog/dialog/dialog.ts
```diff
@@ -132,49 +99,51 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
...
    * Opens the dialog element.
    */
   public open(): void {
-    if (this._state !== 'closed' || !this._dialog) {
+    if (this._state !== 'closed') {
       return;
     }
     this._lastFocusedElement = document.activeElement as HTMLElement;
 
+    // Initialize dialog elements
```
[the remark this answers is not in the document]
dauriamarco: It breaks only if the users don't provide a title or a content. Those are mandatory and that's why I used the "!", I always expect to have them both. The only optional element is the `sbb-dialog-actions`. If the user provides something else it will be simply not be shown.

src/components/dialog/dialog/dialog.ts
```diff
@@ -132,49 +99,51 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
...
+    this._dialogTitleElement = this.querySelector('sbb-dialog-title')!;
+    this._dialogContentElement = this.querySelector('sbb-dialog-content')!.shadowRoot!
```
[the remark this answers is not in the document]
dauriamarco: Not sure, if I use the host all the overflow and scrolling logic would break.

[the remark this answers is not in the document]
jeripeierSBB: do we have any advantage of using a separate component for the content?

src/components/dialog/dialog/dialog.ts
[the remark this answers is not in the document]
jeripeierSBB: Ok, but open is never called there, sorry.

src/components/dialog/dialog/dialog.ts
```diff
@@ -132,49 +99,51 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
...
+      .firstElementChild as HTMLElement;
+    this._dialogActionsElement = this.querySelector('sbb-dialog-actions') || undefined;
```
jeripeierSBB, line 135: Should not better this._dialogActionsElement be able to deal with null?
```suggestion
    this._dialogActionsElement = this.querySelector('sbb-dialog-actions');
```

src/components/dialog/dialog-title/readme.md
```diff
@@ -0,0 +1,64 @@
+The `sbb-dialog-title` component extends the [sbb-title](/docs/components-sbb-title--docs) component. Use it in combination with the [sbb-dialog](/docs/components-sbb-dialog--docs) to display a header in the dialog with a title, a close button and an optional back button.
+
+```html
+<sbb-dialog>
+  <sbb-dialog-title accessibility-close-label="Close button" accessibility-back-label="Back button">
```
jeripeierSBB, line 5: in the case you provide an accessibility-back-label string you should also display backButton ?

mcilurzo:
looks good to me!

src/components/dialog/dialog-title/readme.md
dauriamarco: Not sure, maybe a question for @kyubisation?

kyubisation: Good question. Might be an implicit opt-in.
Would make sense to me, but we should mention this in the `backButton` documentation.

src/components/dialog/dialog/readme.md
```diff
@@ -0,0 +1,131 @@
...
+- disables scrolling of the page content while open;
+- manages focus properly by setting it on the first focusable element;
+- can host a [sbb-dialog-actions](/docs/components-sbb-dialog-actions--docs) component in the footer;
+- has a close button, which is always visible;
+- can display a back button next to the title;
+- adds the appropriate ARIA roles automatically.
+
+```html
+<sbb-dialog>
+  <sbb-dialog-title slot="title">Title</sbb-dialog-title>
```
jeripeierSBB, line 14: I'd never explicitely write the slot in the examples as it is assigned automatically

src/storybook/pages/home/home--logged-in.stories.ts
```diff
@@ -189,24 +189,29 @@ const Template = (args: Args): TemplateResult => html`
...
+                  (document.getElementById('my-stacked-dialog') as SbbDialogElement).open()}
+              >
+                Open stacked dialog
+              </sbb-button>
+            </sbb-dialog-content>
 
-            <sbb-action-group
-              slot="action-group"
+            <sbb-dialog-actions
+              slot="actions"
```
jeripeierSBB, line 214: Also in the examples I'd avoid to set the slot explicitly
```suggestion
```

jeripeierSBB (changes requested):

jeripeierSBB (approved):
LGTM, thanks!!

PR #2233, opened 2023-11-27
refactor(sbb-notification): set fixed story height to work around chromatic issues

jeripeierSBB:


TomMenga (approved):
LGTM 👍

dauriamarco (approved):
LGTM 👍

src/components/notification/notification.stories.ts
```diff
@@ -191,9 +181,21 @@ export const SlottedTitle: StoryObj = {
   args: { ...basicArgs, 'title-content': undefined },
 };
 
+// We set the height of the div in Chromatic to avoid cropped snapshots
 const meta: Meta = {
   decorators: [
-    (story) => html`<div style="padding: 2rem;">${story()}</div>`,
+    (story, context) =>
+      html`<div
+        style="padding: 2rem;display: flex;gap: var(--sbb-spacing-fixed-4x);flex-direction: column;${isChromatic()
```
dauriamarco, line 189: Just a comment: I noticed that now the expansion animation is a bit laggy, is it because of the `flex` here?

src/components/notification/notification.stories.ts
```diff
@@ -191,9 +181,21 @@ export const SlottedTitle: StoryObj = {
...
+          ? 'min-height: 1000px;'
+          : ''}"
+      >
+        ${trigger(context.args)}
+        <div class="notification-container" style="display: flex; flex-direction: column;">
```
dauriamarco, line 194: Just a comment: I noticed that now the expansion animation is a bit laggy, is it because of the `flex` here?

dauriamarco: Or the one on the parent.

jeripeierSBB: Could be, I'll check it. But on the other side it would be a bad sign, as consumers probably will as well often use flex box

jeripeierSBB: Animation problems seem not to be related with this change. I created a separate bug report.

PR #2235, opened 2023-11-28
fix: ssr fixes

TomMenga:


src/components/form-field/form-field/form-field.ts
```diff
@@ -154,7 +154,7 @@ export class SbbFormField extends LitElement {
   }
 
   protected override willUpdate(changedProperties: PropertyValues<this>): void {
-    if (changedProperties.has('label')) {
+    if (changedProperties.has('label') && isBrowser()) {
```
jeripeierSBB, line 157: Maybe it would be more safe to move the check into _renderLabel, in case renderlabel would be used at another place?

src/components/toggle/toggle/toggle.ts
```diff
@@ -76,8 +76,8 @@ export class SbbToggle extends LitElement {
     const options = this._options;
     const selectedOption =
       options.find((o) => o.value === value) ?? options.find((o) => o.checked) ?? options[0];
-    if (!selectedOption && isBrowser()) {
-      console.warn(`sbb-toggle: No available options! (${this.id || 'No id'})`);
+    if (!selectedOption) {
+      isBrowser() && console.warn(`sbb-toggle: No available options! (${this.id || 'No id'})`);
```
jeripeierSBB, line 80: Nice catch :)

jeripeierSBB (approved):
LGTM, one possible change

src/components/form-field/form-field/form-field.ts
```diff
@@ -170,9 +170,13 @@ export class SbbFormField extends LitElement {
...
 
   private _renderLabel(newValue: string): void {
-    let labelElement: HTMLLabelElement | undefined = Array.from(this.children)?.find(
+    if (!isBrowser()) {
+      return;
+    }
+    let labelElement = Array.from(this.children).find(
       (element) => element.tagName === 'LABEL',
-    ) as HTMLLabelElement | undefined;
+    ) as HTMLLabelElement;
```
jeripeierSBB, line 178: it can still be undefined?

TomMenga: Yes, but VsCode linting marked it as superfluous (I agree with him).
But if you'd prefer I can revert it

TomMenga: I've changed my mind, I think you're right in this case 😅 .
My thought process was that `undefined | null` is generally included in the type definition so there is no need to explicit it. But, in some cases like the return type of functions, you want to pass stronger information to the consumer by explicitly saying that `undefined` has a specific meaning.

A case where I would omit the ` | undefined ` is the type declaration of a variable
```ts
let labelElement: HTMLLabelElement = list.find( ... );

instead of 

let labelElement: HTMLLabelElement | undefined = list.find( ... );
```

But anyway, I feel like code style has an influence here. I don't think it's an absolute rule

jeripeierSBB: I just thought if you are casting with "as" you loose the undefined / null option, but maybe I was wrong and also not considering that "let" was used here

TomMenga: Anyway, I've reverted the change 😅

PR #2236, opened 2023-11-28
fix(sbb-selection-panel): fix data-state on selection panels with no content

dauriamarco:
Fix `data-state` on selection panels with no content.


jeripeierSBB (approved):
LGTM

PR #2237, opened 2023-11-28
fix(sbb-header): shows the header if it has visible focus within

dauriamarco:


src/components/core/a11y/focus.ts
```diff
@@ -16,8 +19,11 @@ export const IS_FOCUSABLE_QUERY = [
 // Note: the use of this function for more complex scenarios (with many nested elements) may be expensive.
 export function getFocusableElements(
   elements: HTMLElement[],
-  filterFunc?: (el: HTMLElement) => boolean,
-  findFirstFocusable?: boolean,
+  properties?: {
+    filterFunc?: (el: HTMLElement) => boolean;
+    findFirstFocusable?: boolean;
+    includeNotVisibleElements?: boolean;
```
jeripeierSBB, line 25: Does maybe read better?
```suggestion
    includeInvisibleElements?: boolean;
```

src/components/core/a11y/focus.ts
```diff
@@ -79,7 +91,7 @@ export class FocusTrap {
         const elementChildren: HTMLElement[] = Array.from(
           element.shadowRoot.children,
         ) as HTMLElement[];
-        const focusableElements = getFocusableElements(elementChildren, filterFunc);
+        const focusableElements = getFocusableElements(elementChildren, { filterFunc: filterFunc });
```
jeripeierSBB, line 94: nit pick, could probably be shortened
```suggestion
        const focusableElements = getFocusableElements(elementChildren, { filterFunc });
```

src/components/core/a11y/focus.ts
```diff
@@ -99,6 +111,32 @@ export class FocusTrap {
     );
   }
 
+  public setDataHasVisibleFocusWithin(element: HTMLElement): void {
```
jeripeierSBB, line 114: not sure if we on the method name should indicate that this will be constantly tracked, not only once?

src/components/header/header/header.ts
```diff
@@ -1,6 +1,7 @@
 import { CSSResultGroup, html, LitElement, TemplateResult } from 'lit';
```
jeripeierSBB, line 1: I'd appreciate some e2e tests, testing the new scenarios

src/components/header/header/header.ts
```diff
@@ -170,11 +176,16 @@ export class SbbHeaderElement extends LitElement {
     }
   }
 
+  private _addFocusableElementsListeners(): void {
+    this._focusHandler.disconnect();
```
jeripeierSBB, line 180: I know it's correct but a little bit unreadable that in disconnect, it already sets a new abort controller. It feels more like now it's lost :P

src/components/header/header/header.ts
```diff
@@ -170,11 +176,16 @@ export class SbbHeaderElement extends LitElement {
...
+    this._focusHandler.setDataHasVisibleFocusWithin(this);
+  }
+
   protected override render(): TemplateResult {
     return html`
       <header class="sbb-header">
         <div class="sbb-header__wrapper">
-          <slot></slot>
+          <slot @slotchange=${() => this._addFocusableElementsListeners()}></slot>
```
jeripeierSBB, line 188: Could you use SlotChildObserver in this case?

jeripeierSBB (changes requested):
Thanks for addressing this issue! 
I have one general thought... as Firefox is now released with css :has() selector should we not go with css solution (if it works)? I know we are not supporting the last two versions, but as version 122 will arrive soon, maybe we could accept this.
(:has(:focus-visible))

src/components/core/a11y/focus.ts
dauriamarco: I'm not sure either, do you have any suggestion? `setDataHasVisibleFocusWithinListener`?

dauriamarco:
> Thanks for addressing this issue! I have one general thought... as Firefox is now released with css :has() selector should we not go with css solution (if it works)? I know we are not supporting the last two versions, but as version 122 will arrive soon, maybe we could accept this. (:has(:focus-visible))

I tried but apparently it doesn't work with slotted elements for some reason 😞

jeripeierSBB:
> > Thanks for addressing this issue! I have one general thought... as Firefox is now released with css :has() selector should we not go with css solution (if it works)? I know we are not supporting the last two versions, but as version 122 will arrive soon, maybe we could accept this. (:has(:focus-visible))
> 
> I tried but apparently it doesn't work with slotted elements for some reason 😞

I now also tried around and it seems that it's not supported using combined with :host() and it, as you said, does never consider slotted elements. So yes, very sadly, it's not usable in our case.

src/components/core/a11y/focus.ts
```diff
@@ -99,6 +111,32 @@ export class FocusTrap {
     );
   }
 
+  public setDataHasVisibleFocusWithinListener(element: HTMLElement): void {
```
jeripeierSBB, line 114: naming topic: maybe something like "trackFocusVisibleWithin"?

jeripeierSBB (changes requested):

src/components/core/a11y/focus.ts
```diff
@@ -111,30 +111,26 @@ export class FocusHandler {
...
-      );
-      el.addEventListener(
-        'blur',
-        () => {
-          toggleDatasetEntry(element, 'hasVisibleFocusWithin', false);
-        },
-        { signal: this._controller.signal },
-      );
-    });
+  // Determine whether the element has a visible focus within.
```
jeripeierSBB, line 114: Nice! I think we could now just move it into an own file an create a ReactiveController, so we could just use it in the components where needed, without the need to connect or disconnect it manually. What do you think?
I feel it doesn't belong to the focushandler as it doesn't use any field of it. BTW: FocusHandler could be a ReactiveController as well :)

src/components/core/a11y/focus.ts
```diff
@@ -60,11 +66,14 @@ export function getFirstFocusableElement(
...
-  const focusableElements = getFocusableElements(elements, filterFunc, true);
+  const focusableElements = getFocusableElements(elements, {
+    filterFunc: filterFunc,
+    findFirstFocusable: true,
+  });
   return focusableElements.length ? focusableElements[0] : null;
 }
 
-export class FocusTrap {
+export class FocusHandler {
```
jeripeierSBB, line 76: Does this renaming still make sense?

jeripeierSBB (approved):
LGTM, thanks for the fix

src/components/header/header/header.ts
```diff
@@ -158,6 +163,9 @@ export class SbbHeaderElement extends LitElement {
   }
 
   private _closeOpenOverlays(): void {
+    if (this.hasAttribute('data-has-visible-focus-within')) {
```
jeripeierSBB, line 166: I think this should be part of the new controller, so we don't have to expose the name of the data attribute

src/components/core/a11y/focus.ts
```diff
@@ -16,8 +16,11 @@ export const IS_FOCUSABLE_QUERY = [
 // Note: the use of this function for more complex scenarios (with many nested elements) may be expensive.
 export function getFocusableElements(
   elements: HTMLElement[],
-  filterFunc?: (el: HTMLElement) => boolean,
-  findFirstFocusable?: boolean,
+  properties?: {
```
jeripeierSBB, line 19: the whole refactoring would not be needed anymore?

src/components/header/header/header.ts
dauriamarco: I'll try to implement this improvement in the dialog header PR since I need the same logic there too 👍

src/components/core/a11y/focus.ts
dauriamarco: True but I kinda like it more like this anyway, what do you think?

jeripeierSBB: me too, but it's just an option (includeInvisibleElements?: boolean;) we never use, so kind of unnecessary code.

src/components/header/header/header.ts
jeripeierSBB: perfect, thanks!

PR #2241, opened 2023-11-29
fix(sbb-select): fix template mismatch on hydration

TomMenga:


jeripeierSBB (approved):
Thanks! LGTM

src/components/form-field/form-field/form-field.ts
```diff
@@ -327,7 +327,7 @@ export class SbbFormField extends LitElement {
         signal: this._inputAbortController.signal,
       });
 
-      inputFocusElement = this.querySelector('.sbb-select-invisible-trigger');
+      inputFocusElement = (this._input as SbbSelect).inputElement;
```
jeripeierSBB, line 330: Is this really working? Seems to change the real element?

jeripeierSBB (approved):
LGTM, one question

src/components/form-field/form-field/form-field.ts
TomMenga: Not sure what you mean 🤔 
There are 2 reasons why this was necessary:
1. At this point, the trigger element is not in the light dom yet (furthermore we'd have to assume that the select has not finished initializing, which might not be always true)
2. It is not the responsibility of the form-field to know how to query for the select input element

If you still have doubts we can deep down on it

TomMenga: 👍  nice catch

PR #2242, opened 2023-11-29
fix: convert classes on host in data attributes

dauriamarco:
Avoid setting the class attribute on the host `setAttribute(this, 'class'...)` and prefer data attributes in order to prevent conflicts with pre-existing classes or classes that need to be toggled.

### DIFF
--- a/src/components/tabs/tab-group/tab-group.ts
+++ b/src/components/tabs/tab-group/tab-group.ts
@@ -361,8 +361,6 @@ export class SbbTabGroup extends LitElement {
   }
 
   protected override render(): TemplateResult {
-    console.log(this._isNested);
-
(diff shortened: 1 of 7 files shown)


src/components/tabs/tab-group/tab-group.ts
```diff
@@ -361,7 +361,9 @@ export class SbbTabGroup extends LitElement {
   }
 
   protected override render(): TemplateResult {
-    setAttribute(this, 'class', this._isNested ? 'tab-group--nested' : '');
+    console.log(this._isNested);
```
jeripeierSBB, line 364:
```suggestion
```

jeripeierSBB (approved):
LGTM, please remove console log

PR #2243, opened 2023-11-29
fix: move ref to the back of declarations

jeripeierSBB:
With lit SSR, blocks in attributes that do not generate attributes seem to cause faulty rendering. E.g. skipping attributes or assigning wrong values.


dauriamarco (approved):
LGTM 👍

PR #2244, opened 2023-11-29
fix(sbb-time-input): create get/set for valueAsDate

### LINKED ISSUE #2223: refactor: turn getValueAsDate/setValueAsDate into a get/set property
Convert getValueAsDate/setValueAsDate into a get/set property valueAsDate (sbb-time-input and sbb-datepicker). 
Proposal: use a couple of get/set properties (value, valueAsDate) just like on native input elements in order to avoid ambiguous scenarios like `timeInput.valueAsDate = “2024-10-10T09:00:00.000Z”`.

DavideMininni-Fincons:
Closes #2223

BREAKING CHANGE: `getValueAsDate()` and `setValueAsDate()` methods have been replaced by getter/setter `valueAsDate`.


DavideMininni-Fincons:
Hello @kyubisation 
about this, I noticed that the `valueAsDate` property is added into the readme file (the related attribute is wrong and the fixing the generation should be handled in the story #2246 ).
Is this ok for you? Should I add the `@property` annotation on the setter?

TomMenga (approved):
Good job 👍 
Don'r forget to mark the breaking change

PR #2256, opened 2023-12-04
fix(select): enhance test robustness

TomMenga:
Fix a flickey test on `sbb-select` caused by the recent ssr fixes


jeripeierSBB (approved):
LGTM

PR #2258, opened 2023-12-04
fix: replace attribute access with property where possible

jeripeierSBB:


DavideMininni-Fincons (approved):
LGTM, thanks!

TomMenga (approved):
LGMT too 👍

PR #2259, opened 2023-12-04
fix(sbb-toggle): calc pill position on hydration

jeripeierSBB:


src/components/toggle/toggle/toggle.ts
```diff
@@ -232,7 +232,13 @@ export class SbbToggle extends LitElement {
   protected override render(): TemplateResult {
     setAttribute(this, 'role', 'radiogroup');
     return html`
-      <div class="sbb-toggle" ${ref((toggle) => (this._toggleElement = toggle as HTMLElement))}>
+      <div
+        class="sbb-toggle"
+        ${ref((toggle) => {
+          this._toggleElement = toggle as HTMLElement;
+          this._setCheckedPillPosition(false);
+        })}
```
TomMenga, line 240: Due to how the `ref` directive is implemented, the callback will be called twice for each component render.
I think the best solution here is to call `this._setCheckedPillPosition(false);` inside the `firstUpdated` callback

PR #2263, opened 2023-12-05
test: a11y tree snapshot showcase

TomMenga:
[Docs](https://modern-web.dev/docs/test-runner/commands/#accessibility-snapshot)

Unfortunately, I feel that the snapshot doesn't add much value to tests 😞 
Furthermore, Playwright marks the snapshot functionality as [deprecated](https://playwright.dev/docs/api/class-accessibility#deprecated) and advises using other tools like [Axe](https://www.deque.com/axe/) (which I checked, but is mainly helpful when testing full pages). 

On a positive note, I found something that might be useful during development on a11y matter.
Chrome dev tool offers an a11y tree inspector that's what screen readers use. 
It might be a good starting point to compare the trees when we have some screen reader misbehavior

image

### DIFF
--- a/src/components/autocomplete/autocomplete.e2e.ts
+++ b/src/components/autocomplete/autocomplete.e2e.ts
@@ -1,5 +1,10 @@
 import { assert, expect, fixture } from '@open-wc/testing';
-import { sendKeys, sendMouse } from '@web/test-runner-commands';
+import {
+  a11ySnapshot,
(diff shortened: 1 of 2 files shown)


TomMenga:
We decided not to go further with the implementation because playwright has marked the [API as deprecated](https://playwright.dev/docs/api/class-accessibility#accessibility-snapshot).

PR #2268, opened 2023-12-06
fix(sbb-autocomplete): delay origin look up

jeripeierSBB:


src/components/autocomplete/autocomplete.ts
```diff
@@ -87,7 +87,15 @@ export class SbbAutocomplete extends LitElement {
 
   private _overlay: HTMLElement;
   private _optionContainer: HTMLElement;
-  private _originElement: HTMLElement;
+
+  private get _originElement(): HTMLElement {
```
kyubisation, line 91: Maybe this could even be public as `originElement`?

kyubisation (approved):
LGTM 👍

PR #2270, opened 2023-12-07
fix(sbb-select): adapt setup to work with Next.js hydration

kyubisation:
`<sbb-select>` causes a hydration error in Next.js. This PR implements a workaround by using the render logic of `@lit/react` to run our setup logic only after Next.js hydration or directly, if not in a Next.js environment.


jeripeierSBB (approved):
LGTM

PR #2271, opened 2023-12-07
feat(sbb-container): first implementation

### LINKED ISSUE #1908: story(sbb-container, sbb-sticky-bar): initial implementation
We want to implement a container component and associated sticky bar.
The container component has the same spacing behavior as the header and footer (including the `expanded` property).

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?type=design&node-id=25018%3A468481&mode=design&t=MDGURDdCkcBSQrfL-1)

**References**

- https://www.figma.com/proto/HeBonE3QY13ABQ4f2FShTJ/lyne-projects?page-id=0%3A1&type=design&node-id=2-182675&viewport=1321%2C1370%2C0.32&t=Kaic1GfC9eEVmjNn-1&scaling=min-zoom

**Design/Technical  Spec**

`<sbb-container>`

- Have the same spacing as `<sbb-header>` and `<sbb-footer>`, including a `expanded` property
- Use the sass mixin for spacing as is used in header and footer

`<sbb-sticky-bar>`

- Can be slotted into `<sbb-container>` and assumes its horizontal spacing
- Should stick to the bottom, while the container is visible
- Should have a upper shadow when it is sticky at the bottom
- Logic for the shadow should be self-contained in the `<sbb-sticky-bar>` (if possible)
- Sticky at the top is out-of-scope at the moment

```html
<sbb-container>
  my content
</sbb-container>

// sticky bar
<sbb-container>
  my content
  <sbb-sticky-bar>
    sticky content
  </sbb-sticky-bar>
</sbb-container>
```

---

_Original request_

### Preflight Checklist

### Request type

Request for a new component

Components

### Functionality

The bar should be created including the transition of the stickyness into the content. The content of the sticky bar is a placeholder. It is important that the spacing below the placeholder in the embedded state is subtracted from the section spacing that follows.

https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?type=design&node-id=25018%3A468482&mode=design&t=t3CgF6sOkvbr2UNz-1

MarioCastigliano:
## Preflight Checklist

## Issue

This PR Closes #1908 

## Pull request checklist

## Changes

Changes in this pull request:

-

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

- [ ] JAWS Firefox Desktop
- [ ] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [ ] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/container/container/container.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { expect, fixture } from '@open-wc/testing';
+import { html } from 'lit/static-html.js';
+import '.';
```
DavideMininni-Fincons, line 3: This could be removed

src/components/container/container/container.stories.ts
```diff
@@ -0,0 +1,112 @@
...
+const teaser = (): TemplateResult => html`
+  <sbb-teaser is-stacked>
+    <img slot="image" src=${placeholderImage} alt="400x300" />
+    <span slot="title">Example teaser</span>
+    <p slot="description">With a description.</p>
+  </sbb-teaser>
+`;
+
+const containerContent = (title: string, negative = false): TemplateResult => html`
+  <div style=${styleMap({ overflow: 'auto' })}>
```
DavideMininni-Fincons, line 27: `styleMap` can be removed here, since there's no variable interpolation

src/components/container/container/container.stories.ts
```diff
@@ -0,0 +1,112 @@
...
+};
+
+const defaultArgTypes: ArgTypes = {
+  expanded,
+  variant,
+};
+
+const defaultArgs: Args = {
+  expanded: false,
+  variant: 'transparent',
```
DavideMininni-Fincons, line 60: Use `variant.options[0]`

src/components/container/sticky-bar/readme.md
```diff
@@ -0,0 +1,50 @@
+> Explain the use and the purpose of the component; add minor details if needed and provide a basic example.<br>
```
DavideMininni-Fincons, line 1: Don't forget to write docs before changing to 'ready to review' :)

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,67 @@
...
+const WithStickybarTemplate = (): TemplateResult => html`
+  <sbb-container>
+    ${containerContent('Example title')} ${containerContent('Another one')}
+    ${containerContent('And another one')} ${containerContent('And a last one')}
+
+    <sbb-sticky-bar>
+      <sbb-button style=${styleMap({ marginInlineStart: 'auto' })}>Continue</sbb-button>
+    </sbb-sticky-bar>
+  </sbb-container>
+  <sbb-container variant="milk" style=${styleMap({ paddingBlock: '5rem' })}>
```
DavideMininni-Fincons, line 37: See other comments about `styleMap`

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,67 @@
...
+  render: Template,
+};
+
+export const WithContentAfter: StoryObj = {
+  render: WithStickybarTemplate,
+};
+
+const meta: Meta = {
+  decorators: [
+    (story) => html` <div style=${styleMap({ margin: '-1rem' })}>${story()}</div> `,
```
DavideMininni-Fincons, line 52: Same as others (`stylemap`)

DavideMininni-Fincons (commented):
Good work, I added some comments ;)

src/components/container/container/container.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (stickyBarPosition.bottom >= animationStart) {
+      const animationProgress = (stickyBarPosition.bottom - animationStart) / animationSpan;
+      this.style.setProperty('--sbb-sticky-bar-scroll', `${animationProgress}`);
+    } else {
+      this.style.setProperty('--sbb-sticky-bar-scroll', '0');
+    }
+  }
+
+  private get _stickyBarInnerElement(): HTMLDivElement {
+    return this.querySelector('sbb-sticky-bar')?.shadowRoot.querySelector('.sbb-sticky-bar');
```
TomMenga, line 77: Since Lit, we can avoid querying inside the shadow dom of other components.
I'd prefer the sticky bar to have an `@internal` public getter that exposes the element you need

src/components/container/container/container.ts
```diff
@@ -0,0 +1,110 @@
...
+  private _abort = new ConnectedAbortController(this);
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    const signal = this._abort.signal;
+
+    // Once the component is done rendering the container should find the first scrollable element
+    // This is necessary in case the container itself, or any parent element has a fixed height
+    super.getUpdateComplete().then(() => {
+      this._handlePageScroll();
```
TomMenga, line 33: Discussion point:
Shouldn't the sticky bar animation be the responsibility of the sticky bar component?

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,56 @@
...
+}
+
+.sbb-sticky-bar::before {
+  @include sbb.shadow-level-5-hard;
+
+  position: absolute;
+  display: inline-block;
+  content: '';
+  height: 100%;
+  width: 100vw;
```
TomMenga, line 27: Discussion point:
I understand why `width: 100vw;` is needed but it prevents any possible use of a non-full page sticky bar (imagine a page where you have a side menu for example).

The best solution would be to negate the spacing of the `sbb.page-spacing; | sbb.page-spacing-expanded` but I understand this might be hard to achieve.
A good compromise could be to a CSS variable defined on the `:host` so that the consumer  has a way to handle those scenarios

src/components/container/container/container.spec.ts
```diff
@@ -0,0 +1,13 @@
...
+import { html } from 'lit/static-html.js';
+import './container';
+
+describe('sbb-container', () => {
+  it('renders', async () => {
+    const root = await fixture(html`<sbb-container></sbb-container>`);
+
+    expect(root).dom.to.be.equal(`<sbb-container variant="transparent"></sbb-container>`);
+
+    await expect(root).shadowDom.to.equalSnapshot();
```
TomMenga, line 11: Just to be consistent with the sticky-bar test, we can avoid using `equalSnapshot()` since it is a very simple template

TomMenga (dismissed):
Good job. 👍 
Some small changes and some topic to discuss

Ps. Don't forget to approve the new stories on chromatic when you get the ok from the UX

src/components/container/sticky-bar/sticky-bar.scss
MarioCastigliano: We asked Manuel about this and he was positive the sticky bar is only meant to be used in full width

TomMenga: In any case, I think it is always better to anticipate the changes

MarioCastigliano: After discussing the matter with @kyubisation, we decided to try with an "infinite" sticky-bar with `overflow-x:hidden` on the sbb-container. 
But after some tests I noticed that any kind of overflow rule prevents the `position: sticky` from working. Another solution could be `contain: contents` (which works just like `overflow: hidden`), which does not cause any problems with the sticky positioning, but it messes up any overlay that goes outside the container. 
So we'll use the css variable approach

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,56 @@
...
+  display: contents;
+}
+
+.sbb-sticky-bar {
+  position: sticky;
+  display: flex;
+  width: 100%;
+  bottom: 0;
+  background-color: var(--sbb-sticky-bar-bg-color);
+  padding-block: var(--sbb-spacing-responsive-xxxs);
```
jeripeierSBB, line 17: I asked Florian if we really should provide a default padding. I remember we once decided to not include any default padding.

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,56 @@
...
+  @include sbb.shadow-level-5-hard;
+
+  position: absolute;
+  display: inline-block;
+  content: '';
+  height: 100%;
+  width: var(--sbb-sticky-bar-width);
+  top: 0;
+  left: 50%;
+  transform: translateX(-50%);
```
jeripeierSBB, line 30:
```suggestion
@include sbb.absolute-center-x
```

jeripeierSBB, line 30:
```suggestion
@include sbb.absolute-center-x
```

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,56 @@
...
+  background-color: var(--sbb-sticky-bar-bg-color);
+  animation: fade 1s linear 1 normal both paused;
```
jeripeierSBB, line 32: 1s seems long for an animation?

But first a general question, would it not just work if we detect sticky state of sticky bar and just us transition instead of complicated delayed animation? I remember I once implemented such a sticky bar in another project and also privately (www.lorenapeier.ch If you like to test it :) ). basically I just used an intersection observer. something like discussed here: https://stackoverflow.com/questions/16302483/event-to-detect-when-positionsticky-is-triggered 
This would save you from using scroll listener.

this slowly (depending on scroll position) fade in animation causes also bad situations where the shadow should be visible but is only very slightly seeable:
Bildschirmfoto 2023-12-20 um 10 37 01

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,56 @@
...
+  animation-delay: calc(var(--sbb-sticky-bar-scroll, 0) * -1s);
+}
+
+.sbb-sticky-bar__settled {
```
jeripeierSBB, line 36: it's not wrong but normally we use data attributes to describe state of a component instead of classes.

src/components/container/sticky-bar/sticky-bar.scss
jeripeierSBB: Other than that, when the sticky bar is not sticky anymore, the shadow is still visible (seen on https://lyne-components-preview-pr2271.app.sbb.ch/?path=/story/components-sbb-container-sbb-sticky-bar--with-content-after) which is somehow working in fullscreen but not in stories view. other than that in fullscreen the container below doesn't have a background color but in inline view it has.
Bildschirmfoto 2023-12-20 um 10 39 22

and sometimes the background color is transparent even if sticky:
Bildschirmfoto 2023-12-20 um 10 48 50

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const WithContentAfter: StoryObj = {
+  render: WithStickybarTemplate,
+};
+
+const meta: Meta = {
+  decorators: [
+    (story) => html` <div style="margin: -1rem">${story()}</div> `,
+    withActions as Decorator,
+  ],
```
jeripeierSBB, line 53:
```suggestion
```

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,66 @@
...
+  parameters: {
+    actions: {},
+    backgrounds: {
+      disable: true,
+    },
+    docs: {
+      extractComponentDescription: () => readme,
+    },
+  },
+  title: 'components/sbb-container/sbb-sticky-bar',
```
jeripeierSBB, line 63:
```suggestion
  layout: 'fullscreen',
  title: 'components/sbb-container/sbb-sticky-bar',
```

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,66 @@
...
+    <sbb-sticky-bar>
+      <sbb-button style="margin-inline-start: auto">Continue</sbb-button>
+    </sbb-sticky-bar>
+  </sbb-container>
+  <sbb-container variant="milk" style="padding-block: 5rem">
+    ${containerContent('Content after first container')} ${containerContent('Another one')}
+  </sbb-container>
+`;
+
+export const Standalone: StoryObj = {
```
jeripeierSBB, line 41: Please add some color variants here as well to show inheriting color of container

src/components/container/container/container.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+@customElement('sbb-container')
+export class SbbContainerElement extends LitElement {
+  public static override styles: CSSResultGroup = style;
+
+  /** Whether the container is expanded. */
+  @property({ type: Boolean, reflect: true }) public expanded = false;
+
+  /** Variant of the container, like transparent, white etc. */
+  @property({ reflect: true }) public variant: 'transparent' | 'white' | 'milk' | 'midnight' =
```
jeripeierSBB, line 19:
```suggestion
  @property({ reflect: true }) public color: 'transparent' | 'white' | 'milk' | 'midnight' =
```

jeripeierSBB (changes requested):
Thanks for the implementing these components! There are some things to address and also topics to discuss probably.

- I think we should use the sbb-container in our stories where possible, so at least in the home stories (instead of the grid, at least when container supports grid :) ).
- In page layout docs (src/storybook/styles/layout/readme.md) we should probably mention the new possibility of the sbb-container.

src/components/container/container/container.scss
```diff
@@ -0,0 +1,42 @@
...
+
+// Default component properties, defined for :host. Properties which can not
+// travel the shadow boundary are defined through this mixin
+@include sbb.host-component-properties;
+
+:host {
+  --sbb-container-width: 100vw;
+  --sbb-container-bg-color: var(--sbb-color-black-alpha-0);
+  --sbb-sticky-bar-width: var(--sbb-container-width);
+  --sbb-sticky-bar-bg-color: var(--sbb-color-white-default);
```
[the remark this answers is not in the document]
MarioCastigliano: In the transparent version they don't have the same color. As the container is actually set to transparent but the sticky-bar needs to be white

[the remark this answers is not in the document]
jeripeierSBB: yes but here on container level it's always the same, so I'd still argue for the change :)

src/components/container/container/container.ts
```diff
@@ -0,0 +1,58 @@
...
+      this._updateStickyBar();
+    }
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <div class="sbb-container">
+        <slot></slot>
+      </div>
+      <slot name="sticky-bar"></slot>
```
jeripeierSBB, line 48: on slotchange, the sticky bar should be updated with expanded as well. You can at anytime insert or remove a sticky bar into the container.

jeripeierSBB, line 48: on slotchange, the sticky bar should be updated with expanded as well. You can at anytime insert or remove a sticky bar into the container.

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    <sbb-sticky-bar>
+      <sbb-button style="margin-inline-start: auto">Continue</sbb-button>
+    </sbb-sticky-bar>
+  </sbb-container>
+`;
+
+const WithStickybarTemplate = ({ color, ...args }): TemplateResult => html`
+  <sbb-container ${sbbSpread(args)} color=${color}>
+    ${containerContent('Example title', color === 'midnight')}
```
jeripeierSBB, line 79: why not just passing color into containerContent and only once check for color === 'midnight'?

src/components/container/container/container.ts
jeripeierSBB: alternative approach is to check for parent's expanded in connectedcallback of the sticky bar.

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,153 @@
...
+
+export const WithContentAfter: StoryObj = {
+  render: WithStickybarTemplate,
+  argTypes: defaultArgTypes,
+  args: defaultArgs,
+};
+
+const meta: Meta = {
+  parameters: {
+    actions: {},
```
jeripeierSBB, line 141:
```suggestion
```

src/components/container/sticky-bar/sticky-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  protected override render(): TemplateResult {
+    setAttribute(this, 'slot', 'sticky-bar');
+    return html`
+      <div class="sbb-sticky-bar__wrapper">
+        <div class="sbb-sticky-bar">
+          <slot></slot>
+        </div>
+      </div>
+      <span></span>
```
jeripeierSBB, line 57: As it's made as block element in scss, we could directly use a div? I'd prefer to have a class name which describes the purpose of the div.
```suggestion
      <div class="sbb-sticky-bar__intersector"></div>
```

src/components/container/sticky-bar/sticky-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+  private _toggleShadowVisibility(entries: IntersectionObserverEntry[]): void {
+    entries.forEach((entry) => {
+      toggleDatasetEntry(this, 'settled', entry.isIntersecting || entry.boundingClientRect.top < 0);
+    });
+  }
+
+  private _updateIntersectionObserver(): void {
+    this._observer.disconnect();
+    if (!this._intersector) {
+      this._intersector = this.shadowRoot.querySelector('span');
```
jeripeierSBB, line 37:
```suggestion
      this._intersector = this.shadowRoot.querySelector('.sbb-sticky-bar__intersector');
```

src/components/container/sticky-bar/sticky-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+    if (this._intersector) {
+      this._observer.observe(this._intersector);
+    }
```
jeripeierSBB, line 41: can never be different
```suggestion
    this._observer.observe(this._intersector);
```

src/components/container/sticky-bar/sticky-bar.ts
```diff
@@ -0,0 +1,67 @@
...
+    this._updateIntersectionObserver();
+  }
+
+  private _toggleShadowVisibility(entries: IntersectionObserverEntry[]): void {
+    entries.forEach((entry) => {
+      toggleDatasetEntry(this, 'settled', entry.isIntersecting || entry.boundingClientRect.top < 0);
+    });
+  }
+
+  private _updateIntersectionObserver(): void {
```
jeripeierSBB, line 34: The only place where this function is called is firstUpdated(), so do we really need the disconnect and reconnection? I feel that it could be necessary to call it in connectedcallback as well. Can you test removing, moving and re adding a sticky bar in DOM?

jeripeierSBB (changes requested):
I like the new approach! There are some topics which needed to be addressed.

src/components/container/container/container.scss
```diff
@@ -0,0 +1,42 @@
...
+
+.sbb-container {
+  :host(:not([expanded])) & {
+    @include sbb.page-spacing;
+  }
+
+  :host([expanded]) & {
+    @include sbb.page-spacing-expanded;
+  }
+}
```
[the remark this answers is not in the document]
jeripeierSBB: discussed with @kyubisation, we don't create a property, we don't include grid in container for now.

src/components/container/container/container.stories.ts
```diff
@@ -0,0 +1,101 @@
+import type { InputType } from '@storybook/types';
+import type { ArgTypes, Args, Meta, StoryObj } from '@storybook/web-components';
+import { type TemplateResult, html } from 'lit';
+
+import { sbbSpread } from '../../core/dom';
+
+import '../../button';
+import '../../teaser';
```
jeripeierSBB, line 8:
```suggestion
```

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,155 @@
...
+    </sbb-sticky-bar>
+  </sbb-container>
+`;
+
+const ShortTemplate = ({ color, ...args }: Args): TemplateResult => html`
+  <sbb-container ${sbbSpread(args)} color=${color}>
+    ${containerContent('Example title', color)}
+
+    <sbb-sticky-bar>
+      <sbb-button style="margin-inline-start: auto">Continue</sbb-button>
```
jeripeierSBB, line 76: Maybe nest in an action group? What would be the best example here? @mcilurzo

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,38 @@
...
+
+.sbb-sticky-bar__wrapper {
+  box-shadow: none;
+  position: sticky;
+  inset-block-end: 0;
+  display: inline-block;
+  width: 100%;
+  background-color: var(--sbb-container-background-color, var(--sbb-color-white-default));
+  transition-property: box-shadow;
+  transition-duration: var(--sbb-sticky-bar-animation-duration);
```
jeripeierSBB, line 20: transition-timing-function missing. maybe try with var(--sbb-animation-easing) and introduce --sbb-sticky-bar-animation-easing var

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,38 @@
...
+  --sbb-sticky-bar-animation-duration: var(--sbb-animation-duration-5x);
+
+  display: contents;
+}
+
+.sbb-sticky-bar__wrapper {
+  box-shadow: none;
+  position: sticky;
+  inset-block-end: 0;
+  display: inline-block;
```
jeripeierSBB, line 16: It should not really change something, but this is clearly a block element for me :)
```suggestion
  display: block;
```

jeripeierSBB (changes requested):
Some few additional topics left:
- Possibility to define separate color for sticky bar when sticky.
- High Contrast Mode support
- In page layout docs (src/storybook/styles/layout/readme.md) we should mention the new possibility of the sbb-container.

src/components/container/sticky-bar/sticky-bar.stories.ts
mcilurzo: an action-group would be nice.

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,78 @@
...
+  :host([data-sticking]) & {
+    @include sbb.shadow-level-5-hard;
+
+    @include sbb.if-forced-colors {
+      border-top: var(--sbb-border-width-1x) solid CanvasText;
+    }
+  }
+
+  :host(:not([data-sticking])) & {
+    --sbb-sticky-bar-background-color: var(
```
jeripeierSBB: I'd not toggle vars on .css class level as it's one level than then the host and is therefore "not the same".
Second question, should it not be transparent when not sticking? you could also directly write: "background-color: transparent", or then just toggling it on host level

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,78 @@
...
+
+:host([color='white']) {
+  --sbb-sticky-bar-background-color: var(--sbb-color-white-default);
+}
+
+:host([color='milk']) {
+  --sbb-sticky-bar-background-color: var(--sbb-color-milk-default);
+}
+
+:host([color='midnight']:not([data-settled])) {
```
jeripeierSBB: data-settled is not existing? 
```suggestion
:host([color='midnight']) {
```

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,78 @@
...
+  :host([data-expanded]) & {
+    @include sbb.page-spacing-expanded;
+  }
+}
+
+.sbb-sticky-bar__intersector {
+  position: absolute;
+  width: 100%;
+  height: var(--sbb-sticky-bar-intersector-height);
+  transform: translateY(calc(var(--sbb-sticky-bar-intersector-height) * -1));
```
jeripeierSBB, line 81: Please write a short comment why the 1 pixel and transformation is necessary. Is it worth to have a css var for the intersector height, or could it be inlined as an internal detail?

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,196 @@
...
+    >
+      Link
+    </sbb-link>
+    <sbb-button variant="secondary" sbb-dialog-close> Cancel </sbb-button>
+    <sbb-button variant="primary" sbb-dialog-close> Confirm </sbb-button>
+  </sbb-action-group>
+`;
+
+const containerContent = (title: string, color: string): TemplateResult => html`
+  <div style="overflow: auto;">
```
jeripeierSBB, line 78: Same question, needed and why?

jeripeierSBB (changes requested):

src/components/container/sticky-bar/sticky-bar.scss
MarioCastigliano: On windows the animation dind't work on any browser unless I gave the intersector an height. I'll document it with a comment. Inlined is fine to me.

src/components/container/container/container.stories.ts
```diff
@@ -0,0 +1,100 @@
...
+import { sbbSpread } from '../../core/dom';
+
+import '../../button';
+import '../../title';
+import './container';
+
+import readme from './readme.md?raw';
+
+const containerContent = (title: string, color: string): TemplateResult => html`
+  <div style="overflow: auto">
```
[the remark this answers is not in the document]
MarioCastigliano: image

Otherwise when top element has a margin-top we get things like this

src/components/container/sticky-bar/sticky-bar.scss
jeripeierSBB: It's definitely not that simple as I thought ;)

src/components/container/container/container.scss
```diff
@@ -0,0 +1,33 @@
...
+:host([color='midnight']) {
+  --sbb-container-background-color: var(--sbb-color-midnight-default);
+  --sbb-focus-outline-color: var(--sbb-focus-outline-color-dark);
+
+  color: var(--sbb-color-white-default);
+}
+
+.sbb-container {
+  background-color: var(--sbb-container-background-color, transparent);
+  overflow-y: auto;
```
jeripeierSBB, line 24: Probably this creates a scrollbar if a consumer sets a fixed height on the container? With this I think we remove the possibility to let the overflow be controlled by consumer from outside. I'm not really understanding why margin in a slotted content does influence the parent container (here sbb-container). hmm. Will think about it :)

src/components/container/sticky-bar/sticky-bar.stories.ts
```diff
@@ -0,0 +1,194 @@
...
+  args: { ...defaultArgs, color: color.options[1] },
+};
+
+export const Milk: StoryObj = {
+  render: DefaultTemplate,
+  argTypes: defaultArgTypes,
+  args: { ...defaultArgs, color: color.options[2] },
+};
+
+export const Midnight: StoryObj = {
```
jeripeierSBB, line 169: for chromatic it would be great if we have two stories with a background color set in the container. one with color on sticky and one without. So we would have all combinations.

To test the sticking state in chromatic, maybe we can in some stories enforce a fixed height, so it shows sticky?

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,90 @@
...
+}
+
+.sbb-sticky-bar__wrapper::before {
+  position: absolute;
+  content: '';
+  background-color: var(--sbb-sticky-bar-background-color);
+  z-index: -1;
+  inset: 0;
+  transition-property: background-color, box-shadow;
+  transition-duration: var(--sbb-sticky-bar-animation-duration);
```
jeripeierSBB, line 50: We need the disableAnimation property to disable the animation

src/components/container/container/container.scss
jeripeierSBB: I tested it, it's ok how it is now :) sorry..

src/components/container/sticky-bar/sticky-bar.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    const container = this.closest('sbb-container');
+    if (container) {
+      toggleDatasetEntry(this, 'expanded', container.expanded);
+    }
+    if (this._intersector) {
+      this._observer.observe(this._intersector);
+    }
+
+    toggleDatasetEntry(this, 'disableAnimation', true);
```
jeripeierSBB, line 44:
```suggestion
```

src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -0,0 +1,93 @@
...
+}
+
+:host([data-sticking][color='midnight']) {
+  --sbb-sticky-bar-sticky-background-color: var(--sbb-color-midnight-default);
+  --sbb-focus-outline-color: var(--sbb-focus-outline-color-dark);
+
+  color: var(--sbb-color-white-default);
+}
+
+:host(:is([disable-animation], [data-disable-animation])) {
```
jeripeierSBB, line 36:
```suggestion
:host([disable-animation]) {
```

src/components/container/sticky-bar/sticky-bar.stories.ts
jeripeierSBB: I see new stories but without fixed height, or is the sticky bar sticky in chromatic anyways?

jeripeierSBB (changes requested):

jeripeierSBB (approved):
LGTM. Thanks for your endurance!

jeripeierSBB (approved):
LGTM

PR #2274, opened 2023-12-07
fix(sbb-notification): fix notification animation

### LINKED ISSUE #2257: bug(sbb-notification): animation stucks
Since lit migration, it seems that sbb-notification fade in animation isn't looking smooth.

dauriamarco:
Closes #2257

### DIFF
--- a/src/components/notification/notification.scss
+++ b/src/components/notification/notification.scss
@@ -37,7 +37,7 @@
   // to default color for cases where the notification is used in a negative context.
   --sbb-focus-outline-color: var(--sbb-focus-outline-color-default);
 
-  margin: 0;
+  margin: var(--sbb-notification-margin);
   transition: margin var(--sbb-notification-animation-duration)
     var(--sbb-notification-timing-function);
   animation: {
@@ -61,8 +61,6 @@
       var(--sbb-notification-timing-function),
     opacity var(--sbb-notification-animation-duration) var(--sbb-notification-animation-duration)
       var(--sbb-notification-timing-function);
-
-  margin: var(--sbb-notification-margin);
 }
 
 :host(:is([data-resize-disable-animation], [disable-animation])) {
--- a/src/components/notification/notification.scss
+++ b/src/components/notification/notification.scss
@@ -26,6 +26,7 @@
   --sbb-notification-timing-function: ease-in;
   --sbb-notification-transition: max-height var(--sbb-notification-animation-duration)
       var(--sbb-animation-duration-2x) var(--sbb-notification-timing-function),
+    margin var(--sbb-notification-animation-duration) var(--sbb-notification-timing-function),
     padding var(--sbb-notification-animation-duration) var(--sbb-animation-duration-2x)
       var(--sbb-notification-timing-function),
     border var(--sbb-notification-animation-duration) var(--sbb-animation-duration-2x)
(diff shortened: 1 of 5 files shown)


src/components/notification/notification.ts
```diff
@@ -202,7 +201,13 @@ export class SbbNotification extends LitElement {
         class="sbb-notification__wrapper"
         @transitionend=${(event: TransitionEvent) => this._onNotificationTransitionEnd(event)}
         @animationend=${(event: AnimationEvent) => this._onNotificationAnimationEnd(event)}
-        ${ref((el) => (this._notificationElement = el as HTMLElement))}
+        ${ref(async (el) => {
```
jeripeierSBB, line 204: Would it be possible to not use ref at all? I'ts just we know that with SSR it can cause problems. So when we use it as little as possible we will not run into problems once.

src/components/notification/notification.stories.ts
```diff
@@ -92,8 +92,7 @@ const pageContent = (): TemplateResult => html`
 
 const DefaultTemplate = (args: Args): TemplateResult => html`
   <sbb-notification
-    ${sbbSpread(args)}
-    disable-animation
```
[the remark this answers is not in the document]
dauriamarco: Nope, it wasn't adding the `disable-animation` attr and the notification would expand on page load.

src/components/notification/notification.ts
```diff
@@ -118,6 +117,14 @@ export class SbbNotificationElement extends LitElement {
     SbbNotificationElement.events.didClose,
   );
 
+  private _open(): void {
+    if (this._state === 'closed') {
+      this._state = 'opening';
+      this._willOpen.emit();
+      this.disableAnimation && this._handleOpening();
```
jeripeierSBB, line 124: is this still needed with 0.1ms animation duration?

jeripeierSBB (commented):
Sadly Chromatic problem still open. something is weird with the notification since lit migration :D

src/components/notification/notification.ts
dauriamarco: I surely need it in closing, since I also use `this.disableAnimation` in the `_setNotificationHeight()` method to determine whether to calculate the height or not. So in this case I would keep them in for consistency.

src/components/notification/notification.scss
```diff
@@ -40,6 +40,11 @@
   margin: 0;
   transition: margin var(--sbb-notification-animation-duration)
```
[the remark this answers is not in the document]
dauriamarco: We need it to animate the margin at the close of the notification as well.

src/components/notification/notification.ts
```diff
@@ -138,9 +145,13 @@ export class SbbNotificationElement extends LitElement {
     this._setInlineLinks();
   }
 
-  protected override firstUpdated(): void {
-    this._willOpen.emit();
+  protected override async firstUpdated(): Promise<void> {
+    this._notificationElement = this.shadowRoot.querySelector(
+      '.sbb-notification__wrapper',
+    ) as HTMLElement;
+    await this.updateComplete;
```
jeripeierSBB, line 152: Could you maybe explain in a code comment why we have to wait here?

jeripeierSBB (approved):

jeripeierSBB (approved):
LGTM, thanks!

PR #2275, opened 2023-12-07
refactor: rename components with Element suffix

kyubisation:
This PR renames the web components with the Element suffix (e.g. `SbbAccordion` to `SbbAccordionElement`). This prevents name conflicts when consuming both element and React component (and in the future also Angular component).


kyubisation:
@DavideMininni-Fincons, @federicoisepponfincons, @dauriamarco, @MarioCastigliano, @TomMenga Jeri convinced me to rename the components with the Element suffix in order to avoid naming conflicts for consumers and also us in the future. I apologize for not discussing and deciding this with the team. We want to create the "final" Lit release today and are pressed for time. Thank you for your understanding (and feel free to voice your grievances in chat 😄).

PR #2280, opened 2023-12-11
fix: chromatic instability

### LINKED ISSUE #2184: Resolve open chromatic problems
- Notification chromatic story is cropped -> asked Chromatic -> fixed height now
- Last image is blurred in Firefox -> made image smaller
- Menu has unwanted space at bottom (not really a problem, accepted)
- Breadcrumb: flickering -> smaller content text
- Selection-panel: flickering -> removed unnecessary code

DavideMininni-Fincons:
Closes #2184

Fix for Chromatic issues:
  - sbb-breadcrumb-group: 
      - changed label from 'Breadcrumb' to 'Link' to avoid issue on smaller breakpoints;
      - added `max-width` to handle expanded case in bigger breakpoints
  - sbb-selection-panel:
      - removed a duplicate emitter of did-* events when `disableAnimation` is true
      - with a delay of 9 sec, it seems to correctly load (tested with 4 consecutive builds - can this be trusted?)
  - sbb-image
      - removed useless class in css
      - adding a `max-width` seems to be solve the issue on Firefox

### DIFF
--- a/src/components/breadcrumb/breadcrumb-group/breadcrumb-group.stories.ts
+++ b/src/components/breadcrumb/breadcrumb-group/breadcrumb-group.stories.ts
@@ -158,7 +158,7 @@ export const CollapsedState: StoryObj = {
 const meta: Meta = {
   decorators: [
     (story) => html`
-      <div style="padding: 2rem;">
+      <div style="padding: 2rem; max-width: 1000px;">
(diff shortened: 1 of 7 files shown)


jeripeierSBB (approved):
Thank you very very much for investigating here!

PR #2288, opened 2023-12-12
docs(sbb-map-container): use `sbb-form-field-clear` component in example

jeripeierSBB:


TomMenga (approved):
LGTM 👍 
There are changes in chromatic that needs to be approved I think

PR #2211, opened 2023-11-21
feat(sbb-teaser): redesign

federicoisepponfincons:
BREAKING CHANGE: The property `isStacked` has been removed in favor of `alignment`. Please see the documentation for further info. The `description` is not clamped to two lines anymore (responsibility of consumer). The slotted `image` has now a default width of `300px`. The slot, formerly named `description`, has been replaced by the unnamed slot. Support of nested `p` elements dropped (invalid html).

Closes #1897

## Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/master/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/master/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [pull request tracker](https://github.com/lyne-design-system/lyne-components/pulls) for a Pull Request (PR) that matches the one I want to submit, without success.

## Issue

This PR Closes #1897 

## Pull request checklist

Please check if your PR fulfills the following requirements:

- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been reviewed and added / updated if needed (for bug fixes / features)

See [Review Guidelines](../REVIEW.md) for more information what is checked during review process.

## Changes

Changes in this pull request:

- sbb-chip has been included as optional element in the text part
- features 3 variants related to the text position: right-top, right-centered, bottom-left
- the user can set the size of the picture (by width) and the image ratio

## Browsers

I tested the build on the following browsers:

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [ ] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

I tested the build on the following browsers:

- [ ] JAWS Firefox Desktop
- [ ] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [ ] NVDA Chrome Desktop
- [ ] VoiceOver Safari Desktop
- [ ] VoiceOver Chrome Desktop
- [ ] VoiceOver Safari Mobile
- [ ] Android Accessibility Suite Chrome Mobile

## Pull request type

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] Documentation content changes
[...]

PR #2290, opened 2023-12-12
refactor: replace `namedSlotChangeHandlerAspect` with `NamedSlotStateController` functionality

kyubisation:
With `namedSlotChangeHandlerAspect` we had a functionality that checked for the existence of slotted elements and we rendered slots depending on that state. This is at odds with the server side rendering functionality from Lit. Additionally, the performance gain assumed from less HTML elements is minimal in comparison to the additional DOM operations needed.
This PR replaces the `namedSlotChangeHandlerAspect` with the `NamedSlotStateController`. This is a fundamental change, which instead of rendering slots depending on usage, always renders them, but adds a `data-slot-names` attributes with a list of slot names (and `unnamed` for the unnamed slot) space-separated.
This allows writing the necessary rules in purely CSS, which is from rendering perspective more performant.


src/components/toggle/toggle-option/toggle-option.scss
```diff
@@ -49,7 +49,7 @@ input[type='radio'] {
   border-radius: var(--sbb-toggle-option-border-radius);
   color: var(--sbb-toggle-option-color);
 
-  :host(:not([data-icon-only])) & {
+  :host([data-slot-names~='unnamed']:where([data-slot-names~='icon'], [icon-name])) & {
```
jeripeierSBB, line 52: with the new solution there is a glitch for the icon only variant. see here: https://lyne-components-preview-pr2290.app.sbb.ch/?path=/story/components-sbb-toggle-sbb-toggle--icons-only

jeripeierSBB: here even more noticeable in normal mode: https://lyne-components-preview-pr2290.app.sbb.ch/?path=/story/components-sbb-toggle-sbb-toggle--dynamic-width

kyubisation: Thanks for noticing. My logic for detecting whether something is slotted in a slot was too naive.
I fixed it now.

src/components/button/button.ts
```diff
@@ -129,10 +118,8 @@ export class SbbButtonElement extends LitElement implements LinkButtonProperties
     this._handlerRepository.disconnect();
   }
 
-  private _onLabelSlotChange(event: Event): void {
-    this._hasText = (event.target as HTMLSlotElement)
-      .assignedNodes()
-      .some((n) => !!n.textContent?.trim());
+  private _onLabelSlotChange(): void {
+    this._hasText = this._namedSlots.slots.has('unnamed');
```
jeripeierSBB, line 122: Maybe the field this._hasText is not needed anymore and so also this method at all? @TomMenga

src/components/core/common-behaviors/named-slot-state-controller.ts
```diff
@@ -20,15 +20,18 @@ import { ReactiveController, ReactiveControllerHost } from 'lit';
...
   // We avoid using AbortController here, as it would mean creating
   // a new instance for every NamedSlotStateController instance.
   private _slotchangeHandler = (event: Event): void => {
     this._syncSlots(event.target as HTMLSlotElement);
   };
 
-  public constructor(private _host: ReactiveControllerHost & Partial<HTMLElement>) {
+  public constructor(
+    private _host: ReactiveControllerHost & Partial<HTMLElement>,
+    private _onSlotStateChangeCallback = () => null,
```
jeripeierSBB, line 33: the changes here should not be needed. a callback is not necessary anymore. just use the var, it's updated automatically @TomMenga

src/components/button/button.ts
TomMenga: The answer is the same for both comments
We need the `@state _hasText` to trigger a new render of the component

We could argue that the `data-icon-only` could be replaced by a CSS rule, but I think that would make it less readable

TomMenga:
TODO:
Check if everything is fine after the Teaser #2211 has been merged

src/components/navigation/navigation-list/navigation-list.ts
```diff
@@ -52,18 +39,8 @@ export class SbbNavigationListElement extends SlotChildObserver(LitElement) {
...
-  }
-
-  public override disconnectedCallback(): void {
-    super.disconnectedCallback();
-    this._handlerRepository.disconnect();
-  }
-
   protected override render(): TemplateResult {
-    const hasLabel = !!this.label || this._namedSlots['label'];
+    const hasLabel = !!this.label || this._namedSlots.slots.has('label');
```
jeripeierSBB, line 43: maybe we don't need this condition anymore, as the id is now alwys there and the reference always valid. if it's empty, it's up to the screen reader to read nothing I think. maybe test it?

src/components/radio-button/radio-button/radio-button.scss
```diff
@@ -121,10 +121,20 @@
   }
 }
 
+slot[name='suffix'] {
```
jeripeierSBB, line 124: Is this really needed? normally slot has display: contents so it's not really there, and if there is no content, there is really nothing there.

src/components/selection-panel/selection-panel.ts
```diff
@@ -151,22 +147,16 @@ export class SbbSelectionPanelElement extends LitElement {
...
-  public override disconnectedCallback(): void {
-    super.disconnectedCallback();
-    this._handlerRepository.disconnect();
-  }
-
   private _updateSelectionPanel(): void {
     this._checked = this._input?.checked;
     this._state =
-      this._checked || this.forceOpen || !this._namedSlots['content'] ? 'opened' : 'closed';
+      this.forceOpen || (this._namedSlots['content'] && this._checked) ? 'opened' : 'closed';
```
jeripeierSBB, line 159: Should not `has()` be used here?
```suggestion
      this.forceOpen || (this._namedSlots.slots.has('content') && this._checked) ? 'opened' : 'closed';
```

src/components/skiplink-list/skiplink-list.scss
```diff
@@ -40,6 +40,12 @@
   }
 }
 
+.sbb-link-list-title {
+  :host(:not([data-slot-names~='title'], [title-content])) {
```
jeripeierSBB, line 44:
```suggestion
  :host(:not([data-slot-names~='title'], [title-content])) & {
```

src/components/skiplink-list/skiplink-list.scss
```diff
@@ -40,6 +40,12 @@
   }
 }
 
+.sbb-link-list-title {
```
jeripeierSBB, line 43: title is never visible. does this make any sense at all? Maybe to not read an empty label? Can you test behavior?

src/components/skiplink-list/skiplink-list.ts
```diff
@@ -70,19 +59,9 @@ export class SbbSkiplinkListElement extends SlotChildObserver(LitElement) {
...
-
-  public override disconnectedCallback(): void {
-    super.disconnectedCallback();
-    this._handlerRepository.disconnect();
-  }
-
   protected override render(): TemplateResult {
     let ariaLabelledByAttribute: Record<string, string> = {};
-    if (this._namedSlots.title || this.titleContent) {
+    if (this._namedSlots.slots.has('title') || this.titleContent) {
```
jeripeierSBB, line 64: As commented elsewhere, maybe condition is not needed at all as id is always there now. If still needed, it should update when condition changes (requestupdate or callback)

src/components/status/status.scss
```diff
@@ -61,6 +63,10 @@
 
 .sbb-status__title {
   margin-block: 0;
+
+  :host(:not([data-slot-names~='title'], [title-content])) & {
```
jeripeierSBB, line 67: would it not be more consistent?
```suggestion
  :host(:not(#{$has-title}) & {
```

jeripeierSBB (changes requested):
Thanks for you big effort! There are some issues and questions :) Thanks for addressing it

src/components/button/button.spec.ts
```diff
@@ -144,33 +156,20 @@ describe('sbb-button', () => {
...
 
-  it('should detect icon button', async () => {
-    const root = await fixture(
-      html`<sbb-button><sbb-icon name="app-icon-medium"></sbb-icon></sbb-button>`,
-    );
-
-    expect(root).to.have.attribute('data-icon-only');
-  });
-
-  it('should detect icon button when there is space around icon', async () => {
```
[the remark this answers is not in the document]
TomMenga: In general, a drawback of this approach is that we weakened the spec test snapshots and we can't easily test styles to compensate for the loss. However, we still have chromatic as a safe net.

src/components/checkbox/checkbox/checkbox.scss
```diff
@@ -114,6 +124,11 @@ input[type='checkbox'] {
 }
 
 .sbb-checkbox__label--icon {
+  :host(:not([icon-name], [data-slot-names~='icon'])) &,
+  :host([data-is-selection-panel-input]) & {
```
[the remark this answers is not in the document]
TomMenga: Don't you think it's a bit hard to read?
I'd rather keep the two conditions separated :/
What do you think?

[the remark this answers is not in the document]
jeripeierSBB: my concerns were more about what the output will be:
```
:host(:not([icon-name], [data-slot-names~='icon'])) .sbb-checkbox__label--icon {}
:host([data-is-selection-panel-input]) .sbb-checkbox__label--icon {}
```

vs.
```
host(:is(:not([icon-name], [data-slot-names~='icon']), [data-is-selection-panel-input]) .sbb-checkbox__label--icon
```

second solution would have less output to the browser. but it's marginal.

src/components/checkbox/checkbox/checkbox.scss
```diff
@@ -114,6 +124,11 @@ input[type='checkbox'] {
...
+    display: none;
```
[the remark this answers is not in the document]
TomMenga: Correct. In theory, there was a configuration that allowed you to have an icon that visually broke the selection panel.
I don't think it was meant to be allowed in the first place

src/components/dialog/dialog.scss
```diff
@@ -244,6 +248,11 @@
   margin-block-start: auto;
   background-color: var(--sbb-dialog-background-color);
   border-block-start: var(--sbb-dialog-footer-border);
+
+  :host(:not(:where([data-slot-names~='title'], [title-content]))) &,
+  :host(:not([data-slot-names~='action-group'])) & {
```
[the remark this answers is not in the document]
TomMenga: Nice catch. The `:where` is a leftover from the first implementation.
Here, I also feel by joining it became too hard to understand. But let's take a single decision on one of the previous feedbacks.

src/components/dialog/dialog.ts
```diff
@@ -402,8 +387,7 @@ export class SbbDialogElement extends LitElement {
   }
 
   protected override render(): TemplateResult {
-    const hasTitle = !!this.titleContent || this._namedSlots['title'];
-    const hasActionGroup = this._namedSlots['action-group'] && hasTitle;
+    const hasTitle = !!this.titleContent || this._namedSlots.slots.has('title');
```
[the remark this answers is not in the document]
TomMenga: Again, nice catch!
SlotChange does not trigger a new rendering.
In case a slotted title would be added/removed while the dialog is open, we would be in a faulty state.

src/components/dialog/dialog.ts
```diff
@@ -479,7 +461,7 @@ export class SbbDialogElement extends LitElement {
             >
               <slot></slot>
             </div>
-            ${this._initialized && hasActionGroup ? dialogFooter : nothing}
+            ${this._initialized ? dialogFooter : nothing}
```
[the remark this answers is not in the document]
TomMenga: We can. I'll leave a note on the PR to remind us that we need to test the SSR of the dialog.

TomMenga:
TODO:
- Test the SSR of the dialog component

src/components/checkbox/checkbox/checkbox.scss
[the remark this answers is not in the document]
TomMenga: I believe it's worth it. Maybe we can discuss it in the next refinement and what's the team feeling about it

src/components/link-list/link-list.ts
```diff
@@ -92,26 +81,16 @@ export class SbbLinkListElement extends SlotChildObserver(LitElement) {
...
-  public override disconnectedCallback(): void {
-    super.disconnectedCallback();
-    this._handlerRepository.disconnect();
-  }
-
   protected override render(): TemplateResult {
     let ariaLabelledByAttribute: Record<string, string> = {};
 
-    if (this._namedSlots.title || this.titleContent) {
+    if (this._namedSlots.slots.has('title') || this.titleContent) {
```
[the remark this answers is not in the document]
TomMenga: Correct. I chose to react to the specific slot because the callback is called a lot in this component

src/components/selection-panel/selection-panel.ts
TomMenga: this is bad :| 
I'll blame the holidays for it

src/components/skiplink-list/skiplink-list.scss
TomMenga: Semantically speaking there is a difference between not visible and not displayed and I feel like that should be preserved. But anyway, I'll test the label behaviour and see what happens

src/components/status/status.scss
TomMenga: Maybe yes :/
But, to be consistent, we must change all of the occurrences

jeripeierSBB: in the status.scss file I only see one occurrence. the others in this file do already use $has-title.

TomMenga: ah... my bad :/
i thought you wanted to change every file

src/components/skiplink-list/skiplink-list.scss
TomMenga: As you suggested, the title is ignored if empty.
We can both remove the CSS rule and the conditional `set ID` on the next feedback

src/components/skiplink-list/skiplink-list.ts
```diff
@@ -70,21 +62,10 @@ export class SbbSkiplinkListElement extends SlotChildObserver(LitElement) {
...
-    super.disconnectedCallback();
-    this._handlerRepository.disconnect();
-  }
-
   protected override render(): TemplateResult {
     let ariaLabelledByAttribute: Record<string, string> = {};
-    if (this._namedSlots.title || this.titleContent) {
-      ariaLabelledByAttribute = { 'aria-labelledby': 'sbb-skiplink-list-title-id' };
-    }
+    ariaLabelledByAttribute = { 'aria-labelledby': 'sbb-skiplink-list-title-id' };
```
jeripeierSBB, line 67: I'd vote for inline this in the template

TomMenga: Totally agreeable

src/components/checkbox/checkbox/checkbox.scss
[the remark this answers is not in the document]
kyubisation: Just checked; Prettier would format this as
```scss
:host(
    :is(
        :not([icon-name], [data-slot-names~="icon"]),
        [data-is-selection-panel-input]
      )
  )
  .sbb-checkbox__label--icon {
}
```

src/components/dialog/dialog.scss
```diff
@@ -249,8 +249,7 @@
   background-color: var(--sbb-dialog-background-color);
   border-block-start: var(--sbb-dialog-footer-border);
 
-  :host(:not([data-slot-names~='title'], [title-content])) &,
-  :host(:not([data-slot-names~='action-group'])) & {
+  :host(:not([data-slot-names~='title'], [title-content], [data-slot-names~='action-group'])) & {
```
jeripeierSBB, line 252: It seems that with :not() it is an "and" connection between selectors. So the last change probably broke chromatic. One approach would be using :where() or leave it as it was :)

MDN says: `You can negate several selectors at the same time. Example: :not(.foo, .bar) is equivalent to :not(.foo):not(.bar).`

TomMenga: yea, stupid error :/

jeripeierSBB (approved):
LGTM

PR #2304, opened 2023-12-19
fix(sbb-journey-header): add non breaking space for screen readers

kyubisation:
This PR adds non breaking white space to the `sbb-journey-header`, to avoid text concatination (e.g. `from Zürich to Bern`, instead of `fromZürichtoBern`).


jeripeierSBB (approved):
LGTM

PR #2308, opened 2023-12-25
fix(sbb-icon): fix preserving space during loading

jeripeierSBB:


jeripeierSBB:
@TomMenga I could not determine why you changed this logic once you migrated it to lit. The feature was lost that during loading, the space of the icon should be preserved. Maybe there is something I don't see now? Thanks for checking it :)

dauriamarco (approved):
LGTM 👍

TomMenga:
> @TomMenga I could not determine why you changed this logic once you migrated it to lit. The feature was lost that during loading, the space of the icon should be preserved. Maybe there is something I don't see now? Thanks for checking it :)

It probably was a misinterpretation due to the different timing in the spec test. Once migrated, I noticed that the icon has a `data-empty` attribute for a brief moment even if `icon-name` is set.
I thought, wrongly, that it was the migration fault and tried to preserve the tests as they were in stencil.

TomMenga (approved):

PR #2315, opened 2024-01-03
fix: fix using of new Date()

jeripeierSBB:


src/components/core/datetime/date-adapter.ts
```diff
@@ -145,7 +145,7 @@ export interface DateAdapter<T = any> {
   /** Get the given string as Date.
    * @param value The date in the format DD.MM.YYYY.
    */
```
MarioCastigliano, line 147:
```suggestion
  /** Get the given string as Date.
   * @param value The date in the format DD.MM.YYYY.
   * @param now The current date as Date.
   */
```

Or maybe something like that

MarioCastigliano (commented):
LGTM 👍

src/components/core/datetime/date-adapter.ts
jeripeierSBB: fixed, thanks!

MarioCastigliano (approved):

PR #2321, opened 2024-01-09
fix(sbb-toggle): initial animation glitch

### LINKED ISSUE #2319: bug(toggle): toggle has initial animation glitch
When loading the sbb-toggle, first, the white part is visible all over the toggle width, then it's reduced to the desired width. Can be seen on https://lyne-storybook.app.sbb.ch/?path=/story/components-sbb-toggle-sbb-toggle--size-m during first load.

dauriamarco:
Closes #2319

### DIFF
--- a/src/components/toggle/toggle/toggle.ts
+++ b/src/components/toggle/toggle/toggle.ts
@@ -67,6 +67,7 @@ export class SbbToggleElement extends LitElement {
   @property({ attribute: 'disable-animation', reflect: true, type: Boolean })
   public disableAnimation = false;
 
+  private _loaded = false;
   private _toggleElement: HTMLElement;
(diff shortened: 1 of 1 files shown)


src/components/toggle/toggle/toggle.ts
```diff
@@ -67,6 +67,7 @@ export class SbbToggleElement extends LitElement {
   @property({ attribute: 'disable-animation', reflect: true, type: Boolean })
   public disableAnimation = false;
 
+  private _loaded = false;
```
jeripeierSBB, line 70: would it work to just us hasUpdated of lit https://lit.dev/docs/components/lifecycle/?

jeripeierSBB (approved):
LGTM

PR #2325, opened 2024-01-09
fix(sbb-notification): fix accessibility

dauriamarco:
Fixes accessibility issues.

### DIFF
--- a/src/components/notification/notification.stories.ts
+++ b/src/components/notification/notification.stories.ts
@@ -240,7 +240,11 @@ const meta: Meta = {
         style="padding: 2rem;display: flex;gap: var(--sbb-spacing-fixed-4x);flex-direction: column;"
       >
         ${trigger(context.args)}
-        <div class="notification-container" style="display: flex; flex-direction: column;">
+        <div
+          class="notification-container"
+          style="display: flex; flex-direction: column;"
+          aria-live="polite"
+        >
           ${story()}
         </div>
         ${pageContent()}
--- a/src/components/notification/notification.stories.ts
+++ b/src/components/notification/notification.stories.ts
@@ -243,7 +243,7 @@ const meta: Meta = {
         <div
           class="notification-container"
           style="display: flex; flex-direction: column;"
-          aria-live="polite"
+          aria-live="off"
         >
           ${story()}
         </div>
--- a/src/components/notification/notification.stories.ts
+++ b/src/components/notification/notification.stories.ts
@@ -243,7 +243,7 @@ const meta: Meta = {
         <div
           class="notification-container"
           style="display: flex; flex-direction: column;"
-          aria-live="off"
+          aria-live="polite"
         >
           ${story()}
         </div>
--- a/src/components/notification/notification.ts
+++ b/src/components/notification/notification.ts
@@ -143,7 +143,7 @@ export class SbbNotificationElement extends LitElement {
     this._setNotificationHeight();
     this._open();
     this.shadowRoot
-      .querySelector('.sbb-notification')
+      ?.querySelector<SbbNotificationElement>('.sbb-notification')!
       .setAttribute('role', this.getAttribute('role') ?? 'status');
   }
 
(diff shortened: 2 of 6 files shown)


src/components/notification/notification.ts
```diff
@@ -202,6 +202,7 @@ export class SbbNotificationElement extends LitElement {
   protected override render(): TemplateResult {
     const hasTitle = !!this.titleContent || this._namedSlots['title'];
 
+    setAttribute(this, 'role', 'status');
```
jeripeierSBB, line 205: Not sure if it is better to delegate this to the consumer so it doesn't have any default behavior. on the other hand, it seems a good default :). @kyubisation meaning?

jeripeierSBB: can it be overridden, e.g. with role alert from outside? I feel atm. every re render would override it again.

jeripeierSBB (changes requested):
Thanks for getting into this!

src/components/notification/notification.ts
kyubisation: My gut feeling would be to set this in the `connectedCallback`, if it is not yet set.
e.g. `this.role = this.role ?? 'status';` (we should be able to use `this.role` now, I think).

jeripeierSBB: I think there was something preventing using this.role. maybe tests or chromatic oder something else? we can try again ;) I could not find mdn docs so far. 
Anyways what my main concern is, is basically you have an enclosing container and apply role status there and watch there for changes (like appearing notification). But somehow it works only if notification itself has role status as well.
But in general I think it's ok to predefine role status as long as there is an option to override it.

kyubisation: The same could be achieved with `this.setAttribute('role', this.getAttribute('role') ?? 'status')`, if we can't use this.role yet, I guess.

jeripeierSBB: I'd prefer the `this.setAttribute('role', this.getAttribute('role') ?? 'status')` because it's at one place

jeripeierSBB: You can completely move the role attribute setting into connectedcallback

jeripeierSBB (approved):
LGTM

src/components/notification/notification.ts
```diff
@@ -141,6 +141,9 @@ export class SbbNotificationElement extends LitElement {
     await this.updateComplete;
     this._setNotificationHeight();
     this._open();
+    this.shadowRoot
+      ?.querySelector<SbbNotificationElement>('.sbb-notification')!
+      .setAttribute('role', this.getAttribute('role') ?? 'status');
```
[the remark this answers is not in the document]
jeripeierSBB: @dauriamarco @DavideMininni-Fincons @kyubisation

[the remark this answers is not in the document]
kyubisation: This feels a bit "hacky". Is the behavior consistent with role "status" and "alert"?
I wonder if it would be better to add accessibility documentation (e.g. add role="status" to parent)?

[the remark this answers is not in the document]
dauriamarco: I agree, I would probably remove the `role` attribute from the inner element and document which role we suggest to put on the host itself.

src/components/notification/notification.stories.ts
```diff
@@ -240,7 +240,11 @@ const meta: Meta = {
         style="padding: 2rem;display: flex;gap: var(--sbb-spacing-fixed-4x);flex-direction: column;"
       >
         ${trigger(context.args)}
-        <div class="notification-container" style="display: flex; flex-direction: column;">
+        <div
+          class="notification-container"
+          style="display: flex; flex-direction: column;"
+          aria-live="off"
```
jeripeierSBB, line 246:
```suggestion
          aria-live="polite"
```
?

jeripeierSBB (approved):
LGTM, great!

PR #2327, opened 2024-01-10
fix(sbb-toggle): deal with undefined option component

jeripeierSBB:


src/components/toggle/toggle/toggle.ts
```diff
@@ -75,8 +75,15 @@ export class SbbToggleElement extends LitElement {
 
   private _valueChanged(value: any | undefined): void {
     const options = this._options;
+    // If options are not yet defined, we can check if attribute is already set as a fallback.
```
kyubisation, line 78:
```suggestion
    // If options are not yet defined web components, we can check if attribute is already set as a fallback.
```

PR #2330, opened 2024-01-11
fix(sbb-notification): border left stability

dauriamarco:


src/components/notification/notification.scss
```diff
@@ -129,6 +129,7 @@
   color: var(--sbb-notification-color);
   border-radius: var(--sbb-notification-border-radius);
   overflow: hidden;
+  z-index: 1;
```
jeripeierSBB, line 132: is this for sure creating a local context and not using the global? otherwise we maybe would have even more problems :D

dauriamarco: It should not create problems, the z-index of the host component always prevails on this inner div, but at the same time I don't really need it so I'll remove it

PR #2334, opened 2024-01-12
fix: remove setTimeout from SlotChildObserver

dauriamarco:
Our homepage is currently unstable and would crash on refresh, the problem only occurs on Mac and Chrome when there are a number of components on the page that depend on each other and are nested for more than two levels, it basically happens in the navigation with its various internal components (nav -> marker -> list -> action...). This is due to the new logic to do the children check with the `checkChildren` method implemented to optimise the SSR, by removing the `setTimeout` from the `firstUpdated` callback of the SlotChildObserver the problem is fixed.


kyubisation (approved):
LGTM 👍 
Thanks again 😃

PR #2337, opened 2024-01-15
fix: improve creation of react component

jeripeierSBB:
- Avoid setting element property if already set as attribute.
- Always set id as attribute to ensure id bindings


kyubisation (approved):
LGTM 👍

PR #2340, opened 2024-01-16
fix: adopt sass import guidelines

### LINKED ISSUE #2307: Self referential @use URLs
https://github.com/lyne-design-system/lyne-components/blob/2196fda52f53b09de5e75db4b6541bfb240b001f/src/components/core/styles/core/mediaqueries.scss#L5

There is at least one import which refers to distributions files in an odd way.

`@use 'node_modules/@sbb-esta/lyne-design-tokens/dist/scss/sbb-variables.scss' as sbb-tokens;`

Hopefully a better way can be found to include this.

Perhaps `@use '~@sbb-esta/lyne-design-tokens/dist/scss/sbb-variables.scss' as sbb-tokens;` or
`@use '~lyne-design-tokens/dist/scss/sbb-variables.scss' as sbb-tokens;`.

When running webpack from a working directory other than the one for the project, the compilation blows up because the sass-loader cannot find sub-variables.scss.
I have to fix this by creating an alias for the import.

jeripeierSBB:
Closes #2307

https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "@lit/react": "^1.0.1",
     "@open-wc/lit-helpers": "0.6.0",
     "@open-wc/testing": "4.0.0",
-    "@sbb-esta/lyne-design-tokens": "0.3.7",
+    "@sbb-esta/lyne-design-tokens": "0.3.8",
     "@storybook/addon-a11y": "7.6.8",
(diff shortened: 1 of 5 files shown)


kyubisation (approved):
LGTM 👍

PR #2359, opened 2024-01-22
refactor: extract base functionality of list components into `NamedSlotListElement`

kyubisation:
This PR refactors the list components, by extracting the common functionality into the `NamedSlotListElement` base class.
It also refactors the `SlotChildObserver` to be mostly synchronous, instead of asynchronous.

### DIFF
--- a/src/components/breadcrumb/breadcrumb/breadcrumb.ts
+++ b/src/components/breadcrumb/breadcrumb/breadcrumb.ts
@@ -50,6 +50,7 @@ export class SbbBreadcrumbElement extends SlotChildObserver(LitElement) {
 
   public constructor() {
     super();
+    /** @ignore id is already a well known property. */
     this.id = this.id || `sbb-breadcrumb-${nextId++}`;
   }
 
--- a/src/components/breadcrumb/breadcrumb/breadcrumb.ts
+++ b/src/components/breadcrumb/breadcrumb/breadcrumb.ts
@@ -15,6 +15,8 @@ import style from './breadcrumb.scss?lit&inline';
 
 import '../../icon';
 
+let nextId = 1;
+
 /**
  * It displays a link to a page; used within a `sbb-breadcrumb-group` it can display the path to the current page.
  *
@@ -46,14 +48,16 @@ export class SbbBreadcrumbElement extends SlotChildObserver(LitElement) {
 
   @state() private _hasText = false;
 
+  public constructor() {
+    super();
+    this.id = this.id || `sbb-breadcrumb-${nextId++}`;
+  }
+
   private _language = new LanguageController(this);
   private _handlerRepository = new HandlerRepository(this, actionElementHandlerAspect);
 
   public override connectedCallback(): void {
     super.connectedCallback();
-    this._hasText = Array.from(this.childNodes ?? []).some(
-      (n) => !(n as Element).slot && n.textContent?.trim(),
-    );
     this._handlerRepository.connect();
   }
 
@@ -63,9 +67,9 @@ export class SbbBreadcrumbElement extends SlotChildObserver(LitElement) {
   }
 
   protected override checkChildren(): void {
-    this._hasText = !!this.shadowRoot!.querySelector<HTMLSlotElement>('slot:not([name])')
-      ?.assignedNodes()
-      .some((n) => !!n.textContent?.trim());
+    this._hasText = Array.from(this.childNodes ?? []).some(
+      (n) => !(n as Element).slot && n.textContent?.trim(),
+    );
   }
 
   protected override render(): TemplateResult {
--- a/src/components/breadcrumb/breadcrumb/breadcrumb.ts
+++ b/src/components/breadcrumb/breadcrumb/breadcrumb.ts
@@ -15,8 +15,6 @@ import style from './breadcrumb.scss?lit&inline';
 
 import '../../icon';
 
-let nextId = 1;
-
 /**
  * It displays a link to a page; used within a `sbb-breadcrumb-group` it can display the path to the current page.
  *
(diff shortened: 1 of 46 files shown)


src/components/core/common-behaviors/named-slot-list-element.ts
```diff
@@ -0,0 +1,132 @@
...
+    this.listChildren.forEach((child, index) => {
+      child.setAttribute('slot', `${SLOTNAME_PREFIX}-${index}`);
+      this.formatChild?.(child);
+    });
+
+    // Remove the ssr attribute, once we have actually initialized the children elements.
+    this.removeAttribute(SSR_CHILD_COUNT_ATTRIBUTE);
+  }
+
+  protected formatChild?(child: C): void;
```
kyubisation, line 70: Replace with `willUpdate`

src/components/core/common-behaviors/named-slot-list-element.ts
```diff
@@ -0,0 +1,132 @@
+import type { TemplateResult } from 'lit';
+import { LitElement, html, nothing } from 'lit';
+import { state } from 'lit/decorators.js';
+
+import { SlotChildObserver } from './slot-child-observer';
+
+const SSR_CHILD_COUNT_ATTRIBUTE = 'data-ssr-child-count';
+const SLOTNAME_PREFIX = 'child';
```
kyubisation, line 8: Rename to li

src/components/core/common-behaviors/slot-child-observer.ts
```diff
@@ -22,51 +20,67 @@ export const SlotChildObserver = <T extends Constructor<LitElement>>(
...
+      // is available before createRenderRoot is called.
+      this._needsHydration = !!this.shadowRoot;
+      return super.createRenderRoot();
+    }
+
+    protected override willUpdate(changedProperties: PropertyValues): void {
+      super.willUpdate(changedProperties);
+      // If hydration is not needed we can immediately check for children
+      // in the first update.
+      if (!this._needsHydration && !this.hasUpdated) {
```
kyubisation, line 52: Check if hydration can already be skipped

src/components/navigation/navigation-marker/navigation-marker.ts
```diff
@@ -68,62 +59,37 @@ export class SbbNavigationMarkerElement extends SlotChildObserver(LitElement) {
     this.reset();
     action.active = true;
     this._currentActiveAction = action;
-    this._hasActiveAction = true;
+    setTimeout(() => this._setMarkerPosition());
```
jeripeierSBB, line 62: should we use updatescheduler here? at least it did no cause any problems so far.

src/components/navigation/navigation-marker/navigation-marker.ts
```diff
@@ -15,42 +13,35 @@ import style from './navigation-marker.scss?lit&inline';
...
 
   private _updateMarkerActions(): void {
-    for (const action of this._navigationActions) {
+    for (const action of this.listChildren) {
       action.size = this.size;
     }
 
-    this._hasActiveAction = !!this._activeNavigationAction;
-    this._currentActiveAction = this._activeNavigationAction;
+    this._currentActiveAction = this.listChildren.find((action) => action.active);
```
jeripeierSBB, line 44: should we include fallback to attribute here? we had a similar case if child is not yet initialized

src/components/train/train-wagon/train-wagon.ts
```diff
@@ -204,31 +184,11 @@ export class SbbTrainWagonElement extends SlotChildObserver(LitElement) {
...
-                  ? html`<span class="sbb-screenreaderonly">
-                      ${i18nAdditionalWagonInformationHeading[this._language.current]}
-                    </span>`
-                  : nothing}
-                <slot></slot>
-              </span>
+          ? html`<span class="sbb-train-wagon__icons" ?hidden=${this.listChildren.length === 0}>
+              ${this.renderList({
+                class: 'sbb-train-wagon__icons-list',
+                ariaLabel: i18nAdditionalWagonInformationHeading[this._language.current],
```
jeripeierSBB, line 190: I think a "presentation" element can't have an ariaLabel, this is why before it was added in a span. This could be added into the mixin. for the labelledby casees it's more difficult. But I also see that there were a lot of cases where this was not handled I think it should be also before.

jeripeierSBB (changes requested):
Thanks for addressing this

src/components/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -46,14 +48,17 @@ export class SbbBreadcrumbElement extends SlotChildObserver(LitElement) {
 
   @state() private _hasText = false;
 
+  public constructor() {
+    super();
+    /** @ignore id is already a well known property. */
+    this.id = this.id || `sbb-breadcrumb-${nextId++}`;
```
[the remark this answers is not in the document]
kyubisation: No, that is impossible. It is however possible, that the instances have been newly created. The id increments for every instance created, even if it is not rendered.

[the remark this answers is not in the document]
jeripeierSBB: Maybe it's related to the other bug on the breadcrumb stories. There is happening something unexpected somehow :D

src/components/navigation/navigation-marker/navigation-marker.ts
kyubisation: Should not happen here, because it is user interaction triggered.

kyubisation: You are right. I did not think about that. 😅

src/components/breadcrumb/breadcrumb/breadcrumb.ts
```diff
@@ -63,9 +68,9 @@ export class SbbBreadcrumbElement extends SlotChildObserver(LitElement) {
   }
 
   protected override checkChildren(): void {
-    this._hasText = !!this.shadowRoot!.querySelector<HTMLSlotElement>('slot:not([name])')
-      ?.assignedNodes()
-      .some((n) => !!n.textContent?.trim());
+    this._hasText = Array.from(this.childNodes ?? []).some(
```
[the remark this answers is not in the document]
kyubisation: As discussed; Might require another ssr-extension

jeripeierSBB (approved):
LGTM

PR #2375, opened 2024-01-30
refactor: improve css

jeripeierSBB:
only refactoring with one exception: not enforcing white background if not sticking


src/components/container/sticky-bar/sticky-bar.scss
```diff
@@ -6,33 +6,32 @@
   --sbb-sticky-bar-padding-block: var(--sbb-spacing-responsive-xxxs);
   --sbb-sticky-bar-animation-duration: var(--sbb-animation-duration-5x);
   --sbb-sticky-bar-animation-easing: var(--sbb-animation-easing);
-  --sbb-sticky-bar-background-color: var(
+
+  display: contents;
+}
+
+:host([data-sticking]) {
```
jeripeierSBB, line 13: only defining --sbb-sticky-bar-sticky-background-color when it's really sticking, which makes the not selectors obsolete

MarioCastigliano: Love the new approach 😄

MarioCastigliano (approved):

PR #2377, opened 2024-01-30
feat(sbb-screenreader-only): initial implementation

jeripeierSBB:


src/components/screenreader-only/screenreader-only.e2e.ts
```diff
@@ -0,0 +1,16 @@
+import { assert, fixture } from '@open-wc/testing';
+import { html } from 'lit/static-html.js';
+
+import { SbbScreenreaderOnlyElement } from './screenreader-only';
+
+describe('sbb-screenreader-only', () => {
+  let element: SbbScreenreaderOnlyElement;
+
+  beforeEach(async () => {
+    element = await fixture(html`<sbb-screenreader-only></sbb-screenreader-only>`);
```
kyubisation, line 10: Should this maybe contain text? 😅

src/components/screenreader-only/screenreader-only.ts
```diff
@@ -0,0 +1,26 @@
...
+ * This component can be used to visually hide content but present it to screen readers.
+ *
+ * @slot - Use the unnamed slot to provide content.
+ */
+@customElement('sbb-screenreader-only')
+export class SbbScreenreaderOnlyElement extends LitElement {
+  public static override styles: CSSResultGroup = style;
+
+  protected override render(): TemplateResult {
+    return html`<span><slot></slot></span>`;
```
kyubisation, line 17: I would vote to keep this completely flat, without even the span. This is an internal component for a very specific use case. If someone breaks it, its completely their own fault.

kyubisation (approved):
LGTM 👍

PR #2378, opened 2024-01-30
feat(a11y): implemented a11y tree snapshot function

TomMenga:


src/components/selection-panel/selection-panel.spec.ts
```diff
@@ -2,6 +2,7 @@ import { expect, fixture } from '@open-wc/testing';
 import { html } from 'lit/static-html.js';
 
 import { waitForLitRender } from '../core/testing';
+import { testA11yTreeSnapshot } from '../core/testing/a11y-tree-snapshot';
```
TomMenga, line 5: I haven't added the `testA11yTreeSnapshot` to the `core/testing/index` file because we don't want the `testA11yTreeSnapshot` dependencies to be included in the `core` esModule.
@kyubisation Does it make sense?

web-test-runner.config.js
```diff
@@ -50,6 +51,7 @@ export default {
   testRunnerHtml: (testFramework) => `
     <html>
       <head>
+        <meta name="testEnvironment" ${isDebugMode ? 'debug' : ''}>
```
TomMenga, line 54: Initially, the idea was to skip a11y snapshot tests on puppeteer. Since it seems that there is no way to discern which launcher we are using, the only alternative that came to my mind is to add some meta-data here and use it to discern if we launched the tests in debug mode (which also means that we are using puppeteer)

.github/workflows/continuous-integration.yml
```diff
@@ -52,6 +52,8 @@ jobs:
         run: yarn test
         env:
           NODE_ENV: production
+      - name: 'Assert no new snapshots (run `yarn test --ci` if this fails)'
+        run: git diff --exit-code
```
TomMenga, line 56: @kyubisation I don't know much about github actions :/
Confirm if it's correct pls

kyubisation: Looks correct to me 👍 
(This basically runs "git diff --exit-code" and if the exit code is != 0, it errors. In the logs this will be shown as a step in the test job, with the given name.)

src/components/core/dom/platform.ts
```diff
@@ -70,3 +70,25 @@ export const isSafari = (): boolean =>
 
 /** Whether the application is being rendered in a Next.js environment. */
 export const isNextjs = (): boolean => !!(globalThis as { next?: object }).next;
+
+/**
+ * Note: `userAgentData` is still sperimental. If possible, avoid to use it.
```
kyubisation, line 75:
```suggestion
 * Note: `userAgentData` is still experimental. If possible, avoid to use it.
```

src/components/selection-panel/selection-panel.spec.ts
kyubisation: Yes, I agree. 👍

web-test-runner.config.js
kyubisation: Nice, clean workaround 👍

kyubisation (approved):
LGTM 👍 
Nice work!

PR #2382, opened 2024-02-01
test: implemented a11y tree snapshot

TomMenga:
With this PR we're applying the `a11yTreeEqualSnapshot()` (introduced in #2378) to all the components.

The following components have been temporarily excluded to avoid unmergeable conflicts:
- button
- card-action
- header-action
- link
- menu
- menu-action
- navigation folder

### TODO
After this PR is merged, we need to clean the snapshot files from deprecated test cases (delete the files and run `yarn test --ci`)

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.ts
+++ b/src/components/autocomplete/autocomplete.spec.ts
@@ -2,7 +2,7 @@ import { expect, fixture } from '@open-wc/testing';
 import { html } from 'lit/static-html.js';
 
 import { isSafari } from '../core/dom';
-import { waitForLitRender } from '../core/testing';
+import { describeIf, waitForLitRender } from '../core/testing';
 import { testA11yTreeSnapshot } from '../core/testing/a11y-tree-snapshot';
 import type { SbbFormFieldElement } from '../form-field';
 
@@ -28,7 +28,7 @@ describe('sbb-autocomplete', () => {
       await waitForLitRender(elem);
     });
 
-    (!isSafari() ? describe : describe.skip)('Chrome-Firefox', async () => {
+    describeIf(!isSafari(), 'Chrome-Firefox', async () => {
       it('Dom', async () => {
         await expect(elem).dom.to.be.equalSnapshot();
       });
@@ -38,7 +38,7 @@ describe('sbb-autocomplete', () => {
       });
     });
 
-    (isSafari() ? describe : describe.skip)('Safari', async () => {
+    describeIf(isSafari(), 'Safari', async () => {
       it('Dom', async () => {
         await expect(elem).dom.to.be.equalSnapshot();
       });
@@ -65,7 +65,7 @@ describe('sbb-autocomplete', () => {
       await waitForLitRender(root);
     });
 
-    (!isSafari() ? describe : describe.skip)('Chrome-Firefox', async () => {
+    describeIf(!isSafari(), 'Chrome-Firefox', async () => {
       it('Dom', async () => {
         await expect(root).dom.to.be.equalSnapshot();
       });
@@ -75,7 +75,7 @@ describe('sbb-autocomplete', () => {
       });
     });
 
-    (isSafari() ? describe : describe.skip)('Safari', async () => {
+    describeIf(isSafari(), 'Safari', async () => {
       it('Dom', async () => {
         await expect(root).dom.to.be.equalSnapshot();
       });
--- a/src/components/autocomplete/autocomplete.spec.ts
+++ b/src/components/autocomplete/autocomplete.spec.ts
@@ -1,96 +1,90 @@
 import { expect, fixture } from '@open-wc/testing';
 import { html } from 'lit/static-html.js';
 
+import { isSafari } from '../core/dom';
+import { waitForLitRender } from '../core/testing';
+import { testA11yTreeSnapshot } from '../core/testing/a11y-tree-snapshot';
+import type { SbbFormFieldElement } from '../form-field';
+
+import type { SbbAutocompleteElement } from './autocomplete';
 import '../form-field';
 import '../option';
 import './autocomplete';
-import { isSafari } from '../core/dom';
 
 describe('sbb-autocomplete', () => {
-  it('renders standalone', async () => {
-    await fixture(html`
-      <div id="origin"></div>
-      <input id="trigger" />
-      <sbb-autocomplete origin="origin" trigger="trigger">
-        <sbb-option value="1">1</sbb-option>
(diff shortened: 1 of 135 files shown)


src/components/select/select.spec.ts
```diff
@@ -1,86 +1,86 @@
...
+
+    (isSafari() ? describe : describe.skip)('Safari', async () => {
+      it('Dom', async () => {
+        await expect(elem).dom.to.be.equalSnapshot();
+      });
+
+      it('ShadowDom', async () => {
+        await expect(elem).shadowDom.to.be.equalSnapshot();
+      });
+    });
```
DavideMininni-Fincons, line 45: I saw this a couple of time - what's the issue with Safari?

DavideMininni-Fincons (changes requested):
General change: 
i see that many times the testA11yTreeSnapshot is used with template, but with title = undefined.
I prefer to invert the parameters in the method signature, e.g.
`testA11yTreeSnapshot(template = null, title = 'whatever')`

src/components/select/select.spec.ts
TomMenga: Some components have a different template or behavior depending on the browser. In those cases, we cannot use snapshots unless we separate them by browser. And essentially this is what I did here

TomMenga:
> General change: i see that many times the testA11yTreeSnapshot is used with template, but with title = undefined. I prefer to invert the parameters in the method signature, e.g. `testA11yTreeSnapshot(template = null, title = 'whatever')`

I think I agree, I'll change it asap

src/components/autocomplete/autocomplete.spec.ts
```diff
@@ -1,96 +1,90 @@
...
-        <div class="sbb-autocomplete__panel">
-          <div class="sbb-autocomplete__wrapper">
-            <div class="sbb-autocomplete__options" ${!isSafari() ? listboxAttr : ''}>
-              <slot></slot>
-            </div>
-          </div>
-        </div>
-      </div>
-    `);
+    (!isSafari() ? describe : describe.skip)('Chrome-Firefox', async () => {
```
MarioCastigliano, line 31:
```suggestion
    // On top of the file
    const describeSkipSafari = !isSafari() ? describe : describe.skip;
    //...
    describeSkipSafari('Chrome-Firefox', async () => {
```
Do you think something like this could improve readability? Also we would only check if it's Safari once per test run

src/components/select/select.spec.ts
```diff
@@ -1,86 +1,86 @@
...
+        <sbb-select>
+          <sbb-option value="1">Option 1</sbb-option>
+          <sbb-option value="2">Option 2</sbb-option>
+          <sbb-option value="3">Option 3</sbb-option>
+        </sbb-select>
+      `);
+      await waitForLitRender(elem);
+    });
+
+    (!isSafari() ? describe : describe.skip)('Chrome-Firefox', async () => {
```
MarioCastigliano, line 27: Same question as before

MarioCastigliano (commented):
Davide anticipated me on the argument swap for the testA11yTreeSnapshot function 😅, I was gonna suggest the order `testA11yTreeSnapshot(template, skip, name)` since `skip` has a handful of usages while `name` is only used in the sbb-checkbox (just a nitpick).

src/components/autocomplete/autocomplete.spec.ts
TomMenga: Good idea! 
I would implement something like this in an utility file 
```js
const describeIf = (condition: boolean) => condition ? describe : describe.skip;
const testIf = (condition: boolean) => condition ? test : test.skip;

describeIf(isSafari())('aaaa', () => {});
```

MarioCastigliano: Looks a lot better 👍

TomMenga: I implemented it a bit differently, have a look if you like it

src/components/core/testing/a11y-tree-snapshot.ts
```diff
@@ -22,10 +25,16 @@ async function a11yTreeEqualSnapshot(): Promise<void> {
 /**
  * The function creates and tests the accessibility tree snapshot on each browser.
  * If a template is passed, it will be instantiated before the snapshot is taken.
+ * Note:
+ * We skip a11yTreeSnashots in debug environment because they not consistent on Puppeteer
```
DavideMininni-Fincons, line 29:
```suggestion
 * We skip a11yTreeSnashots in debug environment because they're not consistent on Puppeteer
```

src/components/core/testing/a11y-tree-snapshot.ts
```diff
@@ -22,10 +25,16 @@ async function a11yTreeEqualSnapshot(): Promise<void> {
...
  * @param title The title of the section
  * @param template The optional html template
  */
-export function testA11yTreeSnapshot(title = 'A11y tree', template?: TemplateResult): void {
+export function testA11yTreeSnapshot(
+  template?: TemplateResult,
+  title = 'A11y tree',
+  skip: { chrome?: boolean; firefox?: boolean; safari?: boolean } = {},
```
DavideMininni-Fincons, line 36: Question: the skip param is used only twice. Is it worth it, or can we manage the functionality somehow in the two specific spec files?

DavideMininni-Fincons (changes requested):

src/components/core/testing/a11y-tree-snapshot.ts
TomMenga: Since the `testA11yTreeSnapshot` function is seen as a black box from the components, we don't have a way to influence its behavior if not by using some params :/

src/components/core/testing/mocha-extensions.ts
```diff
@@ -0,0 +1,16 @@
+import type { Func, Suite, Test } from 'mocha';
+
+/**
+ * Skip the `describe` if the condition is not met
+ */
+export const describeIf = (
+  condition: boolean,
+  title: string,
+  fn: (this: Suite) => void,
+): Suite | void => (condition ? Mocha.describe(title, fn) : Mocha.describe.skip(title, fn));
```
MarioCastigliano, line 10: New approach looks great 👍

src/components/core/testing/a11y-tree-snapshot.ts
DavideMininni-Fincons: Got it 👍

MarioCastigliano (approved):
👍

PR #2401, opened 2024-02-08
feat(sbb-image): introduce support for round variant

jeripeierSBB:
BREAKING CHANGE: property `noBorderRadius` has been replaced by `borderRadius` which can receive 'default', 'none' and 'round'.


kyubisation (approved):
LGTM 👍

src/components/image/image.ts
```diff
@@ -269,7 +269,8 @@ export class SbbImageElement extends LitElement {
   /**
    * Whether to have no border-radius on the image
```
TomMenga, line 270: Just a nitpick, this doc needs to be updated

TomMenga (changes requested):

src/components/image/image.ts
jeripeierSBB: Thanks! fixed

TomMenga (approved):

PR #2406, opened 2024-02-12
fix(sbb-sticky-bar): remove unnecessary import to container

jeripeierSBB:


MarioCastigliano (approved):
Thanks for noticing 😅👍

PR #2415, opened 2024-02-13
chore: downgrade 'commitlint/config-conventional' to 18.6

TomMenga:
The last update caused the command not to work anymore. As a temporary fix, we downgraded its version to the last stable.

I expect a fix will arrive soon since someone already opened an [issue](https://github.com/conventional-changelog/commitlint/issues/3909) about it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   },
   "devDependencies": {
     "@commitlint/cli": "18.6.1",
-    "@commitlint/config-conventional": "18.6.0",
+    "@commitlint/config-conventional": "18.6.2",
     "@custom-elements-manifest/analyzer": "0.9.0",
     "@custom-elements-manifest/to-markdown": "0.1.0",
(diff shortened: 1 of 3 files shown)


jeripeierSBB:
Don't we need to temporary exclude it from renovate? I fear otherwise it gets automatically updated soonish again

TomMenga:
> Don't we need to temporary exclude it from renovate? I fear otherwise it gets automatically updated soonish again

Maybe yes... I don't know how renovate works :/
How do we exclude it?

jeripeierSBB:
with 18.6.2 it should be fixed. so we possibly can upgrade now

TomMenga:
> with 18.6.2 it should be fixed. so we possibly can upgrade now

Testing it now

TomMenga:
Renovate updated the `commitlint/config-conventional` to 18.6.2

PR #2446, opened 2024-02-29
fix(sbb-radio-group, sbb-tab-group): avoid incorrect setup if component is invisible during init

jeripeierSBB:


dauriamarco (approved):
Should the check be removed from the checkbox group as well? Otherwise it LGTM 👍

jeripeierSBB:
> Should the check be removed from the checkbox group as well? Otherwise it LGTM 👍

In the checkbox it's a different case because there we only check it whenever a key was pressed so it should be still ok. There it's also not the group which has to manage it's children state.

kyubisation (approved):
LGTM 👍

PR #2447, opened 2024-02-29
docs(multiple): remove undefined type from titleLevel

jeripeierSBB:


src/components/skiplink-list/skiplink-list.ts
```diff
@@ -59,7 +59,7 @@ export class SbbSkiplinkListElement extends SbbNamedSlotListElementMixin<
       <div class="sbb-skiplink-list__wrapper">
         <sbb-title
           class="sbb-link-list-title"
-          level=${this.titleLevel || nothing}
+          level=${this.titleLevel ?? nothing}
```
kyubisation, line 62: Empty string would not evaluate to nothing and be passed to the title. But maybe this should be fixed in the title itself?

kyubisation (approved):
LGTM 👍

PR #2452, opened 2024-03-01
refactor: implement and use hostAttributes decorator

kyubisation:
This replaces the usage of `createRenderRoot` for setting host attributes and instead uses the initializer functionality from lit to set static attributes.

### DIFF
--- a/src/components/core/common-behaviors/host-attributes.ts
+++ b/src/components/core/common-behaviors/host-attributes.ts
@@ -1,5 +1,7 @@
 import { isServer, type ReactiveElement } from 'lit';
 
+import type { AbstractConstructor } from './constructor';
+
 function applyAttributes(instance: ReactiveElement, attributes: Record<string, string>): void {
   for (const [name, value] of Object.entries(attributes)) {
     if (value) {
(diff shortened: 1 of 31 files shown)


jeripeierSBB (commented):
Very nice solution :)

jeripeierSBB:
I'd suggest to check every component (the ones which did not use createrenderroot so gar) as the migration effort is minimal.

src/components/core/common-behaviors/host-attributes.ts
```diff
@@ -0,0 +1,45 @@
...
+export const hostAttributes =
+  (attributes: Record<string, string>) => (target: typeof ReactiveElement) =>
+    target.addInitializer((instance: ReactiveElement) => {
+      if (isServer) {
+        applyAttributes(instance, attributes);
+      } else {
+        // It is not allowed to set attributes during constructor phase. Due to this we use a
+        // controller to assign the attributes during lit initialization phase.
+        instance.addController({
+          hostConnected() {
```
jeripeierSBB, line 39:
```suggestion
          hostUpdate() {
```

jeripeierSBB (approved):
Yay!

PR #2466, opened 2024-03-08
feat: implement experimental support for server side rendering (SSR)

kyubisation:
This PR adds the experimental possibility to use server side rendering with `@sbb-esta/lyne-components`. Most components have been verified to work with SSR, but our testing setup is not yet complete and will be finished in a future PR.


kyubisation:
Blocked until button refactoring is merged and this PR is rebased on top of it.

src/components/menu/menu/menu.ts
```diff
@@ -220,6 +220,10 @@ export class SbbMenuElement extends SbbNamedSlotListElementMixin<
     const signal = this._abort.signal;
     this.addEventListener('click', (e) => this._onClick(e), { signal });
     this.addEventListener('keydown', (e) => this._handleKeyDown(e), { signal });
+    this.shadowRoot?.addEventListener('slotchange', (e) => this._checkListCase(e), {
```
kyubisation, line 223:
```suggestion
    // Due to the fact that menu can both be a list and just a container, we need to check its
    // state before the SbbNamedSlotListMixin handles the slotchange event, in order to avoid
    // it interpreting the non list case as a list.
    this.shadowRoot?.addEventListener('slotchange', (e) => this._checkListCase(e), {
```

PR #2480, opened 2024-03-12
refactor(breakpoint): convert breakpoint-dependent code to pure css

### LINKED ISSUE #2339: feat: SSR rendering for breakpoint dependent code
Using breakpoint dependent code causes problems in SSR environments as there is no breakpoint available. E.g. on server, isBreakpoint() returns currently always false but on client side it depends on screen sizes. This causes errors during hydration.

Wherever calling isBreakpoint() has influence on initial rendering output, we want to replace its usages by a different solution. Most of the cases could be replaced by using just css.

**To be discussed in refinement:**
To provide a reusable help of using triggers, we have the idea of creating some global vars which could be used.
 `--sbb-display-none-from-{breakpoint}`
 `--sbb-display-none-to-{breakpoint}`
 `--sbb-breakpoint-multiplier-from-{breakpoint}`
 `--sbb-breakpoint-multiplier-to-{breakpoint}`

 `--sbb-breakpoint-multiplier-to-{breakpoint}` could be use together with calc like `padding:inline: calc(var(--sbb-spacing-fixed-5x) * var(--sbb-breakpoint-multiplier-from-{breakpoint}))`

```css
display: var(--sbb-display-none-to-small, var(--sbb-display-none-from-large, 'flex'));

padding: calc(var(--base-padding) + (12px * var(--sbb-breakpoint-multiplier-from-small)));
```

**This should not replace media queries entirely, but just an extension to it.**

TomMenga:
This PR closes #2339 

Some components still use `isBreakpoint` in their code. Specifically:
- `sbb-calendar`: it's too complex to convert to a pure CSS approach (we delayed breakpoint-dependent code after hydration)
- `sbb-menu`: use `isBreakpoint` when opened (so no conflict with ssr)
- `navigation-section`: partially converted to CSS approach. The remaining part is not executed on init

### DIFF
--- a/src/components/calendar/calendar.ts
+++ b/src/components/calendar/calendar.ts
@@ -1,5 +1,5 @@
 import type { CSSResultGroup, TemplateResult, PropertyValues } from 'lit';
-import { html, LitElement, nothing } from 'lit';
+import { html, isServer, LitElement, nothing } from 'lit';
 import { customElement, property, state } from 'lit/decorators.js';
 import { classMap } from 'lit/directives/class-map.js';
 
@@ -133,7 +133,7 @@ export class SbbCalendarElement extends LitElement {
   private _weekdays!: Weekday[];
 
   /** Grid of calendar cells representing the dates of the month. */
-  private _weeks!: Day[][];
+  private _weeks: Day[][] = [];
 
   /** Grid of calendar cells representing months. */
   private _months!: Month[][];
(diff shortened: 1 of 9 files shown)


src/components/calendar/calendar.ts
```diff
@@ -1208,6 +1231,7 @@ export class SbbCalendarElement extends LitElement {
   }
 
   protected override render(): TemplateResult {
+    console.log('render');
```
DavideMininni-Fincons, line 1234: Remember to clean this before merge :D

src/components/calendar/calendar.ts
```diff
@@ -253,6 +270,11 @@ export class SbbCalendarElement extends LitElement {
 
   /** Initializes the component. */
   private _init(activeDate?: Date): void {
+    //Due to its complexity, the caledar is inited only on client side
```
kyubisation, line 273:
```suggestion
    // Due to its complexity, the calendar is only initialized on client side.
```

kyubisation (approved):
LGTM 👍 
(Apart from Davides comment, of course 😄 )

src/components/calendar/calendar.ts
TomMenga: aaaaaah

PR #2481, opened 2024-03-12
fix(sbb-navigation): fix navigation actions contrast ratio

dauriamarco:
- Adds a dash icon next to the active navigation actions in the section. 
- Fixes the active and reset mechanism of the markers no open/close.
- Disables animation on resize.

### DIFF
--- a/src/components/navigation/navigation/navigation.ts
+++ b/src/components/navigation/navigation/navigation.ts
@@ -270,9 +270,11 @@ export class SbbNavigationElement extends UpdateScheduler(LitElement) {
 
   private _resetMarkers(): void {
     const activeActions = Array.from(
-      this.querySelectorAll('sbb-navigation-button[data-action-active]:not(.sbb-active)'),
+      this.querySelectorAll(
+        ':is(sbb-navigation-button, sbb-navigation-link)[data-action-active]:not(.sbb-active)',
+      ),
     ) as (SbbNavigationButtonElement | SbbNavigationLinkElement)[];
-    activeActions?.forEach((action) => action.toggleAttribute('data-action-active', false));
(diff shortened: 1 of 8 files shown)


src/components/navigation/navigation/navigation.ts
```diff
@@ -169,6 +172,14 @@ export class SbbNavigationElement extends UpdateScheduler(LitElement) {
     activeAction?.connectedSection?.open();
   }
 
+  private _checkActiveActions(): void {
+    const activeActions = Array.from(this.querySelectorAll('sbb-navigation-button.sbb-active')) as (
```
jeripeierSBB, line 176: otherwise create a @hostAttributes on mixin like with buttons and create a data-navigation-
```suggestion
    const activeActions = Array.from(this.querySelectorAll(':is(sbb-navigation-button, sbb-navigation-link).sbb-active')) as (
```

src/components/navigation/navigation/navigation.ts
```diff
@@ -332,6 +345,24 @@ export class SbbNavigationElement extends UpdateScheduler(LitElement) {
...
+    if (this._resizeObserverTimeout) {
+      clearTimeout(this._resizeObserverTimeout);
+    }
+
+    this.toggleAttribute('data-resize-disable-animation', true);
+
+    // Disable the animation when resizing the navigation to avoid strange height transition effects.
+    this._resizeObserverTimeout = setTimeout(
+      () => this.toggleAttribute('data-resize-disable-animation', false),
+      150,
```
jeripeierSBB, line 362: maybe a constant?

jeripeierSBB (approved):
LGTM, thanks for all the fixes!

PR #2488, opened 2024-03-13
refactor: migrate 'toggleDatasetEntry' calls to 'toggleAttribute'

TomMenga:
Removed `toggleDatasetEntry` usages in favor of `element.toggleAttribute`


jeripeierSBB (approved):
You won! :-) Thanks! Not sure if we should keep the migration script?

TomMenga:
> You won! :-) Thanks! Not sure if we should keep the migration script?

The idea was to keep it for one/two weeks top just in case someone needs it.
Most likely, nobody will use it

jeripeierSBB:
> > You won! :-) Thanks! Not sure if we should keep the migration script?
> 
> The idea was to keep it for one/two weeks top just in case someone needs it. Most likely, nobody will use it though. Also, it might be useful to have it in the history for future reference (as an example of when (if) we'll implement migration scripts for breaking changes)

it's ok :)

PR #2491, opened 2024-03-14
feat(sbb-stepper): introduce `sbb-stepper` component

### LINKED ISSUE #2380: story(stepper): initial implementation
### Feature Description

We want to implement the `sbb-stepper` component, to be used in workflows.

[Figma Spec](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?type=design&node-id=140%3A17&mode=design&t=FXB5x6m9UoyAmzbh-1)

Examples/References:
https://angular.app.sbb.ch/angular/components/stepper/overview
https://material.angular.io/components/stepper/overview

**Design Spec**

- A stepper can be horizontal or vertical
- A stepper step `<sbb-step-label>` can have an icon. If none is specified, the step number should be shown as an icon fallback.
- The underline of the step label should extend to the currently active label. This also applies to the vertical case, but the line stops below the label and not below the content.
- @mcilurzo will provide

**Technical Spec**

`<sbb-stepper>`

- Acts as the container for steps
- Should have role tablist
- Should have property `orientation: 'horizontal' | 'vertical' = 'horizontal'` and related `horizontalFrom` and should set `aria-orientation` depending on direction
- Should have methods `next` and `previous` that invoke the workflows for switching steps
- Should have a method `reset`, which resets the stepper (This should also reset all `<form>` instances inside it.)
- Should have a getter `steps` for all `<sbb-step>` instances
- Should have `selected` getter/setter, which returns the current `<sbb-step>` or undefined, if empty, and if used as a setter (with an `<sbb-step>` element) it will activate the given step
  - This should also trigger step change workflow
- Should have `selectedIndex` getter/setter, which returns the index of the currently selected step and if called with as a setter, activates the corresponding step
  - This should also trigger step change workflow
- Should NOT use the `SbbNamedSlotListMixin`, but instead have two slots. One for labels and one for step content.
- Should have a boolean property `linear`
  - If `linear` is not set, every label can be clicked to activate it
  - If `linear` is set, only the current and previous labels can be clicked to active them

`<sbb-step-label>`

- Should extend the base button element
- Should have role tab
- Should have property `iconName` (use the mixin)
- Should link to the related content with appropriate aria attributes (see sbb-angular or material examples)
- A click on a label should active its step, unless its a future step in linear mode
  - This should also trigger step change workflow
- Should set `aria-posinset="${current index + 1}"` (e.g. the first has `1`), `aria-setsize=${amount of steps}` and `aria-selected=${is currently active}`

`<sbb-step>`

- Should link to the related label with appropriate aria attributes (see sbb-angular or material examples)
- Should have role tabpanel
- Should have a getter for its label
- Should have a click event handler
  - If the target of the event has an `sbb-stepper-next` attribute, the workflow for switching to the next step should be invoked.
  - If the target of the event has an `sbb-stepper-previous` attribute, the workflow for switching to the previous step should be invoked.

**Workflows**

- When a step switch is triggered, a custom event `validate` (payload `{ currentIndex: number, currentStep: SbbStepElement, nextIndex: number, nextStep: SbbStepElement }`) is emitted from the current `<sbb-step>`. If it is aborted, the step switch is also aborted.

### Examples

```html
<sbb-stepper aria-label="Purpose of this flow">
  <sbb-step-label>Step 1</sbb-step-label>
  <sbb-step>
    ...
    <sbb-secondary-button sbb-stepper-previous>Go back</sbb-secondary-button>
    <sbb-button sbb-stepper-next>Go to next</sbb-button>
  </sbb-step>
  <sbb-step-label>Step 2</sbb-step-label>
  <sbb-step>
    ...
  </sbb-step>
</sbb-stepper>
```

```html
<sbb-stepper aria-label="Purpose of this flow">
  <sbb-step-label icon-name="pie-small">Step 1</sbb-step-label>
  <sbb-step>
    ...
  </sbb-step>
  <sbb-step-label>Step 2</sbb-step-label>
  <sbb-step>
    ...
  </sbb-step>
</sbb-stepper>
```

### Definition of Done

- [x] Documentation is provided
- [x] Tests are implemented
- [x] E2E Tests are implemented
- [x] Storybook stories are implemented
- [x] Navigation via keyboard is tested
- [x] Screen reader output is tested
- [x] High-contrast is tested
- [x] Remaining accessibility is tested
- [x] UX approved

dauriamarco:
## Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/main/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

## Issue

This PR Closes #2380

## Pull request checklist

## Changes

Changes in this pull request:

- introduces `sbb-stepper`, `sbb-step-label` and `sbb-step` components.

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/stepper/step-label/step-label.ts
```diff
@@ -0,0 +1,88 @@
...
+  slot: 'step-label',
+  role: 'tab',
+})
+export class SbbStepLabelElement extends SbbIconNameMixin(
+  SbbDisabledTabIndexActionMixin(SbbButtonBaseElement),
+) {
+  public static override styles: CSSResultGroup = style;
+
+  /** @internal */
+  public readonly internals: ElementInternals = this.attachInternals();
```
jeripeierSBB, line 35: Could it be private?

src/components/stepper/step-label/step-label.ts
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Combined with a `sbb-stepper`, it displays a step's label.
+ *
+ * @slot - Use the unnamed slot to provide a label.
+ */
+@customElement('sbb-step-label')
+@hostAttributes({
+  slot: 'step-label',
+  role: 'tab',
```
jeripeierSBB, line 27: could role be part of element internals? (at least in checkbox it worked)

jeripeierSBB (commented):
Not really a review, just something missing which was in my mind: please add sbb-stepper to typography.scss where we have the lists of components to hide until defined ("// Hiding components until they are instantiated")

src/components/stepper/step-label/step-label.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { expect, fixture } from '@open-wc/testing';
```
DavideMininni-Fincons, line 1: - Add inner text in fixture
- Use equalSnapshot
- Add a case with icon and one with disabled
- Add the a11y test

src/components/stepper/step-label/step-label.stories.ts
```diff
@@ -0,0 +1,70 @@
...
+const Template = (args: Args): TemplateResult =>
+  html`<sbb-step-label ${sbbSpread(args)}>Label</sbb-step-label>`;
+
+export const Default: StoryObj = {
+  render: Template,
+  argTypes: defaultArgTypes,
+  args: { ...defaultArgs },
+};
+
+export const Selected: StoryObj = {
```
DavideMininni-Fincons, line 42: Add a story with icon-name set

src/components/stepper/step-label/step-label.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  private _abort = new ConnectedAbortController(this);
+  private _stepper: SbbStepperElement | null = null;
+  private _step: SbbStepElement | null = null;
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    const signal = this._abort.signal;
+    this.id = this.id || `sbb-step-label-${nextId++}`;
+    this._stepper = this.closest('sbb-stepper');
```
DavideMininni-Fincons, line 52: Should we use `closest?.(` for SSR?

src/components/stepper/step/step.spec.ts
```diff
@@ -0,0 +1,21 @@
+import { expect, fixture } from '@open-wc/testing';
```
DavideMininni-Fincons, line 1: See step-label.spec.ts comments

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,133 @@
...
+  private _abort = new ConnectedAbortController(this);
+  private _stepper: SbbStepperElement | null = null;
+  private _label: SbbStepLabelElement | null = null;
+  private _stepResizeObserver = new AgnosticResizeObserver((entries) =>
+    this._onStepElementResize(entries),
+  );
+
+  /**
+   * The label of the step.
+   */
```
DavideMininni-Fincons, line 56:
```suggestion
  /** The label of the step. */
```

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,133 @@
...
+      this._stepper?.style?.setProperty('--sbb-stepper-content-height', `${contentHeight}px`);
+    }
+  }
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    const signal = this._abort.signal;
+    this.id = this.id || `sbb-step-${nextId++}`;
+    this.addEventListener('click', (e) => this._handleClick(e), { signal });
+    this._stepper = this.closest('sbb-stepper');
```
DavideMininni-Fincons, line 99: Same comment on `closest` as in step-label.ts

src/components/stepper/stepper/readme.md
```diff
@@ -0,0 +1,97 @@
...
+</sbb-stepper>
+```
+
+## Interactions
+
+There are two attributes to support navigation between different steps that can be used on elements inside an `sbb-step` to select the next or the previous step when clicked: `sbb-stepper-next` and `sbb-stepper-previous`.
+
+### Linear stepper
+
+The `linear` property can be set to create a linear stepper that requires the user to complete previous steps before proceeding to following steps.
```
DavideMininni-Fincons, line 21: Maybe add an example?

src/components/stepper/stepper/stepper.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+        <sbb-step>Test step content 2</sbb-step>
+        <sbb-step-label disabled>Test step label 3</sbb-step-label>
+        <sbb-step>Test step content 3</sbb-step>
+        <sbb-step-label>Test step label 4</sbb-step-label>
+      </sbb-stepper>
+    `);
+  });
+
+  it('renders', async () => {
+    const root = await fixture(html`<sbb-stepper></sbb-stepper>`);
```
DavideMininni-Fincons, line 25: Since there's the beforeEach code, is this line needed?

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,133 @@
...
+import { ConnectedAbortController, EventEmitter } from '../../core/eventing';
+import { AgnosticResizeObserver } from '../../core/observers';
+import type { SbbStepLabelElement } from '../step-label';
+import type { SbbStepperElement } from '../stepper';
+
+import style from './step.scss?lit&inline';
+
+let nextId = 0;
+
+type ValidateEventDetails = {
```
DavideMininni-Fincons, line 15: Should this be exported? just to be sure about the typing, eg. in the stepper..

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,277 @@
+import {
```
DavideMininni-Fincons, line 1: General comment: inline method's comments when possible.
```
/** This is a method. */
public myMethod(): void {
}
```

DavideMininni-Fincons (changes requested):
Great work! minor details to fix

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,277 @@
...
+    });
+  }
+
+  private _checkOrientation(): void {
+    if (this.horizontalFrom) {
+      this.orientation = isBreakpoint(this.horizontalFrom) ? 'horizontal' : 'vertical';
+      this._updateLabels();
+    }
+    setTimeout(() => this._setMarkerSize(), 0);
+  }
```
TomMenga, line 214: Breakpoint-dependent code during initialization is a problem for SSR (where you do not have breakpoints) because it might cause a DOM mismatch.

We should have a call where I share with you the problem and see if we can find a solution

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  private _onStepperResize(): void {
+    this._checkOrientation();
+
+    if (this._resizeObserverTimeout) {
+      clearTimeout(this._resizeObserverTimeout);
+    }
```
TomMenga, line 221:
```suggestion
    clearTimeout(this._resizeObserverTimeout);
```
I think `clearTimeout` is null safe

TomMenga (changes requested):
Nice work
Just a partial review (mainly focusing on ts logic)

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,277 @@
...
+      currentIndex: this.selectedIndex,
+      currentStep: this.selected,
+      nextIndex: this.selectedIndex !== undefined ? this.selectedIndex + 1 : undefined,
+      nextStep: this.selectedIndex !== undefined ? this.steps[this.selectedIndex + 1] : undefined,
+    };
+    if (this.selected && !this.selected.validate(validatePayload)) {
+      return;
+    }
+    const current = this.selected;
+    if (current && current.label) {
```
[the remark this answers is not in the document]
dauriamarco: No, a step should alway have a label.

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,277 @@
...
+    ) {
+      return;
+    }
+    const offset =
+      this.orientation === 'horizontal'
+        ? this.selected.label.offsetLeft + this.selected.label.offsetWidth
+        : this.selected.label.offsetHeight * (this.selectedIndex + 1) +
+          parseFloat(getComputedStyle(this).getPropertyValue('--sbb-spacing-responsive-m')) *
+            16 *
+            this.selectedIndex;
```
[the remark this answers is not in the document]
dauriamarco: True, initially we didn't have the possibility to have multiline labels and I forgot to adapt the calculations when I added them. Thanks 👍

src/components/stepper/step-label/step-label.stories.ts
dauriamarco: Actually all the stories have an icon-name set, also because otherwise the circle would be empty as the css counter is handled by the stepper component.

src/components/stepper/step-label/step-label.ts
dauriamarco: Idk if I can use the optional chaining like that, anyway the `closest()` method should be null safe.

src/components/stepper/step/step.ts
dauriamarco: Idk if I can use the optional chaining like that, anyway the `closest()` method should be null safe.

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,277 @@
...
+  private _configure(): void {
+    const labels = this.steps.map((s) => s.label);
+    let posInSet = 0;
+    labels.forEach((label) => {
+      if (label) {
+        label.internals.ariaSelected = (this.selected?.label === label).toString();
+        label.internals.ariaPosInSet = `${++posInSet}`;
+        label.internals.ariaSetSize = `${labels.length}`;
+      }
+    });
```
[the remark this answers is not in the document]
dauriamarco: I added the config method inside the `sbb-step-label`. And what do you mean by "an inconsistent state where only aria-selected is set"?

src/components/stepper/stepper/stepper.ts
dauriamarco: As we discussed, with SSR we might be safe because we are acting on the host elements themselves and not changing the rest of the DOM.

dauriamarco: I would also wait for @kyubisation and @jeripeierSBB opinion on this :)

src/components/stepper/step-label/step-label.ts
dauriamarco: I tried it and, although I see the role correctly set in the internals, the screen-reader doesn't seem to read them correctly 😕

jeripeierSBB: ok that's a bad sign. hmm which browser screenreader combination was affected?

dauriamarco: VoiceOver with all the browsers, I didn't test all the others.

src/components/stepper/step/step.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+    const root = await fixture(html`<sbb-step>Step content</sbb-step>`);
+
+    expect(root).dom.to.be.equal(
+      `<sbb-step id="sbb-step-0" role="tabpanel" slot="step">Step content</sbb-step>`,
+    );
+
+    await expect(root).shadowDom.to.be.equalSnapshot();
+  });
+
+  testA11yTreeSnapshot(html`<sbb-step-label>Label</sbb-step-label>`);
```
TomMenga, line 19: Probably a copypaste mistake :(

ps.
see `accordion.spec.ts` approach. You can use the same template to take dom, shadowDom, and a11y tree snapshots

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,278 @@
...
+      closestForm.reset();
+    } else {
+      this.querySelectorAll('form').forEach((form) => form.reset());
+    }
+    this.selectedIndex = 0;
+  }
+
+  private _loaded: boolean = false;
+  private _abort = new ConnectedAbortController(this);
+  private _resizeObserverTimeout: ReturnType<typeof setTimeout> | null = null;
```
TomMenga, line 116: I would move this before the `next` method but maybe we should discuss this in with team

TomMenga (commented):
Good job, just a couple of small changes and we're done 😄

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+    this.toggleAttribute('data-selected', false);
+    this.label.deselect();
+  }
+
+  public validate(eventData: SbbStepValidateEventDetails): boolean {
+    return !!this._validate.emit(eventData);
+  }
+
+  private _handleClick(event: Event): void {
```
jeripeierSBB, line 89: or similar..
```suggestion
 /**
 * Watch content for clicks with sbb-stepper-next or sbb-stepper-previous attributes
 */
  private _handleClick(event: Event): void {
```

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  private _isGoNextElement(element: HTMLElement): boolean {
+    return element.hasAttribute('sbb-stepper-next') && !isValidAttribute(element, 'disabled');
+  }
+
+  private _isGoPreviousElement(element: HTMLElement): boolean {
+    return element.hasAttribute('sbb-stepper-previous') && !isValidAttribute(element, 'disabled');
+  }
+
+  private _onStepElementResize(entries: ResizeObserverEntry[]): void {
```
jeripeierSBB, line 108: does a slotchange (changed content) trigger a resize too? otherwise we would need to respect the slotchange event as well I think

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (!this.hasAttribute('data-selected')) {
+      return;
+    }
+    for (const entry of entries) {
```
jeripeierSBB, line 112: do we really need a loop?

src/components/stepper/step/step.ts
```diff
@@ -0,0 +1,157 @@
...
+    this._stepper = this.closest('sbb-stepper');
+    if (this.previousElementSibling?.tagName === 'SBB-STEP-LABEL') {
+      this._label = this.previousElementSibling as SbbStepLabelElement;
+    }
+  }
+
+  protected override firstUpdated(): void {
+    this._stepResizeObserver.observe(this.shadowRoot!.querySelector('.sbb-step')!);
+    if (this.label) {
+      this.setAttribute('aria-labelledby', this.label.id);
```
jeripeierSBB, line 132: this should also be done whenever this._label is set. Ifthis step gets a new dom connected with a possible different label, it won't update the aria labelledby

src/components/stepper/stepper/stepper.scss
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  :host([orientation='horizontal']) & {
+    gap: var(--sbb-spacing-fixed-4x);
+    padding-block-end: var(--sbb-spacing-fixed-4x);
+    border-block-end: var(--sbb-border-width-1x) solid var(--sbb-color-cloud);
+
+    &::before {
+      inset-block-end: 0;
+      height: #{sbb.px-to-rem-build(3)};
```
jeripeierSBB, line 54: design token sbb-border-width-3x needed maybe? @mcilurzo

src/components/stepper/stepper/stepper.scss
```diff
@@ -0,0 +1,134 @@
...
+  position: relative;
+
+  :host([orientation='horizontal']) & {
+    height: var(--sbb-stepper-content-height);
+    transition: height var(--sbb-stepper-animation-duration) var(--sbb-animation-easing);
+  }
+}
+
+::slotted(sbb-step-label) {
+  @include sbb.text-xxs--regular;
```
jeripeierSBB, line 83: I wonder if text styles would better fit into sbb-step-label itslef?

src/components/stepper/stepper/stepper.scss
```diff
@@ -0,0 +1,134 @@
...
+
+  position: relative;
+  min-width: 0;
+  max-width: fit-content;
+
+  :host([orientation='vertical']) & {
+    transition: margin var(--sbb-stepper-animation-duration) var(--sbb-animation-easing);
+  }
+
+  &::before {
```
jeripeierSBB, line 93: I think some of the props should be moved into step-label. especially if using variables from there.

src/components/stepper/stepper/stepper.scss
```diff
@@ -0,0 +1,134 @@
...
+    counter-increment: step-label;
+    cursor: pointer;
+    position: absolute;
+    color: var(--sbb-step-label-color);
+    inset-block-start: calc(
+      var(--sbb-font-size-text-l) * (var(--sbb-typo-line-height-body-text) / 2) +
+        (var(--sbb-border-width-1x) / 2)
+    );
+    inset-inline-start: calc(var(--sbb-step-label-prefix-size) / 2);
+    transform: translate(-50%, -50%);
```
jeripeierSBB, line 104: use SASS mixin?

src/components/stepper/stepper/stepper.stories.ts
```diff
@@ -0,0 +1,437 @@
...
+  return html`
+    <form
+      @submit=${(e: SubmitEvent) => {
+        e.preventDefault();
+        const formData = new FormData(e.target as HTMLFormElement);
+        for (const [name, value] of formData) {
+          document.querySelector(`.text-block-${name}`)!.textContent = value.toString();
+        }
+      }}
+      @reset=${() => {
```
jeripeierSBB, line 90: Maybe document here why this is needed?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    const signal = this._abort.signal;
+    window.addEventListener('resize', () => this._onStepperResize(), {
+      signal,
+      passive: true,
+    });
+    this.toggleAttribute('data-disable-animation', true);
```
jeripeierSBB, line 237: if moving around in DOM, would it then not block animation forever?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,278 @@
...
+      parseFloat(getComputedStyle(this).getPropertyValue('--sbb-spacing-responsive-m')) *
+        16 *
+        this.selectedIndex
+    );
+  }
+
+  private _configure(): void {
+    const labels = this.steps.filter((s) => s.label).map((s) => s.label!);
+
+    labels.forEach((label, i) => {
```
jeripeierSBB, line 197: why not just chain with previous statements?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,278 @@
...
+      step.label?.setAttribute('data-orientation', this.orientation);
+    });
+  }
+
+  private _checkOrientation(): void {
+    if (this.horizontalFrom) {
+      this.orientation = isBreakpoint(this.horizontalFrom) ? 'horizontal' : 'vertical';
+      this._updateLabels();
+    }
+    setTimeout(() => this._setMarkerSize(), 0);
```
jeripeierSBB, line 215: Maybe a short comment why this is needed?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,278 @@
...
+    return this.querySelector<SbbStepElement>('sbb-step[data-selected]') ?? undefined;
+  }
+
+  public set selected(step: SbbStepElement) {
+    this._select(step);
+  }
+
+  /**
+   * The currently selected step index.
+   */
```
jeripeierSBB, line 68: should it not have a property annotation? or shouldn't it work when setting as attribtue?

jeripeierSBB, line 68: should it not have a property annotation? or shouldn't it work when setting as attribute?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,278 @@
...
+    window.addEventListener('resize', () => this._onStepperResize(), {
+      signal,
+      passive: true,
+    });
+    this.toggleAttribute('data-disable-animation', true);
+  }
+
+  protected override async firstUpdated(): Promise<void> {
+    await this.updateComplete;
+    this.selectedIndex = !this.linear ? Number(this.getAttribute('selected-index')) || 0 : 0;
```
jeripeierSBB, line 242: in readme, selectedIndex is not visible as attribute

jeripeierSBB (changes requested):
Looks very nice! There are some issues and questions.
- In linear mode, steps which I can't click are still shown as clickable (cursor and hover). Shouldn't that be prevented?
- I'm generally not sure how robust the construction is whenever you move some elements in the DOM. Or as well as for initialization order, does it always recover to a stable state?
- General question: Did you think the SSR cases? From fast overlook, it should be ok.
- We should ask Mürsel for a review

src/components/stepper/step/step.ts
dauriamarco: yes, if I change the content the step will be resized accordingly.

src/components/stepper/step-label/step-label.stories.ts
```diff
@@ -0,0 +1,79 @@
...
+  argTypes: defaultArgTypes,
+  args: { ...defaultArgs, 'icon-name': undefined },
+};
+
+const meta: Meta = {
+  decorators: [withActions as Decorator],
+  parameters: {
+    backgrounds: {
+      disable: true,
+    },
```
jeripeierSBB, line 71: not needed anymore
```suggestion
```

src/components/stepper/step/step.stories.ts
```diff
@@ -0,0 +1,28 @@
...
+export const Default: StoryObj = {
+  render: Template,
+};
+
+const meta: Meta = {
+  decorators: [withActions as Decorator],
+  parameters: {
+    backgrounds: {
+      disable: true,
+    },
```
jeripeierSBB, line 20: not needed anymore
```suggestion
```

src/components/stepper/stepper/stepper.stories.ts
```diff
@@ -0,0 +1,457 @@
...
+
+const meta: Meta = {
+  decorators: [withActions as Decorator],
+  parameters: {
+    actions: {
+      handles: [SbbStepElement.events.validate],
+    },
+    backgrounds: {
+      disable: true,
+    },
```
jeripeierSBB, line 449: not needed anymore
```suggestion
```

src/components/stepper/step-label/step-label.ts
```diff
@@ -0,0 +1,116 @@
...
+  protected override firstUpdated(): void {
+    if (this.step) {
+      this.setAttribute('aria-controls', this.step.id);
+    }
+  }
+
+  protected override render(): TemplateResult {
+    return html`
+      <div class="sbb-step-label">
+        <span class="sbb-step-label__prefix">${this.renderIconSlot()}</span>
```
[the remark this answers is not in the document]
dauriamarco: Understood, btw here there should be no problem with the icon as it has no gap and I would need the `span` so I think I can keep it 👍

jeripeierSBB (commented):
Only partial like requested

src/components/stepper/step-label/step-label.scss
```diff
@@ -0,0 +1,143 @@
...
+  --sbb-step-label-color: var(--sbb-color-granite);
+  --sbb-step-label-prefix-border-style: dashed;
+}
+
+:host(:hover:not([disabled])) {
+  @include sbb.hover-mq($hover: true) {
+    --sbb-step-label-cursor: pointer;
+    --sbb-step-label-prefix-background-color: var(--sbb-color-milk);
+    --sbb-step-label-translate-y-content-hover: #{sbb.px-to-rem-build(-1)};
+    --sbb-step-label-prefix-size-grow-hover: calc(var(--sbb-border-width-2x) * -1);
```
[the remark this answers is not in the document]
dauriamarco: In other components like the `header-action` and the `tag` we also use `calc(var(--sbb-border-width-2x) * -1)` for the same thing. Would you change it there as well or maintain consistency?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,321 @@
...
+ * @slot step - Use this slot to provide an `sbb-step`.
+ */
+@customElement('sbb-stepper')
+export class SbbStepperElement extends LitElement {
+  public static override styles: CSSResultGroup = style;
+
+  /**
+   * If set to true, only the current and previous labels can be clicked and selected.
+   */
+  @property({ type: Boolean }) public linear = false;
```
jeripeierSBB, line 33: linear mode missing in tests

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  /**
+   * The steps of the stepper.
+   */
+  public get steps(): SbbStepElement[] {
+    return Array.from(this.querySelectorAll('sbb-step'));
+  }
+
+  private get _enabledSteps(): SbbStepElement[] {
+    return this.steps.filter((s) => !s.label?.hasAttribute('disabled'));
```
jeripeierSBB, line 90: should here also data-disabled be considered?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,321 @@
...
+  }
+
+  /**
+   * Selects the next step.
+   */
+  public next(): void {
+    if (this.selectedIndex !== undefined) {
+      this._select(this.steps[this.selectedIndex + 1]);
```
jeripeierSBB, line 98: what happens if running out of bounds?

src/components/stepper/step-label/step-label.scss
```diff
@@ -0,0 +1,151 @@
+@use '../../core/styles' as sbb;
+
+// Box-sizing rules contained in typography are not traversing Shadow DOM boundaries. We need to include box-sizing mixin in every component.
+@include sbb.box-sizing;
+
+:host {
+  --sbb-step-label-color: var(--sbb-color-iron);
+  --sbb-step-label-animation-duration: var(--sbb-animation-duration-2x);
```
jeripeierSBB, line 8: you need to include the new disable animation mechanism (see other components)

src/components/stepper/step/step.scss
```diff
@@ -0,0 +1,68 @@
...
+// Box-sizing rules contained in typography are not traversing Shadow DOM boundaries. We need to include box-sizing mixin in every component.
+@include sbb.box-sizing;
+
+:host {
+  --sbb-step-position: initial;
+  --sbb-step-inset-block-start: unset;
+  --sbb-step-opacity: 0;
+  --sbb-step-visibility: hidden;
+  --sbb-step-height: 0;
+  --sbb-step-animation-duration: var(--sbb-animation-duration-2x);
```
jeripeierSBB, line 12: you need to include the new disable animation mechanism (see other components)

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,321 @@
...
+    }
+    this.selectedIndex = 0;
+  }
+
+  private _loaded: boolean = false;
+  private _abort = new SbbConnectedAbortController(this);
+  private _resizeObserverTimeout: ReturnType<typeof setTimeout> | null = null;
+
+  private _isValidStep(step: SbbStepElement): boolean {
+    if (!step || step === this.selected || (!this.linear && step.label?.hasAttribute('disabled'))) {
```
jeripeierSBB, line 129: does data-disabled need to be considered?

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,321 @@
...
+  }
+  public get selectedIndex(): number | undefined {
+    return this.selected ? this.steps.indexOf(this.selected) : undefined;
+  }
+
+  /**
+   * The steps of the stepper.
+   */
+  public get steps(): SbbStepElement[] {
+    return Array.from(this.querySelectorAll('sbb-step'));
```
jeripeierSBB, line 86: steps gets called on SSR, where querySelectorAll is not defined
```suggestion
    return Array.from(this.querySelectorAll?.('sbb-step') ?? []);
```

src/components/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,321 @@
...
+   * The currently selected step.
+   */
+  @property({ attribute: false })
+  public set selected(step: SbbStepElement) {
+    if (this._loaded) {
+      this._select(step);
+    }
+  }
+  public get selected(): SbbStepElement | undefined {
+    return this.querySelector<SbbStepElement>('sbb-step[data-selected]') ?? undefined;
```
jeripeierSBB, line 66:
```suggestion
    return this.querySelector<SbbStepElement>.?('sbb-step[data-selected]') ?? undefined;
```

jeripeierSBB (changes requested):
Almost done :P

src/components/stepper/stepper/stepper.e2e.ts
```diff
@@ -0,0 +1,245 @@
...
+    await waitForCondition(() => validate.events.length === 1);
+    expect(validate.count).to.be.equal(1);
+    expect(stepLabelThree).not.to.have.attribute('data-selected');
+    expect(stepLabelThree.step).not.to.have.attribute('data-selected');
+  });
+
+  it('resets the single form wrapping the stepper and returns to the first step', async () => {
+    element = await fixture(html`
+      <form>
+        <sbb-stepper selected-index="0">
```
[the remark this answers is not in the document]
dauriamarco: If I remove the slot assignments the linter complains...

src/components/stepper/stepper/stepper.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+import '../step.js';
+import '../step-label.js';
+
+describe('sbb-stepper', () => {
+  let element: SbbStepperElement;
+
+  beforeEach(async () => {
+    element = await fixture(html`
+      <sbb-stepper selected-index="0">
+        <sbb-step-label slot="step-label">Test step label 1</sbb-step-label>
```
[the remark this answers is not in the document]
dauriamarco: same as for the e2e, the linter would complain

src/components/stepper/step-label/step-label.scss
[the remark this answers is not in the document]
dauriamarco: maybe, but it would be `spacing-fixed-1x / 2` because I need 2px

src/components/stepper/stepper/stepper.ts
dauriamarco: I don't think so, the data attr would always be in combination with `disabled`

src/components/stepper/stepper/stepper.ts
dauriamarco: I don't think so, the data attr would always be in combination with `disabled`

jeripeierSBB: ok, later I thought it would be like that :)

src/components/stepper/stepper/stepper.ts
jeripeierSBB: agree :)

src/components/stepper/step-label/step-label.scss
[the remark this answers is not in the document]
jeripeierSBB: oh ok, my comment suggestion above is wrong ;). yeah then I'm still not sure what is the best :D

src/components/stepper/stepper/stepper.ts
dauriamarco: simply nothing happens

src/components/stepper/stepper/stepper.e2e.ts
```diff
@@ -242,4 +282,88 @@ describe('sbb-stepper', () => {
     expect(stepLabelOne).to.have.attribute('data-selected');
     expect(stepLabelOne.step).to.have.attribute('data-selected');
   });
+
+  it('focuses the correct element in the step cntent', async () => {
```
jeripeierSBB, line 286:
```suggestion
  it('focuses the correct element in the step content', async () => {
```

src/components/stepper/stepper/stepper.e2e.ts
```diff
@@ -242,4 +282,88 @@ describe('sbb-stepper', () => {
...
+    const stepLabelThree = element.querySelector<SbbStepLabelElement>(
+      'sbb-step-label:nth-of-type(3)',
+    )!;
+
+    await sendKeys({ down: 'Tab' });
+    await sendKeys({ down: 'ArrowLeft' });
+
+    expect(stepLabelThree).to.have.attribute('data-selected');
+    expect(stepLabelThree.step).to.have.attribute('data-selected');
+  });
```
jeripeierSBB, line 368: Nice tests which you added. thank you!

jeripeierSBB: Last test question, should we also test / assert the labelled by link from step to step-label?

src/components/stepper/step-label/step-label.scss
```diff
@@ -0,0 +1,157 @@
...
+    transform: translate(
+      -50%,
+      calc(-50% + var(--sbb-step-label-translate-y-content-hover, #{sbb.px-to-rem-build(0)}))
+    );
+    transition: transform var(--sbb-step-label-animation-duration) var(--sbb-animation-easing);
+  }
+
+  @include sbb.mq($from: medium) {
+    --sbb-step-label-prefix-size: #{sbb.px-to-rem-build(36)};
+  }
```
jeripeierSBB, line 46: Already included in size token
```suggestion
```

src/components/stepper/step-label/step-label.scss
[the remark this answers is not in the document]
jeripeierSBB: let's keep the 2x border width. only alternative I see would be px-to-rem with 2px.

jeripeierSBB (changes requested):
Just very small stuff. Looks great now!

jeripeierSBB (approved):
Great work!

src/elements/stepper/stepper/readme.md
```diff
@@ -0,0 +1,111 @@
...
+  <sbb-step-label>Step label 2</sbb-step-label>
+  <sbb-step>
+    <form>
+      <sbb-form-field>...</sbb-form-field>
+    </form>
+  </sbb-step>
+</sbb-stepper>
+```
+
+Calling the `reset()` method on the `sbb-stepper`, will reset the wrapping form or every form of each step and select the first step.
```
DavideMininni-Fincons, line 74:
```suggestion
Calling the `reset()` method on the `sbb-stepper` will reset the wrapping `form` or, if they are present, every `form` in each step; then it will select the first step.
```

src/elements/stepper/stepper/stepper.stories.ts
```diff
@@ -0,0 +1,445 @@
+import { withActions } from '@storybook/addon-actions/decorator';
```
DavideMininni-Fincons, line 1: Minor improvement: I see some duplicated code here in templates, which can be refactored.

src/elements/stepper/step-label/readme.md
```diff
@@ -0,0 +1,54 @@
...
+<sbb-step-label>Step label</sbb-step-label>
+```
+
+## Slots
+
+It has an implicit slot name: `step-label`.
+
+## States
+
+It can be `disabled`.
```
DavideMininni-Fincons, line 13:
```suggestion
The component can be displayed in `disabled` state using the self-named property.
```

src/elements/stepper/step-label/readme.md
```diff
@@ -0,0 +1,54 @@
...
+<!-- Displays a tick icon in the prefix circle -->
+<sbb-step-label icon-name="tick-small">Step label</sbb-step-label>
+
+<!-- Displays a number in the prefix circle -->
+<sbb-step-label>Step label</sbb-step-label>
+```
+
+## Accessibility
+
+The aria values `aria-controls`, `aria-setsize`, `aria-posinset` are set automatically.
```
DavideMininni-Fincons, line 33:
```suggestion
The accessibility properties `aria-controls`, `aria-setsize`, `aria-posinset` are set automatically.
```

DavideMininni-Fincons (changes requested):
Great work! some minor notes, mainly about readmes

src/elements/stepper/stepper/stepper.ts
```diff
@@ -101,6 +101,7 @@ export class SbbStepperElement extends LitElement {
       this.querySelectorAll('form').forEach((form) => form.reset());
     }
     this.selectedIndex = 0;
+    this.selected?.label?.focus();
```
jeripeierSBB, line 104: Is this not a little bit invasive, as also, if programmatically reset, it steals the focus from anywhere to here?

src/elements/stepper/stepper/stepper.stories.ts
```diff
@@ -264,7 +267,9 @@ const WithMultipleFormsTemplate = (args: Args): TemplateResult => {
 
       <sbb-step-label icon-name="tick-small">Step 4</sbb-step-label>
       <sbb-step>
-        <div style="margin-block-end: var(--sbb-spacing-fixed-4x)">You are now done.</div>
+        <div style="margin-block-end: var(--sbb-spacing-fixed-4x)" aria-live="polite">
```
jeripeierSBB, line 270: Does that really work as it is always present?

src/elements/stepper/stepper/stepper.stories.ts
```diff
@@ -155,7 +155,9 @@ const WithSingleFormTemplate = (args: Args): TemplateResult => {
 
         <sbb-step-label icon-name="tick-small">Step 4</sbb-step-label>
         <sbb-step>
-          <div style="margin-block-end: var(--sbb-spacing-fixed-4x)">You are now done.</div>
+          <div style="margin-block-end: var(--sbb-spacing-fixed-4x)" aria-live="polite">
```
jeripeierSBB, line 158: Should we not mention it also in readme? Like pointing out what to think at

jeripeierSBB (commented):
Thanks for the fast fixes

src/elements/stepper/stepper/stepper.ts
dauriamarco: Not sure, If I don't do that the focus would remain on the reset button in the last step while it come back to the first one. So it made sense to me to focus on the first step label.

src/elements/stepper/stepper/stepper.stories.ts
dauriamarco: It works at least the first time you land on the last step, but I didn't test it on every screen reader. I will do it with the next generated link.

dauriamarco: Yes, it works, I tested it with NVDA and JAWS as well.

jeripeierSBB: ok nice. so aria live does not when it's hidden and later presented.

src/elements/stepper/stepper/stepper.ts
jeripeierSBB: what about just calling blur() on the reset button? Or otherwise, only call it if triggered by user action?

src/elements/stepper/stepper/stepper.ts
```diff
@@ -0,0 +1,307 @@
...
+  /** Resets the form in which the stepper is nested or every form of each step, if any. */
+  public reset(): void {
+    const closestForm = this.closest('form');
+    if (closestForm) {
+      closestForm.reset();
+    } else {
+      this.querySelectorAll('form').forEach((form) => form.reset());
+    }
+    this.selectedIndex = 0;
+    this.selected?.label?.focus();
```
kyubisation, line 104: Briefly discussed with Jeri. @dauriamarco What do you think?
```suggestion
    if (document.activeElement?.closest('sbb-stepper') === this) {
      this.selected?.label?.focus();
    }
```

dauriamarco: works for me 👍

src/elements/stepper/stepper/stepper.ts
```diff
@@ -101,7 +101,10 @@ export class SbbStepperElement extends LitElement {
       this.querySelectorAll('form').forEach((form) => form.reset());
     }
     this.selectedIndex = 0;
-    this.selected?.label?.focus();
+    // In case the focus is currently inside the stepper, we reset the focus to the first/selected step label.
```
jeripeierSBB, line 104: Could you add a test case please :)

jeripeierSBB (approved):
Cool, hopefully the last LGTM ;-)

src/elements/core/styles/a11y.scss
```diff
@@ -3,3 +3,17 @@
 .sbb-screen-reader-only {
   @include a11y.screen-reader-only;
 }
+
+.sbb-focus-outline:focus-visible {
+  @include a11y.focus-outline;
+
+  border-radius: var(--sbb-border-radius-1x);
```
jeripeierSBB, line 10: I think we should not include border-radius as it can wrongly override consumers one.

src/elements/stepper/stepper/stepper.e2e.ts
```diff
@@ -19,25 +19,25 @@ describe('sbb-stepper', () => {
       <sbb-stepper>
         <sbb-step-label>Step 1</sbb-step-label>
         <sbb-step>
-          Step one content.
+          <div tabindex="0" id="step-one-content"></div>Step one content.</div>
```
jeripeierSBB, line 22: looks like the tag was wrongly closed (also in next 4 changes)

jeripeierSBB (changes requested):

jeripeierSBB (approved):
LGTM :D

DavideMininni-Fincons (changes requested):
Some minor details to fix:
- the long label does not wrap in vertical. is it ok? didn't find a Figma sketch for it
- found a weird behavior: on the first story, if i clear the input, than set the focus on the first step, moving with right arrow goes to second step (not sure if it's correct). After that i cant get back to the first one.
- refactored a bit the stories, see https://github.com/sbb-design-systems/lyne-components/compare/chore/stepper-stories?expand=1

DavideMininni-Fincons (approved):

src/elements/stepper/step-label/step-label.scss
```diff
@@ -0,0 +1,153 @@
...
+    inset-block-start: calc(
+      var(--sbb-font-size-text-l) * (var(--sbb-typo-line-height-body-text) / 2) +
+        (var(--sbb-border-width-1x) / 2)
+    );
+    inset-inline-start: calc(var(--sbb-step-label-prefix-size) / 2);
+    line-height: 1;
+    z-index: 1;
+    transform: translate(
+      -50%,
+      calc(-50% + var(--sbb-step-label-translate-y-content-hover, #{sbb.px-to-rem-build(0)}))
```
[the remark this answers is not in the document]
jeripeierSBB: I think it's only important to have a unit

PR #2495, opened 2024-03-18
chore: remove 'vite-plugin-turbosnap' dependency

TomMenga:
With Storybook 8.0 they started supporting the generation of `preview-stats.json` (used by turbosnap) with Vite.
We don't need the external plugin anymore 😄 

Ps. Chromatic will fail until we merge this fix


jeripeierSBB (approved):
Cool, thanks!

PR #2506, opened 2024-03-19
feat(sbb-form-field): support the textarea inside the `<sbb-form-field>`

### LINKED ISSUE #2497: story(sbb-form-field): support textarea

dauriamarco:
Closes #2497

BREAKING CHANGE: The css var `--sbb-form-field-height` has been renamed to `--sbb-form-field-min-height`.

### DIFF
--- a/src/components/form-field/form-field/form-field.scss
+++ b/src/components/form-field/form-field/form-field.scss
@@ -265,6 +265,11 @@
   }
 }
 
+// As the calendar should be shown below the form field border, we have to stretch the toggle's height.
+::slotted(sbb-datepicker-toggle) {
+  height: calc((var(--sbb-form-field-min-height) + var(--sbb-size-icon-ui-small)) / 2);
+}
+
 @include sbb.if-forced-colors {
   // Align with prefix color
   ::slotted(*[slot='suffix']) {
--- a/src/components/form-field/form-field/form-field.scss
+++ b/src/components/form-field/form-field/form-field.scss
@@ -10,7 +10,7 @@
   --sbb-form-field-border-style: solid;
   --sbb-form-field-border-radius: var(--sbb-border-radius-4x);
   --sbb-form-field-border-width: var(--sbb-border-width-1x);
-  --sbb-form-field-height: var(--sbb-size-form-element-m-min-height);
+  --sbb-form-field-min-height: var(--sbb-size-form-element-m-min-height);
   --sbb-form-field-label-color: var(--sbb-color-metal);
   --sbb-form-field-prefix-color: var(--sbb-color-metal);
   --sbb-form-field-padding-inline: var(--sbb-spacing-fixed-3x);
@@ -60,7 +60,7 @@
 }
 
 :host([size='l']) {
-  --sbb-form-field-height: var(--sbb-size-form-element-l-min-height);
+  --sbb-form-field-min-height: var(--sbb-size-form-element-l-min-height);
   --sbb-form-field-padding-inline: var(--sbb-spacing-responsive-xxxs);
 }
 
@@ -160,6 +160,10 @@
   --sbb-form-field-background-color: var(--sbb-color-white);
 }
 
+:host([data-input-type='textarea']) {
+  --sbb-form-field-max-height: fit-content;
+}
+
 .sbb-form-field__space-wrapper {
   @include sbb.zero-width-space;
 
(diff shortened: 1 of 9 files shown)


src/components/form-field/form-field/form-field.scss
```diff
@@ -175,13 +179,19 @@
...
 
   :host([data-input-focused][data-focus-origin='keyboard']) & {
     @include sbb.focus-outline;
   }
 
+  // Prevents page content from shifting on textarea focus
+  :host([data-input-focused][data-input-type='textarea']) & {
+    margin-block-end: calc(var(--sbb-form-field-border-width) / 2 * -1);
+  }
+
```
TomMenga, line 194: :(

src/components/form-field/form-field/form-field.scss
```diff
@@ -370,6 +396,26 @@
...
+.sbb-form-field__input ::slotted(textarea) {
+  resize: vertical;
+  white-space: normal;
+  overflow-y: scroll;
+  min-height: calc((var(--sbb-typo-line-height-body-text) * 1em));
+
+  // Prevents input from shifting on textarea focus
+  :host(:not([data-input-focused])) & {
+    padding-block-start: var(--sbb-form-field-border-width);
+  }
```
TomMenga, line 408: :(

src/components/form-field/form-field/form-field.scss
```diff
@@ -291,6 +312,11 @@
     }
   }
 
+  // Prevents label from shifting on textarea focus
+  :host(:not([data-input-focused])[data-input-type='textarea']) & {
+    padding-block-start: var(--sbb-form-field-border-width);
+  }
+
```
TomMenga, line 319: :(

TomMenga (changes requested):
It is LGTM except for the "not focus" rules which feel a bit of a workaround 😞 
Are they unavoidable? Why is it just the textarea that has problems?

src/components/form-field/form-field/form-field.scss
dauriamarco: I konw, I also don't like them and I'm open to possible alternative solutions. To answer your question, it is just a textarea problem due to the `height: fit-content` to enable the resize; as long as you have a fixed height like in any other case, the label/content shifting doesn't occur.

dauriamarco: I would also ask @jeripeierSBB or @kyubisation, what do you think about it?

TomMenga: Understood the problem :/ 
The only alternative I see is to move the focus border to a `::before` element. We already do something similar in other components

src/components/form-field/form-field/form-field.scss
```diff
@@ -203,15 +213,26 @@
   }
 }
 
+::slotted(*[slot='prefix']) {
```
jeripeierSBB, line 221: * should not be needed here
```suggestion
::slotted([slot='prefix']) {
```

src/components/form-field/form-field/form-field.scss
```diff
@@ -203,15 +213,26 @@
...
 // Ensure slotted inline elements are vertically centered
 ::slotted(*[slot='prefix']),
 ::slotted(*[slot='suffix']) {
   min-width: var(--sbb-icon-svg-width);
   display: flex;
-}
 
-::slotted(*[slot='prefix']) {
-  color: var(--sbb-form-field-prefix-color);
+  :host([data-input-type='textarea']) & {
```
jeripeierSBB, line 231: Could / should we do it in general (not scoping to textarea) to avoid to much blocks here?

src/components/form-field/form-field/form-field.scss
```diff
@@ -370,6 +396,69 @@
...
+  resize: vertical;
+  white-space: break-spaces;
+  overflow-y: auto;
+  min-height: calc((var(--sbb-typo-line-height-body-text) * 1em));
+
+  :host([negative]) & {
+    @include sbb.scrollbar($negative: true);
+  }
+
+  :host(:is(:not([data-slot-names~='label'], [label]), [hidden-label])) & {
```
jeripeierSBB, line 412: Can you describe why this is needed?

src/components/form-field/form-field/form-field.scss
```diff
@@ -370,6 +396,69 @@
...
+      var(--sbb-form-field-min-height) / 2 - (var(--sbb-typo-line-height-body-text) * 0.5em) - var(
+          --sbb-form-field-border-width
+        )
+    );
+  }
+}
+
+// Custom resize  handle
+
+:host([data-input-type='textarea']) .sbb-form-field__input-container {
```
jeripeierSBB, line 423: Looks nice, a little bit much of code :)

src/components/form-field/form-field/form-field.ts
```diff
@@ -245,6 +250,10 @@ export class SbbFormFieldElement extends SbbNegativeMixin(LitElement) {
     this._readInputState();
     this._registerInputListener();
 
+    if (this._input.tagName === 'TEXTAREA') {
```
jeripeierSBB, line 253: I'd suggest to create a test for the textarea case like for the other cases.

jeripeierSBB (changes requested):
Nice solution. some topics open

src/components/form-field/form-field/form-field.scss
jeripeierSBB: Could solve it according to your suggestion @TomMenga . Thanks :)

src/components/form-field/form-field/form-field.scss
dauriamarco: Manuel is preparing for us a design proposal to style the resize handle 👍

kyubisation (approved):
LGTM 👍 
Nice work!

jeripeierSBB (approved):
LGTM :)

PR #2508, opened 2024-03-19
fix(sbb-header): fix header shadow on keyboard navigation

dauriamarco:
Doesn't show the shadow if the header is not in `position: fixed`.


jeripeierSBB (approved):
LGTM, thanks :)

PR #2518, opened 2024-03-25
test(sbb-checkbox, sbb-toggle-check): stabilize tests with retries

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
Can you maybe add a comment to explain why the retries are necessary, so in the future we can check whether they can be removed?

PR #2520, opened 2024-03-25
refactor: rename component name from `sbb-screenreader-only` to `sbb-screeen-reader-only`

jeripeierSBB:
As in English language `screen reader` is written in two words, we like to change the naming and also aligning it to existing SASS mixin name.

BREAKING CHANGE: renamed component `sbb-screenreader-only` to `sbb-screeen-reader-only`.


TomMenga (approved):
LGTM 👍

PR #2523, opened 2024-03-26
fix(sbb-form-field): remove label property and attribute

jeripeierSBB:
In order to support SSR we have to remove the label property. As replacement the `<label>` tag should be used.

BREAKING CHANGE: `label` property and attribute of `<sbb-form-field>` was removed. Use `<label>` tag inside `<sbb-form-field>` to provide the label information. E.g. `<sbb-form-field label="Example">...</sbb-form-field>` becomes `<sbb-form-field><label>Example</label>...</sbb-form-field>`


TomMenga (approved):
LGTM 👍

kyubisation (approved):
LGTM 👍

PR #2524, opened 2024-03-26
fix: fix scrollbar styles for Chrome

jeripeierSBB:
Chrome started to support scrollbar-width and scrollbar-color CSS properties which we are currently using as a fallback for Firefox. However, when using it in Chrome, it disables all the -webkit-scrollbar stylings. With the new supports check we can avoid to switch to the scrollbar-color approach in Chrome as long as the old approach still works. Painpoint: the new approach doesn't allow a different color for hover.

### DIFF
--- a/src/components/core/styles/mixins/scrollbar.scss
+++ b/src/components/core/styles/mixins/scrollbar.scss
@@ -36,10 +36,14 @@
     display: none;
   }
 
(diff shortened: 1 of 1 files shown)


dauriamarco (approved):
LGTM 👍

PR #2535, opened 2024-04-02
build: fix react packages types

kyubisation:
This PR fixes the current type issues with the React package and adds a build check to fail the build, if there are type errors.


jeripeierSBB (approved):
LGTM

PR #2507, opened 2024-03-19
refactor(disable-animation): migrate disable-animation mechanism

TomMenga:
Closes #2272 

### For reviewers
See `src/components/core/styles/mixins/a11y.scss` and `src/components/core/styles/typography.scss` for the actual implementation of the new mechanism.
All the other changes are adaptations to the new disable-animation system.

------
### Commit message

refactor(disable-animation): migrate disable-animation mechanism

BREAKING CHANGE: Add the `sbb-disable-animation` CSS class to disable animations for the element and all its children. Consequently, all `disable-animation` properties of the following components have been removed.
- Removed `disable-animation` property from `sbb-accordion`
- Removed `disable-animation` property from `sbb-alert`
- Removed `disable-animation` property from `sbb-autocomplete`
- Removed `disable-animation` property from `sbb-datepicker-toggle`
- Removed `disable-animation` property from `sbb-dialog`
- Removed `disable-animation` property from `sbb-expansion-panel`
- Removed `disable-animation` property from `sbb-image`
- Removed `disable-animation` property from `sbb-loading-indicator`
- Removed `disable-animation` property from `sbb-menu`
- Removed `disable-animation` property from `sbb-navigation`
- Removed `disable-animation` property from `sbb-navigation-section`
- Removed `disable-animation` property from `sbb-notification`
- Removed `disable-animation` property from `sbb-overlay`
- Removed `disable-animation` property from `sbb-popover`
- Removed `disable-animation` property from `sbb-select`
- Removed `disable-animation` property from `sbb-selection-panel`
- Removed `disable-animation` property from `sbb-sticky-bar`
- Removed `disable-animation` property from `sbb-toast`
- Removed `disable-animation` property from `sbb-toggle`

feat(alert): add `animation` property
  BREAKING-CHANGE: Removed `disable-animation` property. Use `animation='none'` instead.

feat(notification): add `animation` property
  BREAKING-CHANGE: Removed `disable-animation` property. Use `animation='none'` instead or depending on the case 'open' or 'close'.

PR #2541, opened 2024-04-03
refactor(expansion-panel, notification): migrate from transition to animation

### LINKED ISSUE #2540: Migrate `sbb-notification` and `sbb-expansion-panel` transition to animation

TomMenga:
This PR Closes #2540 

~~Blocked until #2507 is merged~~


src/components/expansion-panel/expansion-panel/expansion-panel.scss
```diff
@@ -3,19 +3,19 @@
 // Box-sizing rules contained in typography are not traversing Shadow DOM boundaries. We need to include box-sizing mixin in every component.
 @include sbb.box-sizing;
 
+// Open/Close animation vars
+$open-anim-rows-from: 0fr;
+$open-anim-rows-to: 1fr;
+$open-anim-opacity-from: 0;
+$open-anim-opacity-to: 1;
+
```
MarioCastigliano, line 11: I've noticed this pattern (with the same variable names and animations) is repeated in 4 components, would it make sense to extract it in a mixin so that we can reuse it in a more efficient way?

MarioCastigliano (commented):
Thanks for your work, just one quick question

src/components/expansion-panel/expansion-panel/expansion-panel.scss
TomMenga: I've thought about creating a mixin for the open/close animation that would also include the declaration. However, even if part of it is in common, the resulting mixin would have too many parameters. 
I feel like factorizing only a part of the implementation is a bad practice :/

dauriamarco (changes requested):
Hi, I noticed that the open animation when you add a new notification doesn't work. Do you see the same problem?

TomMenga:
> Hi, I noticed that the open animation when you add a new notification doesn't work. Do you see the same problem?

Yep :/
I broke it with the resolution of the last merge conflict. It's fixed now, though

kyubisation (approved):
LGTM 👍 
Thanks! 😃

PR #2542, opened 2024-04-03
refactor: avoid setting attributes in render method

jeripeierSBB:


TomMenga (approved):
LGTM 👍 
Nice work
The poor @DavideMininni-Fincons will not be happy about the changes to the autocomplete & co D:

PR #2547, opened 2024-04-04
fix: fix boolean handling in react wrapper

jeripeierSBB:


src/react/core/create-component.ts
```diff
@@ -285,6 +285,9 @@ export const createComponent = <I extends HTMLElement, E extends EventNames = {}
           reactProps[attributeName] = toAttribute ? toAttribute(v) : v;
         } else if (v) {
           reactProps[attributeName] = '';
+          // Some boolean props like `checke` don't react to mutations after user interaction.
```
kyubisation, line 288:
```suggestion
          // Some boolean props like `checked` don't react to mutations after user interaction.
```

kyubisation (approved):
LGTM 👍

PR #2553, opened 2024-04-05
fix(sass): inline external .scss files during build

### LINKED ISSUE #2513: Inline Design Token SCSS File during prod build
To avoid consumers breaking the project when using our sass, we need to inline/copy the scss file from the design tokens packages during the prod build and adapt any import in the output.

kyubisation:
This allows our .scss files to be consumed without depending on the `@sbb-esta/lyne-design-tokens` package.

Closes #2513

### DIFF
--- a/src/components/core/styles/core/functions.scss
+++ b/src/components/core/styles/core/functions.scss
@@ -1,8 +1,6 @@
 @use 'sass:math';
 
-/**
- * size: value in px (no unit)
- */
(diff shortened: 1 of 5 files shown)


jeripeierSBB (approved):
LGTM, thanks!
Don't forget to link our issue, so it's getting closed.
I'd also like to discuss something directly with you.

PR #2563, opened 2024-04-08
fix: use valid import/export syntax

kyubisation:
Previously we were using the common syntax of importing/exporting files without the extension. However, this is not natively supported in browsers and also not in Node.js with ESM. This change adds a linting rule that automatically adds the .js extension
to any relative import.


jeripeierSBB (approved):
I checked every single file to ensure that everything is correct to finally see at the end of the PR that there is a new linting rule :P

TomMenga (approved):
LGTM 👍 
I think you broke the record for most changed files in a pr

kyubisation:
> LGTM 👍 I think you broke the record for most changed files in a pr

I'm torn between being proud and distraught 😅

PR #2574, opened 2024-04-11
refactor: remove obsolete host-component-properties mixin

jeripeierSBB:


src/components/link/common/link.scss
```diff
@@ -1,8 +1,7 @@
 @use '../../core/styles' as sbb;
 
-// Default component properties, defined for :host. Properties which can not
-// travel the shadow boundary are defined through this mixin
-@include sbb.host-component-properties;
+// Box-sizing rules contained in typography are not traversing Shadow DOM boundaries. We need to include box-sizing mixin in every component.
```
DavideMininni-Fincons, line 3: I see no specific `display` rule on host; should it be added?

src/components/timetable-duration/timetable-duration.scss
```diff
@@ -1,5 +1,8 @@
 @use '../core/styles' as sbb;
 
+// Box-sizing rules contained in typography are not traversing Shadow DOM boundaries. We need to include box-sizing mixin in every component.
```
DavideMininni-Fincons, line 3: Same as link

DavideMininni-Fincons (changes requested):
Great work, only two identical comments about missing (?) rule.

src/components/link/common/link.scss
jeripeierSBB: It's because link.scss is always used together with either block-link-scss or inline-link.scss. they both have their own display property. inline-link: inline-block, and blocklink: block

src/components/timetable-duration/timetable-duration.scss
jeripeierSBB: as there was no display property before migration, I thought to not add one now. for the box-sizing I think it's important that every component uses it to have a good consistency. However to be also consistent with display, I added inline-block to this component :)

src/components/link/common/link.scss
DavideMininni-Fincons: Oh, right :)

DavideMininni-Fincons (approved):
LGTM, danke! :)

PR #2575, opened 2024-04-11
refactor: restructure entry points

kyubisation:
BREAKING CHANGE: The entry points for the `@sbb-esta/lyne-components` have changed from extension-less to including the extension (e.g. `@sbb-esta/lyne-components/button` to `@sbb-esta/lyne-components/button.js`).


jeripeierSBB (approved):
LGTM

PR #2579, opened 2024-04-15
fix: update import paths missed in the migration

kyubisation:
This PR fixes the remaining missed import paths that were not adapted. It additionally adds a vite plugin check, to ensure the build would break, if an illegal import was used.


jeripeierSBB (approved):
LGTM and thanks for fixing my bugs :D

PR #2589, opened 2024-04-16
fix: adapt react package import paths missed during migration

kyubisation:
This PR fixes the remaining mistaken imports in the React package.


jeripeierSBB (approved):
Yaaaay, thanks for the fix

PR #2594, opened 2024-04-17
refactor(sbb-title): extract base class

jeripeierSBB:


src/components/title/title.ts
```diff
@@ -1,63 +1,16 @@
...
-    /* eslint-disable lit/binding-positions */
-    return html`
-      <${unsafeStatic(TAGNAME)} class="sbb-title" role="presentation">
-        <slot></slot>
-      </${unsafeStatic(TAGNAME)}>
-    `;
-    /* eslint-enable lit/binding-positions */
-  }
+export class SbbTitleElement extends SbbTitleBase {
+  public static override styles: CSSResultGroup = [SbbTitleBase.styles, style];
```
kyubisation, line 13: Both base title and actual title import from title.scss. This means the styles are duplicated?

jeripeierSBB: good catch :) chromatic would have claimed soonish :D

kyubisation (approved):
LGTM 👍

PR #2599, opened 2024-04-18
refactor: create base class for overlay functionality

### LINKED ISSUE #2562: story: create base class for overlay functionality
A lot of logic of our overlay components is copied. We should figure out if a base class or mixin make sense and implement it.

DavideMininni-Fincons:
Two points for discussion which possibly need fix:

- naming: `SbbDialogBaseElement` uses 'dialog' for naming props (`dialogRefs`, `dialogController`), which can result not totally correct in `SbbOverlayElement`, which uses 'overlay' keyword (`_overlayContentElement`). 
Adding a `SbbOverlayBase` class (which could be extended also by menu, navigation, etc) increases the noise;

- lint issue: the dialog base class declares `public static readonly events = { ... }` which is common between the dialog and the overlay; however if in the overlay is overridden as `public static readonly override events = { ...super.event, backClick: 'requestBackAction' }`, the 'missing-component-documentation-rule' breaks, because it expects only properties and not a spread operator with super calls (fixable by #2600).
edit: forgot to tag @kyubisation @jeripeierSBB 

BREAKING CHANGE
- type `SbbDialogCloseEventDetails` has been renamed to `SbbOverlayCloseEventDetails`

Closes #2562


jeripeierSBB:
> There is an issue with the generation of the custom-elements.json, and then of the docs.
> 
> The notification component has a `private` open method. If I set the `open` and `close` methods as protected in the OverlayBaseClass and then I override it as `public` in all the components except the notification `open`, the change is not reflected and all the methods disappear from the readmes. If I set the `open` and `close` methods as public in the OverlayBaseClass and then I add `@internal` in the `open` method in the notification to hide it, the internal is not taken into account and the method is visible in the readme.
> 
> Should i open a bug to https://github.com/webcomponents/custom-elements-manifest ?
> 
> edit: also the selection panel has private open/close methods, and has the same issue

Should we maybe exclude them from inheriting the base class? Opening a bug report could anyways help. but potentially they have the same rule as for inheriting itself: public props cannot be made private.

DavideMininni-Fincons:
> > There is an issue with the generation of the custom-elements.json, and then of the docs.
> > The notification component has a `private` open method. If I set the `open` and `close` methods as protected in the OverlayBaseClass and then I override it as `public` in all the components except the notification `open`, the change is not reflected and all the methods disappear from the readmes. If I set the `open` and `close` methods as public in the OverlayBaseClass and then I add `@internal` in the `open` method in the notification to hide it, the internal is not taken into account and the method is visible in the readme.
> > Should i open a bug to https://github.com/webcomponents/custom-elements-manifest ?
> > edit: also the selection panel has private open/close methods, and has the same issue
> 
> Should we maybe exclude them from inheriting the base class? Opening a bug report could anyways help. but potentially they have the same rule as for inheriting itself: public props cannot be made private.

good, i'll exclude them for now and add a note

i would expect that 
base protected => derived public 
should work :'(

src/components/dialog/dialog/dialog.ts
```diff
@@ -137,50 +68,116 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
...
+    // Close dialog on backdrop click
+    this.addEventListener('pointerdown', this._pointerDownListener, {
+      signal: this.dialogController.signal,
+    });
+    this.addEventListener('pointerup', this._closeOnBackdropClick, {
+      signal: this.dialogController.signal,
+    });
+  }
+
+  protected override firstUpdated(_changedProperties: PropertyValues): void {
```
TomMenga, line 96:
```suggestion
  protected override firstUpdated(changedProperties: PropertyValues): void {
```

src/components/dialog/dialog/dialog.ts
```diff
@@ -137,50 +68,116 @@ export class SbbDialogElement extends SbbNegativeMixin(LitElement) {
...
+    // Synchronize the negative state before the first opening to avoid a possible color flash if it is negative.
+    this._dialogTitleElement = this.querySelector('sbb-dialog-title')!;
+    this._syncNegative();
+    super.firstUpdated(_changedProperties);
+  }
+
+  protected override willUpdate(changedProperties: PropertyValues<this>): void {
+    if (changedProperties.has('negative')) {
```
TomMenga, line 104:
```suggestion
  protected override willUpdate(changedProperties: PropertyValues<this>): void {
    super.willUpdate(changedProperties);
    if (changedProperties.has('negative')) {
```

src/components/overlay/overlay.ts
```diff
@@ -58,253 +47,85 @@ export class SbbOverlayElement extends SbbNegativeMixin(LitElement) {
...
-  // Last element which had focus before the overlay was opened.
-  private _lastFocusedElement?: HTMLElement;
 
-  private _language = new SbbLanguageController(this);
-
-  /**
-   * Opens the overlay element.
-   */
+  /** Opens the component. */
   public open(): void {
```
TomMenga, line 73:
```suggestion
  public override open(): void {
```
I would mark the abstract implementation with the `override` keyword (might be a personal preference though)

TomMenga (changes requested):
Good job 👍 
Some minor fixes

src/components/overlay/overlay.ts
DavideMininni-Fincons: hmm.. since they are implementations from zero and not a rewriting of an existing method, i prefer not to add it..

TomMenga (approved):
LGTM 👍

src/components/overlay/readme.md
```diff
@@ -78,27 +78,27 @@ When using a button to trigger the overlay, ensure to manage the appropriate ARI
 | ------------------------- | --------------------------- | ------- | ---------------------------- | ------- | --------------------------------------------------------------------------------------------------------------------- |
 | `accessibilityBackLabel`  | `accessibility-back-label`  | public  | `\| string     \| undefined` |         | This will be forwarded as aria-label to the back button element.                                                      |
 | `accessibilityCloseLabel` | `accessibility-close-label` | public  | `\| string     \| undefined` |         | This will be forwarded as aria-label to the close button element.                                                     |
-| `accessibilityLabel`      | `accessibility-label`       | public  | `string \| undefined`        |         | This will be forwarded as aria-label adn will describe the purpose of the dialog.                                     |
+| `accessibilityLabel`      | `accessibility-label`       | public  | `string \| undefined`        |         | This will be forwarded as aria-label to the relevant nested element.                                                  |
```
jeripeierSBB, line 81: I'd prefer the more specific phrase as the consumer should now what it really does

DavideMininni-Fincons: Replaced with a mix of the two phrases

src/components/dialog/dialog/dialog-base-element.ts
```diff
@@ -0,0 +1,157 @@
...
+    if (this.state === 'opened') {
+      applyInertMechanism(this);
+    }
+  }
+
+  protected override firstUpdated(changedProperties: PropertyValues<this>): void {
+    this.ariaLiveRef =
+      this.shadowRoot!.querySelector<SbbScreenReaderOnlyElement>('sbb-screen-reader-only')!;
+
+    super.firstUpdated(changedProperties);
```
jeripeierSBB, line 78: Is it necessary to have super call at the end?

src/components/dialog/dialog/dialog-base-element.ts
```diff
@@ -0,0 +1,157 @@
...
+import { hostContext, SbbScrollHandler } from '../../core/dom.js';
+import { EventEmitter } from '../../core/eventing.js';
+import { i18nDialog } from '../../core/i18n.js';
+import type { SbbDialogCloseEventDetails } from '../../core/interfaces.js';
+import { SbbNegativeMixin } from '../../core/mixins.js';
+import { applyInertMechanism, removeInertMechanism } from '../../core/overlay.js';
+import type { SbbScreenReaderOnlyElement } from '../../screen-reader-only.js';
+
+// A global collection of existing dialogs
+export const dialogRefs: SbbDialogBaseElement[] = [];
```
jeripeierSBB, line 16: Is that ok, from feature perspective, if we mix dialog and overlay in the same? Probably yes, but really? :D

src/components/notification/notification.ts
```diff
@@ -38,6 +38,7 @@ const DEBOUNCE_TIME = 150;
  */
 @customElement('sbb-notification')
 export class SbbNotificationElement extends LitElement {
+  // FIXME could inherit from SbbOverlayBaseElement but readme generator is flawed
```
jeripeierSBB, line 41: Could we address this somehow?

src/components/overlay/overlay.ts
```diff
@@ -1,50 +1,39 @@
...
  * the `z-index` can be overridden by defining this CSS variable. The default `z-index` of the
  * component is set to `var(--sbb-overlay-default-z-index)` with a value of `1000`.
  */
 @customElement('sbb-overlay')
-export class SbbOverlayElement extends SbbNegativeMixin(LitElement) {
+export class SbbOverlayElement extends SbbDialogBaseElement {
   public static override styles: CSSResultGroup = style;
-  public static readonly events = {
+
+  // FIXME using { ...super.events, backClick: 'requestBackAction' } breaks the eslint missing-component-documentation-rule
```
jeripeierSBB, line 35: This should be fixable as it is on our side

src/components/select/select.ts
```diff
@@ -49,10 +49,12 @@ export interface SelectChange {
   role: ariaRoleOnHost ? 'listbox' : null,
 })
 export class SbbSelectElement extends SbbUpdateSchedulerMixin(
-  SbbDisabledMixin(SbbNegativeMixin(LitElement)),
+  SbbDisabledMixin(SbbNegativeMixin(SbbOverlayBaseElement)),
 ) {
   public static override styles: CSSResultGroup = style;
-  public static readonly events = {
+
+  // FIXME using { ...super.events, backClick: 'requestBackAction' } breaks the eslint missing-component-documentation-rule
```
jeripeierSBB, line 56: Same comment again, we should try adapting our rules :)

src/components/selection-panel/selection-panel.ts
```diff
@@ -25,6 +25,7 @@ import '../divider.js';
  */
 @customElement('sbb-selection-panel')
 export class SbbSelectionPanelElement extends LitElement {
+  // FIXME could inherit from SbbOverlayBaseElement but readme generator is flawed
```
jeripeierSBB, line 28: Don't forget to address

jeripeierSBB (changes requested):
Generally ok, but I think we need a naming discussion ;): Could you maybe set up an appointment with @kyubisation and me?
Topic:
 - Overlay, vs dialog or something else in base classes.
 - Adapting docs generation
 - Adapting linting rule
 - Other open topics?

Thank you :)

src/components/dialog/dialog/dialog-base-element.ts
DavideMininni-Fincons: If i remove it, the `ESLint: Method needs super call.(lyne/needs-super-call-rule)` error appears.

src/components/dialog/dialog/dialog-base-element.ts
DavideMininni-Fincons: As discussed yesterday, i renamed everything to 'overlay<stuff>' as it's a more generic prefix.

src/components/select/select.ts
DavideMininni-Fincons: Added a link to https://github.com/sbb-design-systems/lyne-components/issues/2600 in the fixme description

src/components/notification/notification.ts
DavideMininni-Fincons: On our side, i think we have two ways:
- remove the abstract open/close methods from the `SbbOpenCloseBaseElement` class;
- change the visibility of the `open` method from private to public in the two involved components

DavideMininni-Fincons: As discussed, the fixme is updated with the link to https://github.com/open-wc/custom-elements-manifest/issues/253

kyubisation (approved):
LGTM 👍 
Thanks! 😃

jeripeierSBB (approved):
LGTM

PR #2606, opened 2024-04-22
feat(notification): add size 's'

### LINKED ISSUE #2597: lean: notification: size S
[Figma](https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?node-id=20220%3A602306&mode=dev)

TomMenga:
Closes #2597


mcilurzo:
ux review:
– https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj?type=design&node-id=36822-9499&mode=design#778164857

src/components/notification/notification.scss
```diff
@@ -157,6 +175,10 @@
           ) / 2
       );
     }
+
+    :host([size='s']) & {
+      margin-block-start: 0;
```
jeripeierSBB, line 180: Would it make sense to introduce a css var as well ?. and is it still working without and with title?

src/components/notification/notification.scss
```diff
@@ -82,6 +84,18 @@
   --sbb-notification-type-color-sass: #{color.mix(sbb-tokens.$sbb-color-red, white, 5%)};
 }
 
+:host([size='s']) {
+  --sbb-notification-padding-block: var(--sbb-spacing-responsive-xxxs);
+  --sbb-notification-padding-inline: var(--sbb-spacing-responsive-xxs);
+  --sbb-notification-content-padding-inline: var(--sbb-spacing-responsive-xxxs)
+    var(--sbb-spacing-responsive-xxs);
+
+  @include sbb.mq($to: small) {
```
jeripeierSBB, line 93: could you invert it and use $from ?

src/components/notification/notification.scss
TomMenga: Not sure it improved much in terms of readability though :/

jeripeierSBB (approved):
LGTM, one unimportant comment left :)

src/components/notification/notification.scss
jeripeierSBB: it's the same I think :D but a little more like our standards. for me ok. I hope you can live with it as well :D

src/components/notification/notification.scss
```diff
@@ -138,26 +171,15 @@
...
-      margin-block-start: calc(
-        (
-            (var(--sbb-font-size-title-5) * var(--sbb-typo-line-height-body-text)) - var(
-                --sbb-size-icon-ui-small
-              )
-          ) / 2
-      );
-    }
-  }
+  margin-block-start: var(--sbb-notification-icon-margin-block-start, 'initial');
```
jeripeierSBB, line 182: Does that not work? alternatively, just set it to 0 as standard?
```suggestion
  margin-block-start: var(--sbb-notification-icon-margin-block-start);
```

TomMenga: I would work without a default value too. But, by having one, I feel like it adds the explicit meaning "it is meant to be undefined in some cases".
If you're ok, I'll change it to '0'

jeripeierSBB: ok. Alternatively, in this case, we normally define it directly in the host block. But both is fine :)

TomMenga: I'd prefer that too. I did not do it that way because it's slightly redundant and in general, we try to be very attentive to efficiency.
But I believe I'm overthinking as usual :|

mcilurzo (approved):

PR #2607, opened 2024-04-22
fix(sbb-image): fix alt attribute and provide css var for aspect-ratio

mbleuer:
## Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/main/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

## Issue
There's no connected lyne-components issue.

## Pull request checklist

## Changes

Changes in this pull request:

- Make sure that the `alt` attribute is always present on `img` elements, even if there is no alt text. Otherwise, supporting technologies may be tempted to interpret the image.
- If the `aspectRatio` value is set to `free`, the component sets the CSS declaration to `aspect-ratio: var(--image-aspect-ratio, auto);` instead of `aspect-ratio: auto;`. This would allow consumers to enter the value for the CSS variable `--image-aspect-ratio` from outside the component. This would make it possible to set a different aspect ratio from viewport to viewport. People tried to achieve that behaviour with a workaround by defining two separate image components as can be seen here: https://www.sbbinsurance.li/ (inspect the markup of the hero image). This little fix would allow a lot of additional flexibility without having to break any existing functionality.

## Browsers

- [ ] Firefox Desktop
- [x] Chrome Desktop
- [ ] Edge Desktop
- [ ] Safari Desktop
- [ ] Chrome Mobile
- [ ] Safari Mobile

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/components/image/image.ts
```diff
@@ -501,7 +507,7 @@ export class SbbImageElement extends LitElement {
               ];
             })}
             <img
-              alt=${this.alt || nothing}
+              alt=${this.alt || ''}
```
jeripeierSBB, line 510: Can you add a short comment before this line on why it is necessary in this case? thank you :)

src/components/image/image.ts
```diff
@@ -275,6 +275,12 @@ export class SbbImageElement extends LitElement {
    * Set an aspect ratio
    * default is '16-9' (16/9)
    * other values: 'free', '1-1', '1-2', '2-1', '2-3', '3-2', '3-4', '4-3', '4-5', '5-4', '9-16'
+   * If the aspectRatio value is set to 'free', the component sets the CSS declaration to
```
jeripeierSBB, line 278: As we soon will release version 1.0, I think we should "sbb-"-prefix the very last not prefixed CSS var in the project :) this is causing a breaking change but it is ok for now. Furthermore, you could add a specific entry in the jsdoc section of the SbbImageElement class with "@cssprop":
```
@cssprop [--sbb-image-aspect-ratio=auto] - TBD...
```

With that it appears in readme.md and is therefore well documented and part of the public API.

see e.g. logo component for an example

jeripeierSBB (changes requested):
Thanks for your contribution. I think, as mentioned, we should rename the variable with a breaking change.

src/components/image/image.ts
mbleuer: Sure, I just added the explanation.

jeripeierSBB: Thanks, nicely written.

src/components/image/image.ts
mbleuer: Thanks for the hint, I have indeed overlooked the documentation by @cssprop. It's added now. 

I also renamed the variable. Since it is freshly introduced, it should not cause any breaking change? Or are you referring to --image-border-radius?

mbleuer:
> Thanks for your contribution. I think, as mentioned, we should rename the variable with a breaking change.

Sure thing, thanks for the quick response! Please see the separate comment regarding the renaming/breaking change.

src/components/image/image.ts
jeripeierSBB: yes, I just realized it during re-review :D Probably I will create a follow-up PR for that. BTW, the image is, besides some minimal refactorings, still in the state you guys implemented it :)

jeripeierSBB (approved):
LGTM, thank you

PR #2611, opened 2024-04-23
fix(sbb-container): fix background color for nested containers

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
Thanks for the comment 😄

PR #2615, opened 2024-04-24
feat: split css outputs into smaller modules

### LINKED ISSUE #2601: [Feature request]: Refactor typography.css to Focus on Essential Component Styles and Improve File Structure
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/main/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/main/CODE_OF_CONDUCT.md) that this project adheres to.

### Request type

Components

### Functionality

Hi!

I've been using the lyne-components library and noticed that the `typography.css` file, which is required before using the components, does not solely focus on typography. This file contains not only the styles necessary for the components but also global styles for setting text sizes, creating grids, and formatting lists. It also includes parts of `normalize.css`.

I believe that the current `typography.css` should be refocused solely on the styles necessary for the components. A more appropriate name for this refocused file could be `core.css`. This change would make the purpose of the file clearer and its scope more defined.

Additionally, it would be beneficial to separate the other concerns into their own CSS files, such as:
- `normalize.css` for CSS resets
- `layout.css` for grid systems and layout related styles
- etc.

These files can then be included by users as needed, providing more flexibility and clarity in what each file is responsible for.

Thank you for considering this suggestion. I believe these changes would enhance the modularity and usability of the lyne-components library.

_No response_

### LINKED ISSUE #2614: Evaluate merging font faces with font-weight

jeripeierSBB:
Closes #2601 

BREAKING CHANGE: Renamed several files.
- Renamed `typography.css` to `standard-theme.css`. Notice: there is still a file called `typography.css` that contains only a few typography-related outputs. Therefore, **import adaption from `typography.css` to `standard-theme.css` is mandatory**.
- Changed the name of `fullfont.css` to `font-characters-extension.css`.


kyubisation (approved):
LGTM 👍 
I wonder if "global.css" is the best name. In the future, we might have a separate files for e.g. off-brand or lean context.
An option might be `standard-theme.css`.
What do you think?

PR #2624, opened 2024-04-26
build: implement visual regression testing with @web/test-runner

kyubisation:
This PR sets up the basic configuration for visual regression testing with `@web/test-runner`.

`web-test-runner.config.js` contains a new test group, which is only activated when explicitly called with `yarn test:visual-regression`. This test group introduces a new file pattern `*.snapshot.spec.ts`, which contains the visual regression tests.
Additionally, there is a new GitHub Actions workflow, which runs the visual regression tests and saves failed screenshots in an artifact. This artifact is then downloaded in the secure workflow, which add an `in_progress` check, if any failed screenshots exist.

We do not want to save the snapshot png in the repository, as this would bloat it considerably. Due to this, we use containers as an external storage of the png files. The process works as follows:
Each continuous integration run on our main branch creates/updates a container `baseline` (https://github.com/lyne-design-system/lyne-components/pkgs/container/lyne-components%2Fvisual-regression). For any other CI run (in the secure context) this container is used as a service and the png files are downloaded when needed.
Additionally, the baseline image is hosted at `https://lyne-visual-regression-baseline.app.sbb.ch/` in order to enable local visual regression testing.

As for local visual regression testing; To avoid diffs between operating system browser differences, visual regression testing is only directly run in a Linux environment. If it is executed in any other OS, it will try to run it in a container to prevent OS differences showing up in the snapshots.


TomMenga (commented):
Some elements are still obscure but I have a general idea of the implementation. 
I'm not sure how the remote baseline is used and why we need to use a container to run the diff locally (but I'm sure there is a reason)

Some discussion topics:
- I'm a bit scared of the exponential nature of the number of `cases` (`P1 * P2 * ... * Viewport * browsers * 3 (default, focus & hover)`)
  - Pro: covers almost every use case
  - Con: grows very quickly and requires an "implementation" for each component
- Alternatively, we could reuse the stories template? 
  - Pro: we would avoid most of the `.snapshot.spec` code
  - Con: not exhaustive like combining each case

web-test-runner.config.js
```diff
@@ -80,22 +91,34 @@ const suppressedLogs = [
   '[vite] connecting...',
 ];
 
+const groups = [
+  // Disable ssr tests until stabilized.
+  // { name: 'e2e-ssr-hydrated', files: 'src/**/*.e2e.ts', testRunnerHtml },
+  // { name: 'e2e-ssr-non-hydrated', files: 'src/**/*.e2e.ts', testRunnerHtml },
+];
+
+// The visual regression test group is only added when expicitely set, as the tests are very expensive.
```
jeripeierSBB, line 100:
```suggestion
// The visual regression test group is only added when explicitly set, as the tests are very expensive.
```

tools/visual-regression-testing/diff-app/src/screenshots.ts
```diff
@@ -0,0 +1,177 @@
+// eslint-disable-next-line import-x/no-unresolved
+import { screenshotsRaw } from 'virtual:screenshots';
+
+import type { FailedFiles } from '../vite.config.js';
```
kyubisation, line 4: Maybe extract to own file?

tools/visual-regression-testing/diff-app/src/screenshots.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const viewportOrder = ['zero', 'micro', 'small', 'medium', 'large', 'wide', 'ultra'];
+
+// TODO: discuss whether to include it in creation of screenshotsRaw
+const extractHierarchicalMap = (
```
kyubisation, line 9: Move into vite.config.ts

tools/visual-regression-testing/diff-app/src/main.ts
```diff
@@ -0,0 +1,45 @@
...
+        return true;
+      },
+    },
+    {
+      path: '/compare/:component/:testcase',
+      render: ({ component, testcase }) =>
+        html`<app-test-case
+          .params=${{ componentName: component, testCaseName: testcase }}
+        ></app-test-case>`,
+
```
kyubisation, line 27:
```suggestion
```

kyubisation:
> Some elements are still obscure but I have a general idea of the implementation. I'm not sure how the remote baseline is used and why we need to use a container to run the diff locally (but I'm sure there is a reason)
> 
> Some discussion topics:
> 
> * I'm a bit scared of the exponential nature of the number of `cases` (`P1 * P2 * ... * Viewport * browsers * 3 (default, focus & hover)`)
>   
>   * Pro: covers almost every use case
>   * Con: grows very quickly and requires an "implementation" for each component
> * Alternatively, we could reuse the stories template?
>   
>   * Pro: we would avoid most of the `.snapshot.spec` code
>   * Con: not exhaustive like combining each case

Thanks for the review and feedback!
We have reduced the amount of snapshots by being a bit more selective. 
Additionally I updated the PR comment to better describe the container use cases.
I think it makes sense to reuse story templates where it is reasonable, but I assume it will depend on the case.

PR #2626, opened 2024-04-30
fix: avoid minifying identifiers to support SWC/next.js minifier

jeripeierSBB:


TomMenga (approved):

kyubisation:
I have reservations about this. This is fine, if there are no possible workarounds, but we will challenge this first.

PR #2628, opened 2024-05-01
fix(sbb-tab-group): avoid unwanted margin block spaces

### LINKED ISSUE #2627: [Bug]: screen-reader-only escapes overflow:hidden in tab-group
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/lyne-design-system/lyne-components/blob/main/docs/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/lyne-design-system/lyne-components/blob/main/docs/CODE_OF_CONDUCT.md) that this project adheres to.

### Bug type

Functionality

Components

0.50.2

### Operating system

Windows 10

### Browser / Browser version

Google Chrome 124.0.6367.61 (Official Build) (64-bit)

### Input mode

None

### Additional settings

_No response_

### What happened?

Here is a simple reproduction: https://stackblitz.com/edit/nickzn-lyne-screen-reader-only-issue?file=src%2Fmain.html

When using a component which contains a screen-reader-only element, block-link for example, inside a tab-group in the second tab it creates a white-space at the bottom of the first tab.

It would be an easy fix for us by just adding `position: relative` to the block-link, but we think this should be fixed on the library.

-

### Relevant log output

_No response_

jeripeierSBB:
Closes #2627

### DIFF
--- a/src/components/container/container/container.scss
+++ b/src/components/container/container/container.scss
@@ -16,6 +16,8 @@
 }
 
 .sbb-container {
+  @include sbb.ignore-children-margin;
+
   background-color: var(--sbb-container-background-color);
 
   // In order to support nested sbb-containers (e.g. if used in an overlay), we need to explicitly set background color to transparent.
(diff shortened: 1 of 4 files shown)


dauriamarco (approved):
Thank you for the fix! LGTM 👍

PR #2641, opened 2024-05-07
feat: add root entry point with global component registration

kyubisation:
This feature allows using `import '@sbb-esta/lyne-components'`, which will import all components and add them to `globalThis`, but will not provide an export. This is useful for SSR or for providing examples/reproductions and wanting to load all components.

### DIFF
--- a/tools/vite/generate-root-entry-point.ts
+++ b/tools/vite/generate-root-entry-point.ts
@@ -1,0 +1,47 @@
+import type { LibraryOptions, PluginOption, ResolvedConfig } from 'vite';
+
+export function generateRootEntryPoint(): PluginOption {
+  let viteConfig: ResolvedConfig;
+  return {
+    name: 'package-json-templating',
+    configResolved(config) {
+      viteConfig = config;
+    },
+    async generateBundle() {
+      if (viteConfig.command !== 'build') {
+        return;
+      }
+      const entry = (viteConfig.build.lib as LibraryOptions).entry as Record<string, string>;
+      const classNameMap: Record<string, string> = {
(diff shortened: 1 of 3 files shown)


tools/vite/generate-root-entry-point.ts
```diff
@@ -0,0 +1,47 @@
...
+    },
+    async generateBundle() {
+      if (viteConfig.command !== 'build') {
+        return;
+      }
+      const entry = (viteConfig.build.lib as LibraryOptions).entry as Record<string, string>;
+      const classNameMap: Record<string, string> = {
+        SbbOptgroupElement: 'SbbOptGroupElement',
+      };
+      const toElementName = (key: string) => {
```
jeripeierSBB, line 18: according to warning of lint, add return type?

tools/vite/generate-root-entry-point.ts
```diff
@@ -0,0 +1,47 @@
...
+    configResolved(config) {
+      viteConfig = config;
+    },
+    async generateBundle() {
+      if (viteConfig.command !== 'build') {
+        return;
+      }
+      const entry = (viteConfig.build.lib as LibraryOptions).entry as Record<string, string>;
+      const classNameMap: Record<string, string> = {
+        SbbOptgroupElement: 'SbbOptGroupElement',
```
jeripeierSBB, line 16: why do we need this translation here?

kyubisation: Because it is the only file that does not match the casing of the file name
`optgroup.ts` => `SbbOptGroupElement`
Usually I infer the class name from the file name, but this does not work for this specific file

PR #2642, opened 2024-05-07
refactor: use typed PropertyValues and always call super method of willChange

jeripeierSBB:


kyubisation (approved):
LGTM 👍 
Thanks!
Looking at this, we should probably have a linting rule to cover these cases.

PR #2644, opened 2024-05-07
fix: fix super calls of lit element lifecycle methods

jeripeierSBB:


tools/eslint/needs-super-call-rule.ts
```diff
@@ -0,0 +1,92 @@
...
+const methodsToCheckForSuperCall = [
+  'connectedCallback',
+  'disconnectedCallback',
+  'requestUpdate',
+  'performUpdate',
+  'shouldUpdate',
+  'willUpdate',
+  'update',
+  'firstUpdated',
+  'updated',
```
kyubisation, line 20: Nitpick; You could spread hasChangedProperties here.

kyubisation (approved):
LGTM 👍 
Cool! 😃

PR #2679, opened 2024-05-21
build: improve test runner concurrency

kyubisation:
By using streaming in our vite to web-test-runner proxy, we can improve the concurrency capabilities of running tests.


jeripeierSBB (approved):
LGTM :)

PR #2692, opened 2024-05-27
fix(sbb-radio-button): fix checked initialization

jeripeierSBB:


TomMenga (approved):
LGTM 👍 
Given that I agree with the component changes, would the react wrapper change have solved the issue?

jeripeierSBB:
> LGTM 👍 Given that I agree with the component changes, would the react wrapper change have solved the issue?

Yes, but not using onwillopen hook was necessary too. But as you said, as well, being more robust in our implementation is a good idea as well :)

kyubisation (approved):
LGTM 👍

PR #2694, opened 2024-05-27
test: test visual regression focus without waiting

jeripeierSBB:


tools/web-test-runner/visual-regression-plugin-config.js
```diff
@@ -60,4 +60,7 @@ export const visualRegressionConfig = (update) =>
         'utf8',
       );
     },
+    diffOptions: {
+      includeAA: false,
```
kyubisation, line 64: I think this is false by default?

kyubisation (approved):
LGTM 👍

tools/web-test-runner/visual-regression-plugin-config.js
```diff
@@ -60,4 +60,7 @@ export const visualRegressionConfig = (update) =>
...
+    },
```
kyubisation, line 65:
```suggestion
```

PR #2705, opened 2024-05-30
fix(sbb-block-link): always show underline except for footer links

### LINKED ISSUE #2700: bug: due to accessibility reasons links should always have an underline
The block link should receive an underline.
There is only one exception: Links in a linklist inside the footer should not receive an underline.

dauriamarco:
Closes #2700


src/elements/link/common/link.scss
```diff
@@ -7,6 +7,10 @@
   // Use !important here to not interfere with Firefox focus ring definition
   // which appears in normalize css of several frameworks.
   outline: none !important;
+
+  @include sbb.if-forced-colors {
```
jeripeierSBB, line 11: This should go into the link-base mixin

src/elements/core/styles/mixins/link.scss
```diff
@@ -6,11 +6,7 @@
   --sbb-link-color-normal: var(--sbb-color-granite);
   --sbb-link-color-hover: var(--sbb-color-red125);
   --sbb-link-color-active: var(--sbb-color-red150);
-  --sbb-link-text-decoration: none;
-
-  @include a11y.if-forced-colors {
-    --sbb-link-text-decoration: underline;
-  }
+  --sbb-link-text-decoration: var(--sbb-link-text-decoration-none, underline);
```
jeripeierSBB, line 9: I still don't get why we need this extra variable?

jeripeierSBB (changes requested):

mcilurzo (approved):

src/elements/core/styles/mixins/link.scss
dauriamarco: I discussed this later with Lukas and the problem is that if I override the `--sbb-link-text-decoration` from the footer it doesn't work, so I need the second var with the fallback approach.

jeripeierSBB: Shouldn't we remove "--sbb-link-text-decoration" here and apply it as a fallback where it is currently used?

jeripeierSBB: (to avoid duplicating var)

jeripeierSBB (changes requested):
Breadcrumb are inheriting from sbb-link-base, but should not be affected by underline. Could you fix this? Otherwise looks good

jeripeierSBB (approved):
LGTM, thank you

PR #2715, opened 2024-06-05
feat(sbb-table-wrapper): initial implementation

### LINKED ISSUE #2670: story(table): implement design and wrapper
### Feature Description

We want to implement the table design and wrapper component.
NOTE: We want to reuse the native table and due to this, the styles for the table must be global.

Related issue: #2038

Figma: https://www.figma.com/design/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?m=auto&node-id=10643%3A325077&t=dcna1LTlbn0zOlPn-1
Reference (for wrapper): https://angular.app.sbb.ch/angular/components/table/overview

Design Spec:

- The table can have two sizes: `s` and `m` (default)
- The table can be negative

Tech Spec:

- Create css classes `sbb-table`, `sbb-table-s` and `sbb-table-m` (`sbb-table` and `sbb-table-m` are currently the same, but in the future we might provide a different theme, where `sbb-table` and `sbb-table-s` are the same)
  - This should style all inner elements (e.g. `<tr>`, `<th>`, `<td>`, `<caption>`, etc.)
- The table is striped per default, but can be turned off by using the css class `sbb-table-unstriped` (To discuss: naming)
  This is to enable e.g. collapsible rows, which would not work by even/odd stripes.
- Add a css class `sbb-table-row-striped` to manually apply the striped gray background to a tr element.
- Create css class `sbb-table-negative`, which applies the negative design to the table when added to any other `sbb-table` css class.
- For each css class, also provide a sass mixin which can be used by our consumers.

`sbb-table-wrapper`:
- Create component `sbb-table-wrapper` in module `table` (To discuss: Acceptable?)
- Should have the styles `display: block; overflow: auto;`
- Has no functionality (yet)

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

TomMenga:
## Preflight Checklist

## Issue

This PR Closes #2670

## Pull request checklist

See [Review Guidelines](../docs/REVIEW.md) for more information what is checked during review process.

## Changes

Changes in this pull request:

- Table styles and mixins
- `sbb-table-wrapper` implementation

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [ ] Edge Desktop
- [x] Safari Desktop
- [ ] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

- [ ] JAWS Firefox Desktop
- [ ] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [ ] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/elements/table/table-wrapper/readme.md
```diff
@@ -0,0 +1,28 @@
+The `sbb-table-wrapper` is a wrapper for a table. Its aim is to enhance/automate tha native table capabilities.
```
DavideMininni-Fincons, line 1:
```suggestion
The `sbb-table-wrapper` is a wrapper for a `<table>`; its goal is to enhance/automate the native table capabilities.
```

src/elements/table/table-wrapper/readme.md
```diff
@@ -0,0 +1,28 @@
...
+
+As of now, it only handles overflow (vertical and horizontal).
```
DavideMininni-Fincons, line 3:
```suggestion
Currently it only handles overflow (vertical and horizontal).
```

src/elements/table/table-wrapper/table-wrapper.stories.ts
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const defaultArgTypes: ArgTypes = {
+  negative: negative,
+  size: size,
+};
+
+const defaultArgs: Args = {
+  negative: false,
+  size: 'm',
```
DavideMininni-Fincons, line 29:
```suggestion
  size: size!.options[1],
```

src/elements/table/table-wrapper/table-wrapper.stories.ts
```diff
@@ -0,0 +1,115 @@
...
+const size: InputType = {
+  control: {
+    type: 'inline-radio',
+  },
+  options: ['s', 'm'],
+};
+
+const defaultArgTypes: ArgTypes = {
+  negative: negative,
+  size: size,
```
DavideMininni-Fincons, line 24:
```suggestion
  negative,
  size,
```

src/storybook/styles/table/table.stories.ts
```diff
@@ -0,0 +1,134 @@
...
+const striped: InputType = {
+  control: {
+    type: 'boolean',
+  },
+};
+
+const defaultArgTypes: ArgTypes = {
+  size: size,
+  negative: negative,
+  striped: striped,
```
DavideMininni-Fincons, line 31:
```suggestion
  size,
  negative,
  striped,
```

src/storybook/styles/table/table.stories.ts
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const defaultArgs: Args = {
+  size: 'm',
```
DavideMininni-Fincons, line 35:
```suggestion
  size: size!.options[1],
```

DavideMininni-Fincons (changes requested):
Looks good! some minor changes

TomMenga:
> Generally LGTM 👍 Some open topics:
> 
> * Should the default width of the table be 100%? This was previously the case in the sbb-angular library. Thoughts @TomMenga, @jeripeierSBB, @mcilurzo?
> * The shadow logic can be compared with the sbb-angular implementation. Basically add a scroll listener to the wrapper and use the resize event from the window to calculate shadow necessity. https://github.com/sbb-design-systems/sbb-angular/blob/main/src/angular/table/table/table-wrapper.ts
> * Maybe for the wrapper story, add more columns to demonstrate the responsiveness.

I'll reply point by point:
* In the first implementation I set the width to 100% but then I thought that the table layout (whether fixed or auto) and size should be in charge of the consumer. I'm not against having a default 100% width though.
* It's not the same, unfortunately. We do not have a problem understanding when to show the shadow (we already do that in the sticky header for example). The problem stands in the shadow effect itself since we do not have any element that could cast that shadow from the table borders. The angular wrapper only has a shadow effect on fixed columns (because the column itself can cast it). I've tried different approaches but none of them worked 100%.
The closest thing I found was by simulating a shadow using backgrounds ([Codepen](https://codepen.io/ramiro-ruiz/pen/YzKVOGK)). The drawback is that the appearance is bound to the scroll and that's not an actual shadow. Maybe we can deep down on it in a call
* On it 👍

src/storybook/styles/table/readme.md
```diff
@@ -0,0 +1,72 @@
...
+  </caption>
+</table>
+```
+
+### Striped table
+
+By default, tables are styled with alternating row stripes to enhance readability. To remove this striping, add the `sbb-table-unstriped` class:
+
+```html
+<table class="sbb-table sbb-table-unstriped">
```
jeripeierSBB, line 30: I'm not sure what is best, I only remember for the sbb-text--bold, we used '--' to show that it can't be used standalone and is only a modifier of the class. (this comment would also apply to the sizes and negative) Probably to be discussed.

```suggestion
<table class="sbb-table sbb-table--unstriped">
```

src/storybook/styles/table/table.stories.ts
```diff
@@ -0,0 +1,134 @@
+import type { InputType } from '@storybook/types';
```
jeripeierSBB, line 1: would it make sense to create a visual regression file also here? Like I did for the other folders in the styles folder.

src/storybook/styles/table/table.stories.ts
```diff
@@ -0,0 +1,134 @@
...
+    <tr>
+      <td>KAREN</td>
+      <td>Web performance</td>
+      <td>36</td>
+    </tr>
+  </tbody>
+`;
+
+const footer: () => TemplateResult = () => html`
+  <!-- TODO -->
```
jeripeierSBB, line 82: Something to be removed or implement now? or really planned for somewhen?

src/elements/core/styles/mixins/table.scss
```diff
@@ -0,0 +1,135 @@
...
+/*
+ * SBB table mixin
+ * Notes:
+ * We cannot use `border-collapse` because it is not compatible with the `border-radius` property
+ */
+@mixin table {
+  @include table-striped;
+  @include table-m;
+
+  --sbb-table-border: 1px solid var(--sbb-table-border-color);
```
jeripeierSBB, line 12: Couldn't we use the var?
```suggestion
  --sbb-table-border: var(--sbb-border-width-1x) solid var(--sbb-table-border-color);
```

src/elements/core/styles/mixins/table.scss
```diff
@@ -0,0 +1,135 @@
...
+
+  caption {
+    @include table-caption;
+  }
+}
+
+@mixin table-header-cell {
+  @include t.text-xs--bold;
+
+  border-bottom: var(--sbb-table-border);
```
jeripeierSBB, line 57: Is it possible to use `border-block-end` and `border-inline-end` ? Also for similar following occurrences

src/elements/core/styles/mixins/table.scss
```diff
@@ -0,0 +1,135 @@
+@use './typo' as t;
```
jeripeierSBB, line 1: Nitpick, but normally we use `typo` instead of t.

src/elements/core/styles/mixins/table.scss
```diff
@@ -0,0 +1,135 @@
...
+  tbody tr:nth-child(odd) {
+    background-color: unset;
+  }
+}
+
+@mixin table-row-striped {
+  background-color: var(--sbb-table-row-striped-color);
+}
+
+@mixin table-negative {
```
jeripeierSBB, line 115: Shouldn't we use table--negative here? probably it's not consistent in all our repo so far. Also for following mixins.

src/elements/table/table-wrapper/readme.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+## Style
+
+See the [Table](/docs/styles-table--docs) style section.
+
+The component has a `negative` variant which can be set with the self-named property.
+
+```html
+<sbb-table-wrapper negative>
```
jeripeierSBB, line 25: Could we pass down the negative property to the table so a consumer doesn't need to set it twice?

src/elements/table/table-wrapper/table-wrapper.scss
```diff
@@ -0,0 +1,33 @@
...
+::slotted(.sbb-table) {
+  width: 100%;
+}
+
+.sbb-table-wrapper {
+  overflow: auto;
+  @include sbb.scrollbar($size: thick, $track-visible: true);
+
+  &::-webkit-scrollbar {
+    border-radius: 0 0 var(--sbb-border-radius-4x) var(--sbb-border-radius-4x);
```
jeripeierSBB, line 26: As it is used 4 times, should we not create an addition var for the border-radius-4x? like --sbb-table-wrapper-border-radius?

src/elements/table/table-wrapper/table-wrapper.scss
```diff
@@ -0,0 +1,33 @@
...
+@include sbb.box-sizing;
+
+:host {
+  display: block;
+}
+
+:host([data-has-horizontal-scrollbar]) {
+  ::slotted(.sbb-table) {
+    // Remove the border-radius from the bottom of the table if the scroll is visible
+    --sbb-table-border-radius: var(--sbb-border-radius-4x) var(--sbb-border-radius-4x) 0 0 !important;
```
jeripeierSBB, line 13: could we not use a solution without !important by only overriding the var? maybe as a fallback var? If not, please provide a comment on why !important is needed.

src/elements/table/table-wrapper/table-wrapper.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { assert } from '@open-wc/testing';
+import { html } from 'lit/static-html.js';
+
+import { fixture } from '../../core/testing/private.js';
+
+import { SbbTableWrapperElement } from './table-wrapper.js';
+
+describe(`sbb-table-wrapper with ${fixture.name}`, () => {
```
jeripeierSBB, line 8:
```suggestion
describe(`sbb-table-wrapper`, () => {
```

src/elements/table/table-wrapper/table-wrapper.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+  let element: SbbTableWrapperElement;
+
+  beforeEach(async () => {
+    element = await fixture(html`<sbb-table-wrapper></sbb-table-wrapper>`, {
+      modules: ['./table-wrapper.ts'],
+    });
```
jeripeierSBB, line 14:
```suggestion
    element = await fixture(html`<sbb-table-wrapper></sbb-table-wrapper>`);
```

src/elements/table/table-wrapper/table-wrapper.visual.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  const cases = {
+    size: ['m', 's'],
+    negative: [false, true],
+    striped: [false, true],
+    // TODO add scrollbar cases
+  };
+
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    describeEach(cases, ({ size, negative, striped }) => {
```
jeripeierSBB, line 24: In my eyes, the tests of the size could be separate to negative/striped. the size is not influenced by negative nor striped. To save snapshots.

jeripeierSBB (changes requested):
Nice work, some discussion points

src/storybook/styles/table/readme.md
TomMenga: I think I agree with you. I only see an inconsistency in the size classes since the text ones are like `sbb-text-${size}`.
I'll change negative and stripe for now, we can discuss the size eventually

src/elements/table/table-wrapper/table-wrapper.visual.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+describe(`sbb-table-wrapper`, () => {
+  let root: HTMLElement;
+
+  const cases = {
+    negative: [false, true],
+    scrollbar: [false, true],
+  };
+
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
```
jeripeierSBB, line 21: Maybe there was a little misunderstanding, I did not mean to test striped or the size not at all. I just thought it could be in a separate test case, so it doesn't generate each combination.

jeripeierSBB: Ah you have pushed separate test in styles folder. But there the sizes are combined as well with every other case?

src/elements/table/table-wrapper/readme.md
TomMenga: I thought we decided not to do that during our call but maybe we should rediscuss it.
Essentially we have three options here:
1. No automation (the consumer sets the negative property and the class)
2. The wrapper checks for the negative class on the table and syncs itself (I don't think we have a consistent way to react to an eventual class change though)
3. The wrapper sets the negative class on the table (The property has priority)

I believe none of them is optimal, maybe the 3rd is best :/
What do you think? @jeripeierSBB @kyubisation

src/elements/table/table-wrapper/table-wrapper.scss
TomMenga: I'm not happy with it but I believe it is the less evil solution. I'll add a comment on why it's needed

src/elements/table/table-wrapper/table-wrapper.spec.ts
TomMenga: To be sure, we need to explicitly declare modules with the SSR fixture, correct?

src/elements/table/table-wrapper/readme.md
kyubisation: Technically we could add the selector `sbb-table-wrapper[negative]` (or maybe `sbb-table-wrapper[negative] .sbb-table`) to the selector `.sbb-table-negative`.

jeripeierSBB: But this would still be only top-down and not bottom-up. But yeah it's a simple solution for the top down approach ;)

src/elements/table/table-wrapper/table-wrapper.spec.ts
jeripeierSBB: yes but only for ssr in the ssr file. in all other spec files we don't need it.

src/elements/table/table-wrapper/readme.md
TomMenga: I like it. Top-down is covered. 
Do we do something for the bottom-up? We could check for the negative class in a @slotchange ( even though it does not cover every case )

src/storybook/styles/table/readme.md
jeripeierSBB: yeah but the sbb-text sizes do include all the rules, so it could be used standalone. We could also have this approach in table, but would generate more output

src/elements/core/styles/mixins/table.scss
jeripeierSBB: I agree, here it doesn't matter. I was just thinking at our base rules to everywhere use "modern" properties. In general I'd like to not have a border-radius and user border-collapse :D... But let's start with how it was designed and we will see how far we will bring it with all future table features.

src/storybook/styles/table/readme.md
TomMenga: Mmm ok. So we do not have an inconsistency if we change the table size classes to `sbb-table--*size*`.
I'd rather avoid making them standalone because I expect that the table CSS will grow in size with the advanced features

src/storybook/styles/table/readme.md
```diff
@@ -0,0 +1,72 @@
...
+
+```html
+<table class="sbb-table sbb-table--unstriped">
+  ...
+</table>
+```
+
+### Size
+
+The available sizes are `m` (default) and `s`. Use the respective `sbb-table-*size*` classes to specify the size:
```
jeripeierSBB, line 37:
```suggestion
The available sizes are `m` (default) and `s`. Use the respective `sbb-table--*size*` classes to specify the size:
```

src/storybook/styles/table/readme.md
```diff
@@ -0,0 +1,72 @@
+The `sbb-table` class is designed to handle the styling of common table use cases, ensuring a consistent and visually appealing presentation of tabular data.
```
jeripeierSBB, line 1: I think we should mention the wrapper here and have a link to it.

src/storybook/styles/table/table.visual.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      beforeEach(async function () {
+        root = await visualRegressionFixture(
+          tableTemplate({
+            'sbb-table': true,
+            'sbb-table--negative': negative,
+            'sbb-table--unstriped': !striped,
+          }),
+          {
+            backgroundColor: negative ? 'var(--sbb-color-black)' : undefined,
+            focusOutlineDark: negative,
```
jeripeierSBB, line 62: nitpick, as there is nothing to focus.
```suggestion
```

src/storybook/styles/table/table.visual.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+        visualDiffDefault.with(async (setup) => {
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+
+    // Size cases
+    for (const size of sizeCases.size) {
+      it(
+        `size=${size} ${visualDiffDefault.name}`,
```
jeripeierSBB, line 78: the simpler, the better I think. I did this simliar for simliar cases
```suggestion
        `size=${size}`,
```

src/storybook/styles/table/table.visual.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+      it(
+        visualDiffDefault.name,
+        visualDiffDefault.with(async (setup) => {
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+
+    // Size cases
+    for (const size of sizeCases.size) {
```
jeripeierSBB, line 76: if m is tested as default, we could even save the for loop and only have one test for size s.

src/elements/table/table-wrapper/table-wrapper.visual.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+                    <td>45</td>
+                  </tr>
+                </tbody>
+              </table>
+            </sbb-table-wrapper>
+            <p class="sbb-table-caption">Table caption</p>
+          `,
+          {
+            backgroundColor: negative ? 'var(--sbb-color-black)' : undefined,
+            focusOutlineDark: negative,
```
jeripeierSBB, line 59:
```suggestion
```

src/elements/core/styles/table.scss
```diff
@@ -0,0 +1,46 @@
+@use './mixins/table' as t;
```
jeripeierSBB, line 1: maybe to be consistent use
```suggestion
@use './mixins/table';
```
and then table.*

jeripeierSBB (commented):
Almost done ;-) thanks!

src/elements/core/styles/mixins/table.scss
```diff
@@ -0,0 +1,136 @@
...
+      @include table-data-row;
+
+      & > td {
+        @include table-data-cell;
+      }
+    }
+  }
+
+  tfoot {
+    // TBD
```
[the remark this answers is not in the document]
TomMenga: Since I removed the other footer TODO, I'll remove this too

src/elements/table/table-wrapper/table-wrapper.scss
```diff
@@ -0,0 +1,36 @@
...
+:host {
+  --sbb-table-wrapper-border-radius: var(--sbb-border-radius-4x);
+
+  display: block;
+}
+
+:host([data-has-horizontal-scrollbar]) {
+  ::slotted(.sbb-table) {
+    // Remove the border-radius from the bottom of the table if the scroll is visible (!important is needed because ::slotted has the lower specificity)
+    --sbb-table-border-radius: var(--sbb-table-wrapper-border-radius)
```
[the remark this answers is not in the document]
TomMenga: Definitely better 👍

src/storybook/styles/table/readme.md
kyubisation: The specification intended for sbb-table, sbb-table-s and sbb-table-m to be usable standalone.
I.e. you could use `sbb-table` or `sbb-table-m`.
But I'm open for discussion.

src/storybook/styles/table/table.visual.spec.ts
TomMenga: I've slightly modified the test and now it has some value.

src/storybook/styles/table/readme.md
TomMenga: I've misinterpreted the specification :/
In terms of output, it doesn't change much to have them standalone (if we implement them like this)

Note: Bootstrap combines them 

image

jeripeierSBB:
If you wonder why tests are failing, it seems that the spec file was formatted wrongly somehow. new generation of it could solve it.

TomMenga:
> If you wonder why tests are failing, it seems that the spec file was formatted wrongly somehow. new generation of it could solve it.

yeah :/ thanks

src/elements/table/table-wrapper/table-wrapper.scss
```diff
@@ -0,0 +1,36 @@
+@use '../../core/styles/index' as sbb;
```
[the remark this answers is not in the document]
TomMenga: Apparently, scrollbar styling is not supported anymore since iOS 14. 
The sad thing is that `@supports` still returns positive so we don't have a way to discern which OS we are on :|
I think the time has come where we ask to drop the border-radius on the table 😞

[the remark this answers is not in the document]
TomMenga: We have the same issue on Firefox :/

src/elements/table/table-wrapper/readme.md
kyubisation: I wonder if we need the bottom-up case. We could also just document when you are using the sbb-table-wrapper, you need to consider the correct usage.

jeripeierSBB: I think documentation is a good starting point :) But as written, I'm almost certain, that some projects will forget to manually set `negative` and nobody will realize it because the outcome is not obviously different. But for sure, we could accept it or start with documentation only and if we see a lot of misusage, we could provide additional logic at a later time.

src/elements/table/table-wrapper/table-wrapper.ts
```diff
@@ -0,0 +1,57 @@
...
+  private _resizeObserver = new AgnosticResizeObserver(() => this._checkHorizontalScrollbar());
+  private _tableWrapper!: HTMLElement;
+
+  public override disconnectedCallback(): void {
+    super.disconnectedCallback();
+    this._resizeObserver.disconnect();
+  }
+
+  protected override firstUpdated(_changedProperties: PropertyValues): void {
+    super.firstUpdated(_changedProperties);
```
kyubisation, line 33: Nitpick
```suggestion
  protected override firstUpdated(changedProperties: PropertyValues): void {
    super.firstUpdated(changedProperties);
```

kyubisation (approved):
LGTM 👍 
Nice work!

PR #2718, opened 2024-06-05
refactor: move css logic and avoid duplications

jeripeierSBB:


src/elements/core/mixins/panel-mixin.ts
```diff
@@ -24,33 +21,6 @@ export const SbbPanelMixin = <T extends AbstractConstructor<LitElement>>(
 
     /** Whether the unselected panel has a border. */
     @property({ reflect: true, type: Boolean }) public borderless = false;
-
```
MarioCastigliano, line 27: Is this mixin still needed? Given it's providing just two properties it's starting to look a bit like over-engineering. What do you think? 🤔

src/elements/selection-expansion-panel/selection-expansion-panel.ts
```diff
@@ -19,6 +19,10 @@ import '../divider.js';
 
 import style from './selection-expansion-panel.scss?lit&inline';
 
+const isSelectionPanelInput = (input: HTMLElement): boolean =>
+  !!input.closest('sbb-selection-expansion-panel') &&
+  !input.closest?.('sbb-selection-expansion-panel [slot="content"]');
+
```
MarioCastigliano, line 25: If the purpose of this function is just to distinguish the events coming from the unnamed slot from those coming from the `content` slot I think it's okay to remove this entire logic (since panel components are not allowed inside the `content` slot). And to foolproof it we can listen to the events on the unnamed slot anyway (instead of the host), so that any event coming from the content slot just won't trigger the listener. Don't you think? (hope this was clear)

MarioCastigliano (commented):
Everything seems to be working/looking as it should to me, just a couple of thoughts

src/elements/core/mixins/panel-mixin.ts
jeripeierSBB: I think yes, also because we have the corresponding style file. Could you live with keeping it? ;)

MarioCastigliano: Yep, it makes sense on a structural level

MarioCastigliano (approved):
LGTM 👍

PR #2726, opened 2024-06-07
docs(sbb-stepper): improve stories

DavideMininni-Fincons:


kyubisation (approved):
LGTM 👍 
Thanks! 😃

jeripeierSBB (approved):
LGTM thank you :)

dauriamarco (changes requested):
Thank you! A comment on the step content and the length of the labels: they need to be different lengths to correctly test all step change animations, in the horizontal variants the content should be of a different number of lines to see the change in height and the same for long labels in the vertical variant to correctly test the marker line animation.

dauriamarco (approved):
LGTM 👍

PR #2727, opened 2024-06-07
build: mangle private properties in prod build to reduce size

kyubisation:
This PR adds a configuration to mangle private properties in prod builds. Currently this saves about 40kB (of about 800kB total).
This technically also prevents consumers from using private API, as it will be reduced to a single letter usually.

Personal opinion: While this does not have a huge impact, I still feel like we should try to optimize where possible. But I would like to hear your opinions on this @jeripeierSBB, @TomMenga.

### DIFF
--- a/src/elements-experimental/vite.config.ts
+++ b/src/elements-experimental/vite.config.ts
@@ -46,6 +46,7 @@ export default defineConfig((config) =>
           ]
         : []),
     ],
+    esbuild: isProdBuild(config) ? { mangleProps: /^_/ } : {},
     build: {
       cssMinify: isProdBuild(config),
       lib: {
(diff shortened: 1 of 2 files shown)


TomMenga:
> This PR adds a configuration to mangle private properties in prod builds. Currently this saves about 40kB (of about 800kB total). This technically also prevents consumers from using private API, as it will be reduced to a single letter usually.
> 
> Personal opinion: While this does not have a huge impact, I still feel like we should try to optimize where possible. But I would like to hear your opinions on this @jeripeierSBB, @TomMenga.

Generally, I'm a conservative developer but a 5% in size reduction is worth imho.
The [esBuild docs](https://esbuild.github.io/api/#mangle-props) warns about subtle problems but I cannot see them in our case (hopefully)

image

PR #2736, opened 2024-06-11
fix: fix stacked overlays inert mechanism

dauriamarco:


jeripeierSBB (approved):
Probably we should have a method for it as line 45-48 and 98-101 seem duplicated.

PR #2737, opened 2024-06-11
test(sbb-image): migrate visual regression tests

jeripeierSBB:


TomMenga (approved):

src/elements/image/image.visual.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+      '16-9',
+    ],
+  };
+
+  const additionalBorderRadiusCases = {
+    borderRadius: ['none', 'round'],
+  };
+
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    describeEach(aspectRatioCases, ({ aspectRatio }) => {
```
kyubisation, line 35: Nitpick; I'm not sure whether we should use the describeEach for single parameters. We could use just a for loop here with the it here. Thoughts?

DavideMininni-Fincons (approved):

kyubisation (approved):
LGTM 👍

PR #2740, opened 2024-06-11
build: enable visual regression app for Firefox and Safari

jeripeierSBB:


src/visual-regression-app/src/main.ts
```diff
@@ -4,6 +4,11 @@ import { customElement } from 'lit/decorators.js';
 
 import '@sbb-esta/lyne-elements/core/styles/standard-theme.scss';
 
+// @ts-expect-error: Property 'UrlPattern' does not exist
```
kyubisation, line 7: Nitpick; I'm not fond of these escape hatches. Would it be possible to use `!('URLPattern' in globalThis)`?

kyubisation (approved):
LGTM 👍

PR #2751, opened 2024-06-13
test(sbb-accordion, sbb-expansion-panel): add visual test

TomMenga:
Implemented visual spec tests for:
- sbb-accordion
- sbb-expansion-panel
- sbb-expansion-panel-header
- ~~sbb-expansion-panel-content~~ (not needed)


src/elements/expansion-panel/expansion-panel-header/expansion-panel-header.visual.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+          <sbb-expansion-panel-header
+            ?disabled=${disabled}
+            .iconName=${!state.slotted ? state.icon : undefined}
+          >
+            ${state.slotted ? html`<sbb-icon name=${state.icon!}></sbb-icon>` : nothing} Label
+          </sbb-expansion-panel-header>
+        `);
+      });
+
+      for (const state of visualDiffStandardStates) {
```
jeripeierSBB, line 38: In theory there is not active state, so you could also use a subarray without active state. But it's also not so bad. Moreover, wouldn't it be sufficient that this component is already tested in expansion-panel? Or do the tests here provide any additional value? e.g. the focus state does not look correctly when not using inside expansion panel.

src/elements/expansion-panel/expansion-panel/expansion-panel.visual.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+      it(
+        visualDiffDefault.name,
+        visualDiffDefault.with((setup) => {
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+
+    // Title level test
+    for (const titleLevel of titleLevelCases.titleLevel) {
```
jeripeierSBB, line 86: inlining the array would also be ok in my eyes (other tests using inline arrays)

src/elements/expansion-panel/expansion-panel/expansion-panel.visual.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+          >
+            <sbb-expansion-panel-header icon-name="arrow-right-small">
+              Header
+            </sbb-expansion-panel-header>
+            <sbb-expansion-panel-content> Content </sbb-expansion-panel-content>
+          </sbb-expansion-panel>
+        `);
+      });
+
+      for (const state of visualDiffStandardStates) {
```
jeripeierSBB, line 54: active state could be skipped

jeripeierSBB (commented):
Basically looks good, thank you! I think some snapshot could / should be saved -> see comments

src/elements/expansion-panel/expansion-panel-header/expansion-panel-header.visual.spec.ts
TomMenga: I wanted to separate the test responsibilities but maybe that's not the way to go. 
In theory, separating them helps to troubleshoot (you immediately know what component has it). Do you have an opinion?
Ps. I removed the focus tests (the expansion-panel is responsible for it)

src/elements/expansion-panel/expansion-panel/expansion-panel.visual.spec.ts
TomMenga: Isn't it nice to have all the 'combinations' on top of the file? 
I have difficulties skimming through the boilerplate code. Maybe we can discuss it?

src/elements/expansion-panel/expansion-panel-header/expansion-panel-header.visual.spec.ts
jeripeierSBB: Basically, I agree with you - we should test at specific component level :) We can also start like you proposed for now. 
In general, I try to avoid too much noise by e.g. flickering tests (The more tests the more potential problems). And on the other hand, the chance that if in the expansion-header there is a failing test but in expansion-panel tests not is very low, I think. So in most (all?) cases, both will fail. So we would have more diffs to check.

TomMenga: That's also true :/
Ideally, we should try not to overlap the test cases as much as possible.
As you said, the only exclusive cases for the panel-header are the icon combinations
I'm convinced, I'll merge them with the expansion panel

jeripeierSBB (approved):
LGTM, thank you

PR #2776, opened 2024-06-14
fix: avoid breaking type declarations due to alias resolution

kyubisation:
Currently the cross-package import paths are resolved by the dts vite plugin. However, the import paths to our packages should remain untouched. Due to this, we add configuration to prevent our aliases from being converted.


jeripeierSBB (approved):
LGTM, thank you :)

kampLoubna:
shut up

kampLoubna:
bitch

PR #2784, opened 2024-06-17
test(sbb-alert): add visual regression tests

dauriamarco:


src/elements/alert/alert/alert.visual.spec.ts
```diff
@@ -0,0 +1,74 @@
+import type { Args } from '@storybook/web-components';
```
DavideMininni-Fincons, line 1: I like the 'storybook-like' approach here, but on the other hand i wish to discuss if we can consider it as a 'standard'

kyubisation: I think this approach is generally fine, although I would prefer a more specific approach.
i.e. don't use Args here, as it only hides types and don't use sbbSpead, instead just use the attribute/property assignment.
@dauriamarco, @DavideMininni-Fincons What do you think?

dauriamarco: sounds good to me 👍

kyubisation: Quick PoC:
```ts

  const defaultArgs = {
    size: 'm',
    readonly: false,
    icon: 'info',
    titleContent: 'Interruption between Berne and Olten',
    linkContent: undefined as string | undefined,
    href: 'https://www.sbb.ch' as string | undefined,
  };

  const contentSlotText = html`Between Berne and Olten from 03.11.2021 to 05.12.2022 each time from
  22:30 to 06:00 o'clock construction work will take place. You have to expect changed travel times
  and changed connections.`;

  const alertTemplate = ({
    size,
    readonly,
    icon,
    titleContent,
    linkContent,
    href,
  }: typeof defaultArgs): TemplateResult => html`
    <sbb-alert
      size=${size}
      ?readonly=${readonly}
      icon-name=${icon}
      title-content=${titleContent}
      .linkContent=${linkContent}
      .href=${href}
      >${contentSlotText}</sbb-alert
    >
  `;
```

DavideMininni-Fincons (approved):
LGTM

jeripeierSBB (approved):
LGTM

PR #2787, opened 2024-06-17
test: fix freeze on assert fail

TomMenga:
Solves: 
- Tests freeze when an assertion fails

The problem was introduced with the update of the `@web/dev-server-core` package (dependency of `@web/test-runner`).
Specifically, the [faulty change](https://github.com/modernweb-dev/web/commit/feb0c72815bd9f803ef61eab0819c9b6fd4f05d8#diff-f245ba1e1d0627b81dcc9602856b458d1b36f15d8121126f1cfce7f2c96e991f) is the use of `structuredClone`. 
The `stable` function is called before reporting the test suite result which contains a reference to the element on which the assertion failed (in our case, the web component, which cannot be cloned).

TODO: Open an issue in https://github.com/modernweb-dev/web

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,6 +149,7 @@
   "resolutions": {
     "@types/node": "20.14.2",
     "@webcomponents/template-shadowroot": "0.2.1",
+    "@web/dev-server-core": "0.7.1",
     "jackspeak": "2.1.1",
     "lit": "3.1.4"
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍 
Nice catch! 😃

PR #2789, opened 2024-06-17
test(sbb-calendar): add visual tests

DavideMininni-Fincons:


src/elements/calendar/calendar.snapshot.spec.ts
```diff
@@ -4,22 +4,30 @@ import { html } from 'lit/static-html.js';
...
-      `<sbb-calendar now="2023-01-04T00:00:00" selected="2023-01-20T00:00:00"></sbb-calendar>`,
+    it('renders - Shadow DOM', async () => {
+      await expect(element).shadowDom.to.be.equalSnapshot();
+    });
+
+    testA11yTreeSnapshot(
+      html` <sbb-calendar now="2023-01-04T00:00:00" selected="2023-01-20T00:00:00"></sbb-calendar>`,
+      undefined,
+      { safari: true }, // We skip safari because it has an inconsistent behavior on ci environment
     );
```
TomMenga, line 31:
```suggestion
    testA11yTreeSnapshot(
      undefined,
      undefined,
      { safari: true }, // We skip safari because it has an inconsistent behavior on ci environment
    );
```
You can use the same template declared in the beforeEach

TomMenga (changes requested):
The case test names are a bit unfortunate, not sure we can do something about it :/

image

DavideMininni-Fincons:
> The case test names are a bit unfortunate, not sure we can do something about it :/
> 
> image

I did a fix for it but i'm not sure it can matches all the possible cases

src/elements/core/testing/private/describe-each.ts
```diff
@@ -2,7 +2,7 @@ function generateDescribeName(payload: Record<string, unknown>): string {
   return Object.entries(payload)
     .map(
       ([key, value]) =>
-        `${key}=${typeof value === 'object' && value ? `(${generateDescribeName(value as Record<string, unknown>)})` : value}`,
+        `${key}=${typeof value === 'object' && !(value instanceof Date) && value ? `(${generateDescribeName(value as Record<string, unknown>)})` : value instanceof Date ? `${value.valueOf()}` : value}`,
```
TomMenga, line 5:
```suggestion
        `${key}=${ value instanceof Date ? `${value.valueOf()}` : typeof value === 'object' && value ? `(${generateDescribeName(value as Record<string, unknown>)})` : value}`,
```
A little improvement. 
Although, if in the future we would have to update this code again, we should probably refactor it in a separate method

TomMenga (changes requested):

src/elements/core/testing/private/describe-each.ts
DavideMininni-Fincons: I'm fixing as suggested;
using `value.toISOString()` breaks the build due to the presence of the dot in the timestamp.

TomMenga (approved):
LGTM 👍

PR #2798, opened 2024-06-20
test(sbb-container, sbb-sticky-bar): add visual spec

TomMenga:


src/elements/container/container/container.visual.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  `;
+
+  describeViewports({ viewportHeight: 600 }, () => {
+    for (const color of colorCases) {
+      it(
+        `color=${color}`,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(
+            html` <sbb-container color=${color}> ${containerContent()} </sbb-container> `,
+            { backgroundColor: 'var(--sbb-color-silver)' },
```
jeripeierSBB, line 35: padding should be set to zero for all test cases of the container (even the not expanded)

src/elements/container/sticky-bar/sticky-bar.visual.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+  });
+
+  it(
+    `viewport=medium_short content`,
+    visualDiffDefault.with(async (setup) => {
+      await setViewport({ width: SbbBreakpointMediumMin, height: 400 });
```
jeripeierSBB, line 85: sadly this could causing webkit to fail in some cases. I suggest to use the same logic as it internally does.
```suggestion
      this.ctx['requestViewport'] = { width: SbbBreakpointMediumMin, height: 400 }
```

src/elements/container/sticky-bar/sticky-bar.visual.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      await setup.withFixture(
```
jeripeierSBB, line 86: padding should be zero

jeripeierSBB (changes requested):
Nice tests. Some suggestions

src/elements/container/container/container.visual.spec.ts
TomMenga: I meant to leave the padding because it helps to have a contrast with the background color. Now I have a clear visual separation between the component colors (particularly handy with the transparent variant)

src/elements/container/sticky-bar/sticky-bar.visual.spec.ts
TomMenga: Just for info, what problem do we have with Webkit?

ps. I don't think the solution would work (it is already too late to set the ctx variable)

jeripeierSBB: Ok, then the solution would maybe be to move the viewport after the fixture creation. 

The problem is, that Safari does not apply all media queried variables correctly if the viewport is already set when the fixture is being created. But, it's reproducible: if you see, that your screenshots look fully correct, it's ok like you did.

src/elements/container/container/container.visual.spec.ts
jeripeierSBB: Hmm, the problem I see is, that from a design perspective, it's only allowed to use the container (and sticky bar) full width. So if we have the padding here, it's not showing how it's supposed to be used (even if it's only for our internal use). The part with the background color should also work if the padding is 0. Or which exact case cannot be tested?

src/elements/container/sticky-bar/sticky-bar.visual.spec.ts
TomMenga: It seems fine but to be sure I'll move the setViewport after the fixture :)

src/elements/container/container/container.visual.spec.ts
TomMenga: Makes sense.
The background thing was just a convenience factor, it works without the padding too

jeripeierSBB (approved):
LGTM, thanks :)

PR #2799, opened 2024-06-20
style: fix mini-button mixin

DavideMininni-Fincons:
The two datepicker buttons are not displayed correctly in the disabled and focused state in both Firefox and Safari; perhaps this was not detected before due to browsers being removed from Chromatic.
A similar fix was made to the popover-trigger and is also available in the new autocomplete-grid.

image
image


TomMenga (approved):
LGTM 👍

PR #2800, opened 2024-06-20
test(sbb-divider): add visual spec

TomMenga:


DavideMininni-Fincons (approved):
Very good, some minor details to look at

PR #2803, opened 2024-06-21
style(table): add 'color-iron' variant

TomMenga:


mcilurzo (approved):

src/elements/core/styles/table.scss
```diff
@@ -27,6 +27,14 @@ sbb-table-wrapper[negative] .sbb-table,
   @include table.table--unstriped;
 }
 
+.sbb-table--color-iron {
```
kyubisation, line 30: I'm still not exactly sure what the best name here would be.
We have previously used color to indicate background-color, but here it is text color.
I'm not sure whether I would prefer .sbb-table--text-color-iron or something else entirely.

kyubisation (approved):
LGTM 👍 
One open question about naming, but I would be fine with the current name.

src/elements/core/styles/table.scss
TomMenga: I'm also not certain about the name. I'd rather avoid binding the class to the property it changes (also because it does more than setting the text color to iron).
What if we intend it to be a variant color palette? It could be the "iron theme" of the table and the class name would be `.sbb-table--variant-iron` or `.sbb-table--theme-iron`
What do you think?

kyubisation: I do like `.sbb-table--theme-iron` 👍

TomMenga: Renamed 👍

PR #2807, opened 2024-06-21
test(sbb-form-field): add visual spec

DavideMininni-Fincons:
To discuss: reduce test cases?


src/elements/form-field/form-field/form-field.visual.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+        disabled: state.disabled,
+        readonly: state.readonly,
+      };
+
+      for (const [name, template] of component.entries()) {
+        it(
+          `slot=none ${name} ${visualDiffDefault.name}`,
+          visualDiffDefault.with(async (setup) => {
+            await setup.withFixture(html`${formField(args, template(args))}`, {
+              backgroundColor: negative ? '#484040' : undefined,
```
jeripeierSBB, line 198:
```suggestion
              backgroundColor: negative ? 'var(--sbb-color-black)' : undefined,
```

src/elements/form-field/form-field/form-field.visual.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+            });
+          }),
+        );
+
+        it(
+          `slot=buttons ${name} ${visualDiffDefault.name}`,
+          visualDiffDefault.with(async (setup) => {
+            const templateResult: TemplateResult = html`${template(args)} ${buttonsAndPopover(args)}`;
+            await setup.withFixture(html`${formField(args, templateResult)}`, {
+              backgroundColor: negative ? '#484040' : undefined,
```
jeripeierSBB, line 208:
```suggestion
              backgroundColor: negative ? 'var(--sbb-color-black)' : undefined,
```

src/elements/form-field/form-field/form-field.visual.spec.ts
```diff
@@ -0,0 +1,303 @@
+import { html, nothing, type TemplateResult } from 'lit';
```
jeripeierSBB, line 1: I'm not sure whether the focus is always where it's meant to be. Do you really want to test the mini button focus or was the goal to focus the input. In the second case you could use the method "withStateElement()" and pass the input element. there would be an example in the typography.visual.spec.ts file.

jeripeierSBB (changes requested):
Nice work! Basically I think it's ok if there are a lot of combinations. It's in the nature of the form field that there are a lot of possible combinations and visual outcomes.

src/elements/form-field/form-field/form-field.visual.spec.ts
DavideMininni-Fincons: Removed the focus from the states array and moved the tests with button/popover apart

jeripeierSBB (approved):
LGTM, not sure whether the active and focus state should also be tested with negative? I tend to a yes

PR #2813, opened 2024-06-24
test(sbb-dialog): add visual spec

TomMenga:


src/elements/dialog/dialog/dialog.visual.spec.ts
```diff
@@ -0,0 +1,140 @@
+import { html, nothing, type TemplateResult } from 'lit';
```
DavideMininni-Fincons, line 1: Question: in all the tests the `dialog.open();` is called after the `setup.withSnapshotElement(dialog);`
why?

DavideMininni-Fincons (approved):

src/elements/dialog/dialog/dialog.visual.spec.ts
TomMenga: `setup.withSnapshotElement(dialog);` does not actually take the snapshot. It only tells what HTML element should be snapped.
In this case, I have to specify it because the dialog is an absolute positioned element

PR #2815, opened 2024-06-24
refactor(sbb-notification): add CSS property to define icon size

kyubisation:
This PR addresses a specific requirement from sbb.ch. However, we do not want to generally provide this as a feature, so we do not provide documentation for this.
The property is named `--sbb-notification-icon-size` and can be set to `0` to hide the icon.

### DIFF
--- a/src/elements/notification/notification.scss
+++ b/src/elements/notification/notification.scss
@@ -197,6 +197,7 @@ $open-anim-height-to: calc(
   color: var(--sbb-notification-icon-color);
   margin-block-start: var(--sbb-notification-icon-margin-block-start);
 
+  // This is an undocumented feature for sbb.ch, to have the ability to hide the icon.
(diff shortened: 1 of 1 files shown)


src/elements/notification/notification.scss
```diff
@@ -196,6 +196,8 @@ $open-anim-height-to: calc(
 .sbb-notification__icon {
   color: var(--sbb-notification-icon-color);
   margin-block-start: var(--sbb-notification-icon-margin-block-start);
+
+  --sbb-icon-svg-width: var(--sbb-notification-icon-size);
```
jeripeierSBB, line 200: As we don't document this, we should add a comment here that this is here on purpose and why. Otherwise, it feels like a code smell that can be removed (nowhere used, not publicly documented).

PR #2816, opened 2024-06-24
fix(sbb-tab-group): fix nested tab groups logic

### LINKED ISSUE #2814: [Bug]: Nested tab groups don't work via dialog
### Preflight Checklist

### Bug type

Functionality

Components

1.0.0

### Operating system

macOS 14.5

### Browser / Browser version

all

### Input mode

Mouse/Keyboard

### Additional settings

_No response_

### What happened?

I have a case where a button inside a tab group opens a dialog which has a tab group in it. The "inner" tab group does not show up. The elements are properly in the DOM but they don't seem to be properly instantiated as lyne element. My guess is that the elements aren't properly instantiated because the second tab group is already a child of an instance.

This happens in an Angular project, though I am not sure if it's dependent on that.
Also, I'm fully aware that this isn't your typical use case for nested tab groups, I hope the difference is clear.

1. [Open the Stackblitz example](https://stackblitz.com/edit/angular-at-ktmnsn?file=src%2Fmain.ts,src%2Fapp%2Finner%2Finner.component.html)
2. Click the button "open dialog"
3. there won't be a tab group in the dialog

### Relevant log output

```shell
none - there are no errors showing up
```

dauriamarco:
Closes #2814


src/elements/tabs/tab/tab.ts
```diff
@@ -15,6 +16,8 @@ import style from './tab.scss?lit&inline';
 export class SbbTabElement extends LitElement {
   public static override styles: CSSResultGroup = style;
 
+  public nestedTabGroup: SbbTabGroupElement | null = null;
```
kyubisation, line 19: I would prefer not to add a property for this. I do not think it is necessary.

kyubisation (changes requested):
Thank you for your work!

jeripeierSBB (approved):
LGTM

PR #2820, opened 2024-06-25
test(sbb-header): add visual tests

DavideMininni-Fincons:


src/elements/header/header-button/header-button.visual.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+          await setup.withFixture(html`
+            <sbb-header-button icon-name="hamburger-menu-small" type="button"
+              >Menu</sbb-header-button
+            >
+          `);
+        }),
+      );
+
+      it(
+        `multiple=true ${state.name}`,
```
TomMenga, line 22: Does the multiple add any value?
I don't see any CSS rule in the header that affects multiple elements

src/elements/header/header-link/header-link.visual.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+        state.name,
+        state.with(async (setup) => {
+          await setup.withFixture(html`
+            <sbb-header-link icon-name="hamburger-menu-small">Menu</sbb-header-link>
+          `);
+        }),
+      );
+
+      it(
+        `multiple=true ${state.name}`,
```
TomMenga, line 20: Does the multiple add any value?
I don't see any CSS rule in the header that affects multiple elements

src/elements/header/header/header.visual.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+      it(
+        `expanded=${state.expanded} ${visualDiffDefault.name}`,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(
+            html`
+              <style>
+                ${' .last-element {display: none;} '}
+                ${' @media screen and (width >= 840px}) { .last-element { display: block; } }'}
+                ${' @media screen and (width < 1023px) { .sbb-header-spacer { display: none; } }'}
+              </style>
```
TomMenga, line 38: In theory, we could add this global CSS in the `web-test-runner.config.ts` but I think it's better if we keep it here

TomMenga (approved):
LGTM 👍 
Some minor topics to discuss maybe

src/elements/header/header-button/header-button.visual.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import './header-button.js';
+
+describe(`sbb-header-button`, () => {
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    for (const state of visualDiffStandardStates) {
+      it(
+        state.name,
+        state.with(async (setup) => {
+          await setup.withFixture(html`
+            <sbb-header-button icon-name="hamburger-menu-small" type="button"
```
jeripeierSBB, line 14: type button is default
```suggestion
            <sbb-header-button icon-name="hamburger-menu-small"
```

src/elements/header/header-button/header-button.visual.spec.ts
jeripeierSBB: I'd vote to remove them. multiple are tested with sbb-header?

src/elements/header/header/header.visual.spec.ts
jeripeierSBB: I'd prefer to have it only in the test :)

src/elements/header/header/header.visual.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { html } from 'lit';
```
jeripeierSBB, line 1: Could we also test the scrolling stuff?

DavideMininni-Fincons: I gave it some try yesterday reducing the container height and using scrollTo/mockScrollTo, with no result :|

jeripeierSBB: I later try if I find a solution

jeripeierSBB (approved):
LGTM

PR #2826, opened 2024-06-25
test(sbb-icon): add visual tests

DavideMininni-Fincons:


src/elements/icon/icon.visual.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { html } from 'lit';
```
jeripeierSBB, line 1: looking at icon.scss, tests which could be added are color mutations, and testing color-immutable. And if it's possible a test with "data-empty" (no icon loaded). If it's too time consuming we can save it for another improvement ;)

jeripeierSBB (approved):
LGTM, some idea for improvements

src/elements/icon/icon.visual.spec.ts
DavideMininni-Fincons: I can easily add a test with color=red and a test with no icon name; 
for the color-immutable in the call stack from 'loadSvgIcon' to 'validateContent' i see no way to do it programmatically

src/elements/icon/icon.visual.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  describeViewports({ viewports: ['small'] }, () => {
+    for (const size of ['small', 'medium', 'large']) {
+      it(
+        `icon=circle-information size=${size} ${visualDiffDefault.name}`,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(html`<sbb-icon name="circle-information-${size}"></sbb-icon>`);
+        }),
+      );
+
+      it(
```
jeripeierSBB, line 17: You could take out the color test from the size loop as color is not dependant on size

jeripeierSBB (approved):
LGTM

PR #2827, opened 2024-06-25
test(sbb-toggle, sbb-toggle-check): add visual spec

MarioCastigliano:


src/elements/toggle-check/toggle-check.visual.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+        it(
+          state.name,
+          state.with(async (setup) => {
+            await setup.withFixture(html` <sbb-toggle-check checked> Label </sbb-toggle-check> `);
+          }),
+        );
+      }
+    });
+
+    describe('without label', () => {
```
jeripeierSBB, line 60: For test cases with only one 'it' you dont need a describe. You can directly set it as title of 'it' also state (default here) is not needed

src/elements/toggle-check/toggle-check.visual.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+            <sbb-toggle-check checked
+              >Label
+              <sbb-icon slot="icon" name="eye-small"></sbb-icon>
+            </sbb-toggle-check>
+          `);
+        }),
+      );
+    });
+
+    describeEach({ checked: [false, true] }, ({ checked }) => {
```
jeripeierSBB, line 105: for single arrays you can just us a for of loop

src/elements/toggle-check/toggle-check.visual.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+        }),
+      );
+    });
+
+    describe('block variant', () => {
+      it(
+        visualDiffDefault.name,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(html`
+            <div>
```
jeripeierSBB, line 127: The outer div should not be needed as it lands anyways in a div.
```suggestion
```

src/elements/toggle/toggle-option/toggle-option.visual.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+  visualDiffDefault,
+  visualRegressionFixture,
+} from '../../core/testing/private.js';
+
+import './toggle-option.js';
+
+describe(`sbb-toggle-option`, () => {
+  let root: HTMLElement;
+
+  const cases = {
```
jeripeierSBB, line 15: For simple arrays you can directly use an array without a complex object. And then using a for of loop

src/elements/toggle/toggle-option/toggle-option.visual.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { html, nothing } from 'lit';
```
jeripeierSBB, line 1: Wondering if it is worth to have separate toggle option tests at all, or if the toggle tests are sufficient? What do you think?

src/elements/toggle/toggle-option/toggle-option.visual.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+describe(`sbb-toggle-option`, () => {
+  let root: HTMLElement;
+
+  const cases = {
+    label: [true, false],
+  };
+
+  describeViewports({ viewports: ['zero'] }, () => {
+    describe('default', () => {
```
jeripeierSBB, line 20: You can directly use 'it' without 'describe' (see other comment).

src/elements/toggle/toggle-option/toggle-option.visual.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+          <sbb-toggle-option value="Value 1">
+            ${label ? `Option` : nothing}
+            <sbb-icon name="app-icon-small" slot="icon"></sbb-icon>
+          </sbb-toggle-option>
+        `);
+      });
+
+      it(
+        `slotted icon ${visualDiffDefault.name}`,
+        visualDiffDefault.with((setup) => {
```
jeripeierSBB, line 60: I think here I'd directly use await setup.withFixture and remove the beforeEach block (as it has only this test)

src/elements/toggle/toggle/toggle.visual.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        it(
+          state.name,
+          state.with((setup) => {
+            setup.withSnapshotElement(root);
+          }),
+        );
+      }
+    });
+
+    describeEach(sizeCases, ({ size }) => {
```
jeripeierSBB, line 55: Could be simplified to a array with a for of loop and without beforeEach block.

src/elements/toggle/toggle/toggle.visual.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+      it(
+        `disabled ${visualDiffDefault.name}`,
+        visualDiffDefault.with((setup) => {
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+
+    describeEach(iconCases, ({ size, label }) => {
+      beforeEach(async function () {
```
jeripeierSBB, line 71: beforeEach could be removed and used with setup.withFixture

src/elements/toggle/toggle/toggle.visual.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      it(
+        `icon ${visualDiffDefault.name}`,
+        visualDiffDefault.with((setup) => {
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+
+    describe('slotted icon', () => {
```
jeripeierSBB, line 85: describe block not needed.

jeripeierSBB (changes requested):
Nice work! I have some little suggestions and some questions:
- Could you try to mimic this test as well: https://lyne-storybook.app.sbb.ch/?path=/story/elements-sbb-toggle-sbb-toggle--dynamic-width ?
- Could you add focus tests for the block variant and long label tests?

src/elements/toggle/toggle-option/toggle-option.visual.spec.ts
MarioCastigliano: Since there are only a few cases, I don't think there's any harm in keeping them

jeripeierSBB (approved):
Very nice :)

PR #2839, opened 2024-06-26
test(sbb-link-list): add visual spec

TomMenga:


src/elements/link-list/link-list.visual.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  let root: HTMLElement;
+
+  const cases = {
+    negative: [false, true],
+    orientation: ['horizontal', 'vertical'],
+    size: ['xs', 's', 'm'],
+  };
+
+  const linksTemplate = (): TemplateResult => html`
+    <sbb-block-link href="#">Rückerstattungen</sbb-block-link>
```
jeripeierSBB, line 23: Probably the same also in the stories, but according to our coding standards we would like english textes. If you don't mind, can you change it? Content doesn't matter ;)

jeripeierSBB (approved):
LGTM, one possible improvement

src/elements/link-list/link-list.visual.spec.ts
TomMenga: Sure 👍  The stories were fine by the way

jeripeierSBB (approved):
LGTM. thanks

PR #2847, opened 2024-06-27
test(sbb-map-container): add visual spec

TomMenga:


src/elements/map-container/map-container.visual.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+import '../title.js';
+
+describe(`sbb-map-container`, () => {
+  describeViewports({ viewports: ['zero', 'medium'], viewportHeight: 500 }, () => {
+    it(
+      visualDiffDefault.name,
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(
+          html`
+            <sbb-header expanded hide-on-scroll>
```
jeripeierSBB, line 17: Does it make a difference if hide-on-scroll is set or not?

jeripeierSBB (approved):
LGTM (not checked the outcome)

src/elements/map-container/map-container.visual.spec.ts
TomMenga: mmm no, it's a leftover from the tentative of testing the scroll behavior :/

jeripeierSBB: maybe you could remove it to not generate wrong assumptions, but not important

PR #2850, opened 2024-06-27
fix: ensure slotchange events are handled correctly in hydration

kyubisation:
This PR slightly changes how we handled slotchange events during hydration. Previously we would stop slotchange events during hydration and re-dispatch them afterwards. However, it seems that during true page load with server side rendering, slotchange events are not dispatched at all. So we changed the logic to just generally suppress slotchange events during hydration and manually dispatch a slotchange event, for any slot having content assigned, after the hydration is finished.


jeripeierSBB (approved):
LGTM! Is there no way we could test this? It feels like a lot of changes but no tests for it

PR #2852, opened 2024-06-27
test(sbb-menu): add visual tests

DavideMininni-Fincons:
NOTE FOR REVIEWERS

there's an issue in the menu.visual.spec.ts: 
- the `viewportHeight: 400` in the describeEach seems to be required to correctly display the component in zero-breakpoint;
- the explicit `await setViewport({ width: SbbBreakpointMediumMin, height: 400 });` is needed to correctly display the menu in medium-breakpoint, otherwise its position seems to be at top-left of the page.


src/elements/menu/menu-button/menu-button.stories.ts
```diff
@@ -8,7 +8,6 @@ import { sbbSpread } from '../../../storybook/helpers/spread.js';
 
 import readme from './readme.md?raw';
 import './menu-button.js';
-import '../../icon.js';
```
TomMenga, line 11: Are you sure it's not needed?

src/elements/menu/menu-button/menu-button.visual.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+            },
+          );
+        }),
+      );
+    });
+
+    for (const disabled of [true, false]) {
+      it(
+        `long label disabled=${disabled} ${visualDiffDefault.name}`,
+        visualDiffDefault.with(async (setup) => {
```
TomMenga, line 74: Maybe I'd also test hover and focus here

src/elements/menu/menu-link/menu-link.visual.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+            },
+          );
+        }),
+      );
+    });
+
+    for (const disabled of [true, false]) {
+      it(
+        `long label disabled=${disabled} ${visualDiffDefault.name}`,
+        visualDiffDefault.with(async (setup) => {
```
TomMenga, line 74: Maybe I'd also test hover and focus here

src/elements/menu/menu/menu.visual.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+              </sbb-menu-link>
+              <sbb-menu-button icon-name="pen-small" amount="16" disabled> Edit </sbb-menu-button>
+              <sbb-menu-button icon-name="swisspass-small" amount="123"> Details</sbb-menu-button>
+              <sbb-divider></sbb-divider>
+              <sbb-menu-button icon-name="cross-small">Cancel</sbb-menu-button>
+            </sbb-menu>
+          `,
+          { minHeight: '400px' },
+        );
+        await setViewport({ width: SbbBreakpointMediumMin, height: 400 });
```
TomMenga, line 47:
```suggestion
        await setViewport(this.test?.ctx?.['requestViewport']);
```
Little suggestion 

 Ps.
From my understanding, you need to set the viewport before opening the menu to position it correctly which seems reasonable.
I think @jeripeierSBB mentioned a stability problem on webkit if we set the viewport before the fixture. Not sure if it applies here

src/elements/menu/menu/menu.visual.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+              <sbb-menu-button>Nederlands</sbb-menu-button>
+              <sbb-menu-button>Suomi</sbb-menu-button>
+              <sbb-menu-button>українська мова</sbb-menu-button>
+              <sbb-menu-button>አማርኛ</sbb-menu-button>
+              <sbb-menu-button>ქართული ენა</sbb-menu-button>
+              <sbb-menu-button>Afrikaans</sbb-menu-button>
+              <sbb-menu-button>Svenska</sbb-menu-button>
+              <sbb-menu-button>Dansk</sbb-menu-button>
+              <sbb-menu-button>Nederlands</sbb-menu-button>
+              <sbb-menu-button>Suomi</sbb-menu-button>
```
TomMenga, line 83: I would use the repeat or something like
```
new Array(10).map(_ => html`...`)
```

TomMenga (changes requested):
GG
Some minor fixes and topics

src/elements/menu/menu-button/menu-button.visual.spec.ts
DavideMininni-Fincons: disabled case has no focus/icon, but i could split in two (disabled default + long label default/hover/focus)
what do you think?

src/elements/menu/menu/menu.visual.spec.ts
DavideMininni-Fincons: `this` is undefined, but i totally agree with you about moving the logic before the `withFixture`
the same issue is present in the notification and i possibly in every overlay component (hope not)

src/elements/menu/menu-button/menu-button.visual.spec.ts
TomMenga: I'd like to test that focus and hover do not have any visual effect (since it's something that we handle manually in our implementation). A sort of non-regression test

src/elements/menu/menu-button/menu-button.visual.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    disabled,
+    slottedIcon,
+  }: typeof defaultArgs): TemplateResult => html`
+    ${repeat(
+      new Array(3),
+      (_, index) => html`
+        <sbb-menu-button
+          amount=${amount || nothing}
+          icon-name=${iconName || nothing}
+          .disabled=${disabled}
```
jeripeierSBB, line 37: doesn't matter but would also work :P
```suggestion
          ?disabled=${disabled}
```

src/elements/menu/menu-link/menu-link.visual.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    slottedIcon,
+  }: typeof defaultArgs): TemplateResult =>
+    html` ${repeat(
+      new Array(3),
+      (_, index) => html`
+        <sbb-menu-link
+          href="#"
+          amount=${amount || nothing}
+          icon-name=${iconName || nothing}
+          .disabled=${disabled}
```
jeripeierSBB, line 38: optional
```suggestion
          ?disabled=${disabled}
```

jeripeierSBB (commented):
Should we also discuss the viewport stuff?

jeripeierSBB (approved):
LGTM

PR #2853, opened 2024-06-28
build: inline koa-connect

kyubisation:
This PR inlines the `koa-connect` functionality, as the logic is fairly simple.


TomMenga (approved):
LGTM 👍

PR #2869, opened 2024-07-01
test(sbb-title): implement visual regression tests

MarioCastigliano:
not sure if it's worth testing all viewport sizes

### DIFF
--- a/src/elements/title/title.visual.spec.ts
+++ b/src/elements/title/title.visual.spec.ts
@@ -1,0 +1,35 @@
+import { html } from 'lit';
+
+import { describeViewports, visualDiffDefault } from '../core/testing/private.js';
+
+import '../title.js';
+
+describe(`sbb-title`, () => {
+  describeViewports({ viewports: ['zero', 'small', 'large', 'ultra'] }, () => {
+    for (const level of ['1', '2', '3', '4', '5', '6']) {
+      it(
+        `visual level ${level}`,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(
+            html`<sbb-title level=${level}>
+              Data without insights are trivial, and insights without action are pointless
+            </sbb-title>`,
+          );
+        }),
+      );
+    }
+
+    it(
+      'negative',
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(
+          html` <div style="padding: 1rem; background-color: var(--sbb-color-charcoal);">
+            <sbb-title negative>
+              Data without insights are trivial, and insights without action are pointless
+            </sbb-title>
+          </div>`,
+        );
+      }),
+    );
+  });
+});
--- a/src/elements/title/title.visual.spec.ts
+++ b/src/elements/title/title.visual.spec.ts
@@ -5,15 +5,21 @@ import { describeViewports, visualDiffDefault } from '../core/testing/private.js
 import '../title.js';
 
 describe(`sbb-title`, () => {
-  describeViewports({ viewports: ['zero', 'small', 'large', 'ultra'] }, () => {
+  describeViewports({ viewports: ['zero', 'small', 'medium', 'large', 'ultra'] }, () => {
     for (const level of ['1', '2', '3', '4', '5', '6']) {
       it(
-        `visual level ${level}`,
+        `level=${level}`,
         visualDiffDefault.with(async (setup) => {
           await setup.withFixture(
             html`<sbb-title level=${level}>
-              Data without insights are trivial, and insights without action are pointless
-            </sbb-title>`,
+                Data without insights are trivial, and insights without action are pointless
+              </sbb-title>
+              <article>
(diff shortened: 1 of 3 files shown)


src/elements/title/title.visual.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    it(
+      'negative',
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(
+          html` <div style="padding: 1rem; background-color: var(--sbb-color-charcoal);">
+            <sbb-title negative>
+              Data without insights are trivial, and insights without action are pointless
+            </sbb-title>
+          </div>`,
+        );
```
TomMenga, line 31:
```suggestion
        await setup.withFixture(
          html`
            <sbb-title negative>
              Data without insights are trivial, and insights without action are pointless
            </sbb-title>
          `, { backgroundColor: negative ? 'var(--sbb-color-charcoal)' : undefined}
        );
```

src/elements/title/title.visual.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { describeViewports, visualDiffDefault } from '../core/testing/private.js';
+
+import '../title.js';
+
+describe(`sbb-title`, () => {
+  describeViewports({ viewports: ['zero', 'small', 'large', 'ultra'] }, () => {
+    for (const level of ['1', '2', '3', '4', '5', '6']) {
+      it(
+        `visual level ${level}`,
```
TomMenga, line 11:
```suggestion
        `level=${level}`,
```

src/elements/title/title.visual.spec.ts
```diff
@@ -0,0 +1,35 @@
+import { html } from 'lit';
+
+import { describeViewports, visualDiffDefault } from '../core/testing/private.js';
+
+import '../title.js';
+
+describe(`sbb-title`, () => {
+  describeViewports({ viewports: ['zero', 'small', 'large', 'ultra'] }, () => {
```
TomMenga, line 8: I'm conflicted about whether we should test all the viewports :/
@jeripeierSBB what do you think?

TomMenga (changes requested):
Good job :)
Some minor fixes

src/elements/title/title.visual.spec.ts
jeripeierSBB: I think the best would be ['zero', 'small', 'medium', 'large', 'ultra'] as for all those values, some title sizes are toggled specifically

jeripeierSBB: or maybe just use every viewport size...

src/elements/title/title.visual.spec.ts
jeripeierSBB: I think it could be combined into a describeEach with title levels and negative prop in one loop.

src/elements/title/title.visual.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    for (const level of ['1', '2', '3', '4', '5', '6']) {
+      it(
+        `visual level ${level}`,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(
+            html`<sbb-title level=${level}>
+              Data without insights are trivial, and insights without action are pointless
```
jeripeierSBB, line 15: I'd suggest (even when it was not like that in Chromatic) to add a <p> tag below the title with a short random text to cover also the margins which are part of the sbb-title.

MarioCastigliano: Good idea

src/elements/title/title.visual.spec.ts
```diff
@@ -0,0 +1,42 @@
+import { html } from 'lit';
+
+import { describeViewports, visualDiffDefault } from '../core/testing/private.js';
+
+import '../title.js';
+
+describe(`sbb-title`, () => {
+  describeViewports({ viewports: ['zero', 'small', 'medium', 'large', 'ultra'] }, () => {
+    for (const level of ['1', '2', '3', '4', '5', '6']) {
```
jeripeierSBB, line 9: what about using
```
const cases = {negative: [false, true], level: [1,2,3,4,5,6]}
```
and the describeEach?
With that you can share the template and backgroundColor can be toggled by negative prop.

jeripeierSBB (approved):
LGTM, one possible refactor

src/elements/title/title.visual.spec.ts
MarioCastigliano: I just assumed we didn't need to test the negative variant with all the levels and wanted to keep a relatively small number of screenshots. But I could live either way, what do you think it's best?

jeripeierSBB: excuse me, I misread the closing tag ;) It's fine like you did! :)

jeripeierSBB (changes requested):
please adapt also the snapshot.spec.ts file

jeripeierSBB (approved):
LGTM, thank you

PR #2878, opened 2024-07-02
test(sbb-select): add visual tests

DavideMininni-Fincons:


src/elements/select/select.visual.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  describeViewports({ viewports: ['zero', 'medium'], viewportHeight: 400 }, () => {
+    for (const negative of [false, true]) {
+      for (const visualDiffState of [visualDiffDefault, visualDiffFocus]) {
+        it(
+          `state=above negative=${negative} ${visualDiffState.name}`,
+          visualDiffState.with(async (setup) => {
+            await setup.withFixture(
+              html`
+                <div style="position: absolute; inset-block-end: 2rem; width: calc(100% - 4rem);">
```
jeripeierSBB, line 107: Did you try without width information? I think it causes horizontal problems: https://lyne-components-visual-regression-diff-pr2878.app.sbb.ch/compare/sbb-select/state=above_negative=false_default.

src/elements/select/select.visual.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+import '../form-field.js';
+import '../option.js';
+import './select.js';
+
+describe('sbb-select', () => {
+  const valueEllipsis: string = 'This label name is so long that it needs ellipsis to fit.';
+  const defaultArgs = {
+    borderless: false,
+    negative: false,
+    floatingLabel: false,
```
jeripeierSBB, line 15: I see no tests with floatingLabel true, should we add or remove floatingLabels from arguments?

src/elements/select/select.visual.spec.ts
```diff
@@ -0,0 +1,251 @@
...
+    withOptionGroup,
+    disableGroup,
+    withEllipsis,
+    ...args
+  }: typeof defaultArgs): TemplateResult => {
+    if (args.multiple && args.value) {
+      args.value = [args.value as string];
+    }
+    return html`
+      <div>
```
jeripeierSBB, line 69: Can this div be removed?

src/elements/select/select.visual.spec.ts
DavideMininni-Fincons: There are no stories in storybook too; i propose to remove it, since the floating case is handled in the form-field.
What do you think?

jeripeierSBB: ok!

src/elements/select/select.visual.spec.ts
DavideMininni-Fincons: I removed as a test, in local the form-field was a bit shrinked, let's see :)

jeripeierSBB: should be ok I think.

jeripeierSBB (approved):
LGTM

PR #2882, opened 2024-07-03
test(sbb-popover, sbb-popover-trigger): add visual spec

TomMenga:


DavideMininni-Fincons (commented):
Almost perfect

TomMenga:
> Almost perfect
> 
> edit: looking at storybook, it could be possibile to test also the hover case?

Do you mean to test the open when hovering? 
I feel like that's not a responsibility of visual specs though

DavideMininni-Fincons (approved):

PR #2883, opened 2024-07-03
test(sbb-toast): add visual spec

TomMenga:


src/elements/toast/toast.visual.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  const positionCases = [
+    'top-start',
+    'top-center',
+    'top-end',
+    'bottom-start',
+    'bottom-center',
+    'bottom-end',
+  ];
+
+  describeViewports({ viewports: ['zero', 'medium'], viewportHeight: 300 }, () => {
```
jeripeierSBB, line 24: I see some zero screenshots not present. is there any error present?

jeripeierSBB: e.g. `https://lyne-components-visual-regression-diff-pr2883.app.sbb.ch/compare/sbb-toast/icon=false-action=button_`

src/elements/toast/toast.visual.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    describeEach(cases, ({ icon, action }) => {
+      it(
+        '',
```
jeripeierSBB, line 27: This leads to an underscore at the end in the test name. should we clean this somehow? (see link above)

src/elements/toast/toast.visual.spec.ts
TomMenga: I believe the empty name is the cause of it. I'll have a look and see if we can easily fix it

 I just wanted to avoid a useless 'default' :/

TomMenga: I forgot an await :| 
I was unfair to blame the test name

src/elements/toast/toast.visual.spec.ts
TomMenga: I added a `trim()` before the title transformation. It solves the problem and should not lead to any regression (except for the tests that ended with a white space)

PR #2888, opened 2024-07-04
test(skiplink-list): add visual spec

TomMenga:


src/elements/skiplink-list/skiplink-list.visual.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+`;
+
+describe(`sbb-skiplink-list`, () => {
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    it(
+      'not focused',
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(
+          html` <sbb-skiplink-list title-content="Skip to"> ${links()} </sbb-skiplink-list> `,
+          { minHeight: '100px' },
```
DavideMininni-Fincons, line 21: you can refactor this by creating a const and passing it in tests

src/elements/skiplink-list/skiplink-list.visual.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+const links = (): TemplateResult => html`
+  <sbb-block-link href="#">Link 1</sbb-block-link>
+  <sbb-block-link href="#">Link 2</sbb-block-link>
+  <sbb-block-link href="#">Link 3</sbb-block-link>
+`;
+
+describe(`sbb-skiplink-list`, () => {
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    it(
+      'not focused',
```
DavideMininni-Fincons, line 17: this looks like the 'no title' but unfocused; could we have a for cycle with default/focus?

DavideMininni-Fincons (commented):
Looks good, some comments added

src/elements/skiplink-list/skiplink-list.visual.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            <sbb-skiplink-list>
+              <span slot="title">Slotted title</span>
+              ${links()}
+            </sbb-skiplink-list>
+          `,
+          { minHeight: '100px' },
+        );
+      }),
+    );
+  });
```
jeripeierSBB, line 63: Should we really test visually hidden stuff? btw. there is a visual-level attribute on the title which could be removed

TomMenga: Maybe I'm being too zealous but I believe it's the only way we have to be sure that they stay hidden :/

jeripeierSBB: ok that's true, but maybe remove just the slottted one?

DavideMininni-Fincons (approved):
LGTM

jeripeierSBB (approved):
LGTM

PR #2890, opened 2024-07-04
fix(sbb-button): fix high contrast styles

jeripeierSBB:


src/elements/button/button-static/button-static.visual.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+import { describeViewports, visualDiffStandardStates } from '../../core/testing/private.js';
+
+import './button-static.js';
+
+// We test only the differences to the sbb-button and sbb-button-link
+describe(`sbb-button-static`, () => {
+  describeViewports({ viewports: ['zero'] }, () => {
+    for (const forcedColors of [false, true]) {
+      describe(`forcedColors=${forcedColors}`, () => {
+        for (const state of visualDiffStandardStates) {
```
DavideMininni-Fincons, line 12: static button don't need focus since it cant be focused

DavideMininni-Fincons (approved):
LGTM, just a case can be removed

src/elements/button/button-static/button-static.visual.spec.ts
jeripeierSBB: my plan was to ensure it's not focused :D

DavideMininni-Fincons: oh, so it's fine as it is :D and maybe i should recheck some old commits of mine, because i feel i avoided that specific case

jeripeierSBB: The more I think I tend to remove it. :D

jeripeierSBB: I removed it :D__

PR #2895, opened 2024-07-05
test(sbb-selection-expansion-panel): add visual spec

TomMenga:


MarioCastigliano (approved):
Looking good, maybe I would try to optimize the number of screenshots since for all the cases where the panel is closed the visual tests on the radio-button/checkbox would practically overlap

DavideMininni-Fincons (approved):
Looks good!

jeripeierSBB (commented):
Basically looks good. One thought:
Comparing https://lyne-components-visual-regression-diff-pr2895.app.sbb.ch/compare/sbb-selection-expansion-panel/checkbox_group_with_error to https://lyne-components-visual-regression-diff-pr2895.app.sbb.ch/compare/sbb-selection-expansion-panel/radio_button_group_with_error has a shift between the price. is this only an example problem or do we have style problems in the component itself?

TomMenga:
> https://lyne-components-visual-regression-diff-pr2895.app.sbb.ch/compare/sbb-selection-expansion-panel/checkbox_group_with_error

I was already investigating it and I'd say it's a component problem. This is due to the different "dom structure" between the radio and the checkbox. 
Since the fix might impact several components, I'd like a baseline to compare to. So I'm going to merge this and open a new one for the fix

PR #2896, opened 2024-07-05
test(sbb-slider): add visual tests

DavideMininni-Fincons:


src/elements/slider/slider.visual.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  const variants = [
+    { name: 'standalone', template: standaloneTemplate },
+    { name: 'form-field', template: formFieldTemplate },
+  ];
+
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    for (const variant of variants) {
+      for (const visualDiffState of [visualDiffDefault, visualDiffFocus]) {
+        it(
```
MarioCastigliano, line 52: Since it's 5 tests instead of inlining variant.name each time you could wrap them in a describe block, what do you think?

src/elements/slider/slider.visual.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const variants = [
+    { name: 'standalone', template: standaloneTemplate },
+    { name: 'form-field', template: formFieldTemplate },
+  ];
+
+  describeViewports({ viewports: ['zero', 'medium'] }, () => {
+    for (const variant of variants) {
+      describe(`${variant.name}`, () => {
+        for (const visualDiffState of [visualDiffDefault, visualDiffFocus]) {
```
jeripeierSBB, line 52: active could be tested too I think?

jeripeierSBB (commented):
Basically looks good, what about the active state?

src/elements/slider/slider.visual.spec.ts
DavideMininni-Fincons: good catch :) state added

src/elements/slider/slider.visual.spec.ts
DavideMininni-Fincons: it should be fixed now

jeripeierSBB (approved):
LGTM

PR #2897, opened 2024-07-05
fix(various): ensure slotchange event is handled properly in hydration

kyubisation:
In components that have a `slotchange` event handler for an initialization and no fallback, we need to use the `SbbHydrationMixin` to ensure the `slotchange` event is fired even with declarative Shadow DOM.


jeripeierSBB (approved):
LGTM, good catch. Would it be possible to have a linting rule for that as tests seem to not cover it?

kyubisation:
Not sure, as I don't think we can differentiate between initializing and updating call. Or do you have an idea? (Also asked via Teams.)

kyubisation:
Discussed via Teams; Linting rule not possible.

PR #2900, opened 2024-07-08
fix(sbb-radio-button-panel): fix spacing between label and suffix

TomMenga:


src/elements/radio-button/common/radio-button-common.scss
```diff
@@ -79,6 +79,7 @@
 .sbb-radio-button__label-slot {
   display: flex;
   align-items: flex-start;
+  gap: var(--sbb-spacing-fixed-2x);
```
jeripeierSBB, line 82: I think we should avoid using `gap` here, as this would influence the slotted content (display: contents on <slot>). But the same problem would also exist in the checkbox.

TomMenga: I don't see a problem with it though 🤔 
If a `<slot>` element is empty it is not considered at all by the flex container, so it does not reserve any gap space
Not sure if that's what you meant

jeripeierSBB: If you e.g. slot <span>first text part</span></span>second part</span> it would create a gap between, but this should be user controlled.

TomMenga: True,  did not think about that. I guess I would live with it since I believe it to be a rare edge case.
Also because I don't like the alternative of surrounding the slots with a container and using the `SlotStateController` to show/hide them

jeripeierSBB (approved):
LGTM :)

PR #2903, opened 2024-07-08
test(sbb-time-input): add visual spec

TomMenga:


MarioCastigliano (approved):
👍

src/elements/time-input/time-input.visual.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { html, nothing, type TemplateResult } from 'lit';
```
DavideMininni-Fincons, line 1: i would add a case without prefix/suffix icons

DavideMininni-Fincons (commented):
LGTM

src/elements/time-input/time-input.visual.spec.ts
TomMenga: Makes sense :)

PR #2908, opened 2024-07-08
test(sbb-radio-button, sbb-radio-button-group, sbb-radio-button-panel): implement visual spec

MarioCastigliano:


src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+const suffixStyle: Readonly<StyleInfo> = {
+  display: 'flex',
+  alignItems: 'center',
+  marginInline: 'var(--sbb-spacing-fixed-2x)',
+};
+
+const suffixAndSubtext = (): TemplateResult => html`
+  <span slot="subtext">Subtext</span>
+  <span slot="suffix" style="margin-inline-start: auto;">
+    <span style=${styleMap(suffixStyle)}>
```
jeripeierSBB, line 31: this span potentially could be combined with the parent span. Is all the additional styling in the suffix slot really needed?

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  { name: 'radio-button', template: radioButtons() },
+  { name: 'radio-button-panel', template: radioButtonPanels() },
+];
+
+describe(`sbb-radio-button-group`, () => {
+  describeViewports({ viewports: ['small', 'medium'] }, () => {
+    describeEach(cases, ({ orientation, size, disabled }) => {
+      for (const variant of variants) {
+        for (const state of [visualDiffDefault, visualDiffFocus]) {
+          (disabled && state === visualDiffFocus ? it.skip : it)(
```
jeripeierSBB, line 61: I'd not use "skip" as this appears in the statistics. maybe only create an if around it? or leave it and then we test that disabled is not focusable :-)

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+            `${variant.name} ${state.name}`,
```
jeripeierSBB, line 62: not a problem, but could also have a `describe(variant.name)` directly below `for (const variant of variants) {`

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(
+            html`<sbb-radio-button-group allow-empty-selection>
+              ${variant.template}
+            </sbb-radio-button-group>`,
+          );
+        }),
+      );
+    }
+
```
jeripeierSBB, line 91: should we also add a test for horizontal-from property?

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    for (const variant of ['vertical', 'horizontal']) {
+      it(
+        `error message ${variant}`,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(
+            html`<sbb-radio-button-group allow-empty-selection orientation=${variant}>
+              ${variants[0].template}
```
jeripeierSBB, line 98: should the panels also have such tests?

src/elements/radio-button/radio-button-panel/radio-button-panel.visual.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    describeEach(cases, ({ checked, disabled, size }) => {
+      it(
+        visualDiffDefault.name,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(html`
+            <sbb-radio-button-panel ?checked=${checked} ?disabled=${disabled} size=${size}>
+              Value
+              <span slot="subtext">Subtext</span>
+              <span slot="suffix" style="margin-inline-start: auto;">
+                <span style="display:flex;align-items:center;">
```
jeripeierSBB, line 25: same question, could the spans be combined? (for all tests in this file)

src/elements/radio-button/radio-button-panel/radio-button-panel.visual.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+          `);
+        }),
+      );
+    }
+
+    it(
+      'with bold label',
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(
+          html` <sbb-radio-button-panel>
```
jeripeierSBB, line 93: Maybe it could be helpful to share the template as it's a lot of template for every test? just an idea, you don't have to.

src/elements/radio-button/radio-button/radio-button.visual.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import { describeEach, describeViewports, visualDiffDefault } from '../../core/testing/private.js';
+
+import '../radio-button.js';
+
+const cases = {
+  size: ['m', 's'],
+  checked: [true, false],
+  disabled: [false, true],
+  bold: [false, true],
```
jeripeierSBB, line 11: I wonder whether we should test the bold case (for all the test files in this PR). this is entirely on the consumers' side (was just an example in chromatic).

jeripeierSBB (commented):
Nice work :)

src/elements/radio-button/radio-button/radio-button.visual.spec.ts
MarioCastigliano: Makes sense

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
MarioCastigliano: I was thinking the same but I wouldn't know how. What do you suggest?

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
MarioCastigliano: The component allows it technically, so I think it's worth a screenshot. Don't you think?

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
jeripeierSBB: I think it would be best because this probably nowhere really tested (the visual outcome of the css rules)?

src/elements/radio-button/radio-button-group/radio-button-group.visual.spec.ts
jeripeierSBB: yes, let's add it :)

jeripeierSBB (approved):
LGTM, nice work

PR #2910, opened 2024-07-09
chore: add visual.spec.ts to boilerplate

DavideMininni-Fincons:
For reviewers:
no idea of what else should be added; a describeEach? specify viewports maybe?

### DIFF
--- a/tools/generate-component/boilerplate/component.visual.spec.ts
+++ b/tools/generate-component/boilerplate/component.visual.spec.ts
@@ -1,6 +1,11 @@
 import { html } from 'lit';
 
-import { describeViewports, visualDiffStandardStates } from '../core/testing/private.js';
+import {
+  describeViewports,
+  describeEach,
+  visualDiffDefault,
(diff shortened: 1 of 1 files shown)


tools/generate-component/boilerplate/component.visual.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+import { describeViewports, visualDiffStandardStates } from '../core/testing/private.js';
+
+import './__noPrefixName__.js';
+
+describe('__name__', () => {
+  /**
+   * Add the `viewports` param to test only specific viewport;
+   * add the `viewportHeight` param to set a fixed height for the browser.
+   */
+  describeViewports(() => {
```
jeripeierSBB, line 12: I think a cases with describEach example would be great. But at the end it doesn't matter. most important is, that this file is here :D

jeripeierSBB (approved):
LGTM thanks

PR #2911, opened 2024-07-09
test(sbb-pearl-chain): add visual spec

TomMenga:


src/elements-experimental/pearl-chain/pearl-chain.visual.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    {
+      name: 'mixed',
+      legs: [pastLeg, progressLeg, longFutureLeg, cancelledLeg, futureLeg],
+      now: new Date('2022-12-05T12:11:00').valueOf(),
+    },
+  ];
+
+  describeViewports({ viewports: ['medium'] }, () => {
+    for (const c of cases) {
+      it(
```
jeripeierSBB, line 55: would it also work to use describeEach() ?

src/elements-experimental/pearl-chain/pearl-chain.visual.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+        c.name,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(html`
+            <sbb-pearl-chain
+              .legs=${c.legs}
+              now=${(c.now ?? new Date('2022-12-01T12:11:00').valueOf()) / 1000}
```
jeripeierSBB, line 61: I think for all tests the `now` should be mandatory. Even if it does not change the outcome, just to always talk about a specific state

TomMenga: Do you mean to remove the default and explicitly set it for each case?

src/elements-experimental/pearl-chain/pearl-chain.visual.spec.ts
TomMenga: Don't think so 🤔 
Or better, we could but it's a mono-dimension array of preconfigured cases (copied from the stories)
I don't see the value in using the describeEach since there is no actual combination of params

src/elements-experimental/pearl-chain/pearl-chain.visual.spec.ts
jeripeierSBB: No, I'm sorry, I probably was not concentrated during the review :D. I missed the fallback :D (don't ask why, I don't know)

src/elements-experimental/pearl-chain/pearl-chain.visual.spec.ts
jeripeierSBB: I agree :)

jeripeierSBB (approved):
LGTM, thanks!

PR #2921, opened 2024-07-10
fix(sbb-container, sbb-tab-group): avoid reserving invisible space

### LINKED ISSUE #2835: [Bug]: tab-content pseudo-elements create scrollbar inside map-container
### Preflight Checklist

### Bug type

Functionality

Components

1.0.0

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

Found a bug on a mobile-sized screen when combining map-container with tab-group. An unnecessary scrollbar ist shown.
Reproduction: https://stackblitz.com/edit/tab-content-after-creates-scrollbar-inside-map-container?file=src%2Fmain.html

After some testing it seems to be a combination of 2 issues related to this mixin: [src/elements/tabs/tab-group/tab-group.scss#L21](https://github.com/sbb-design-systems/lyne-components/blob/main/src/elements/tabs/tab-group/tab-group.scss#L21)

One is that the `::after` pseudo-element still occupies space underneath the tab-content.
This could be solved by adding `overflow: hidden` to the mixin

The second is that the height which is set here: [src/elements/tabs/tab-group/tab-group.ts#L247](https://github.com/sbb-design-systems/lyne-components/blob/main/src/elements/tabs/tab-group/tab-group.ts#L247) is too small because the margin-top of the paragraph inside isnt included.
This can also be seen in the [storybook](https://lyne-storybook.app.sbb.ch/?path=/docs/elements-sbb-tab-sbb-tab-group--docs#default%20tabs%20size%20l) by changing to "Tab title four" the scrollbar appears.
This could be solved by moving the mixin from the tab-group to the tab itself.

-

### Relevant log output

_No response_

jeripeierSBB:
Closes #2835 

Open topic: wrong height calculation

### DIFF
--- a/src/elements/container/container/container.visual.spec.ts
+++ b/src/elements/container/container/container.visual.spec.ts
@@ -25,14 +25,16 @@ describe(`sbb-container`, () => {
     <sbb-secondary-button style="margin-block-end: 3rem;">See more</sbb-secondary-button>
   `;
 
+  const wrapperStyles = { backgroundColor: 'var(--sbb-color-silver)', padding: '0' };
+
   describeViewports({ viewportHeight: 600 }, () => {
     for (const color of colorCases) {
       it(
         `color=${color}`,
         visualDiffDefault.with(async (setup) => {
           await setup.withFixture(
-            html` <sbb-container color=${color}> ${containerContent()} </sbb-container> `,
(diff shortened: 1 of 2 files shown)


kyubisation (approved):
LGTM 👍

PR #2937, opened 2024-07-15
style: apply scss extension rule

kyubisation:
The rule `scss/at-import-partial-extension` was not actually applied and is being replaced by the new `scss/load-partial-extension`.


TomMenga (approved):
LGTM 👍

PR #2966, opened 2024-07-25
fix(sbb-clock): fix internal await to actually finish

kyubisation:
Previously `_startClock` would not finish, as the awaited Promise would not resolve. This prevented other internal methods from completing.


TomMenga (approved):
LGTM 👍

PR #2971, opened 2024-07-26
feat(sbb-icon): provide compatibility with sbb-angular

### LINKED ISSUE #2746: Backwards compatibility for sbb-icon to Angular sbb-icon
If using Lyne and sbb-angular in a mixed Angular project, the sbb-icon of Lyne and sbb-angular do not work together as they share the name.
We want to detect with attributeChangedCallback() wether `svgicon` attribute was set (knowing that we are in angular env.). If yes, we replace the render() with only <slot></slot> as return value, where the svg of the angular library will land in.

kyubisation:
Closes #2746


kyubisation:
Proof of concept: https://stackblitz.com/edit/q8ssqa?file=src%2Fmain.ts

PR #2986, opened 2024-08-05
fix: fix inert mechanism for overlays

### LINKED ISSUE #2969: [Bug]: data-sbb-inert blocks interaction for dialogs
### Preflight Checklist

### Bug type

Functionality

Components

1.2.0

### Operating system

Windows 10

### Browser / Browser version

Chrome

### Input mode

None

### Additional settings

_No response_

### What happened?

the "data-sbb-inert" or "inert" attribute is blocking the interaction for new opened dialogs.

1.  use a sbb-dialog is hierarchically at the top of the DOM with an action button.
2. The action button opens a new sbb-dialog, which is positioned hierarchically at the bottom of the DOM.
3. The newly opened dialog is displayed, but its interaction is blocked by the data-sbb-inert attribute of the first dialog.

### Relevant log output

_No response_

jeripeierSBB:
Closes #2969


src/elements/core/overlay/inert.ts
```diff
@@ -0,0 +1,108 @@
+const IGNORED_ELEMENTS = ['script', 'head', 'template'];
```
WalkingOS, line 1: is there maybe more elements which should be ignored? Maybe <link/> <style/> <meta/> ? Are we assuming here that those elements are usually inside the <head/> element?

jeripeierSBB: Thanks for the hint, we can add style as well. These elements in the list have no direct impact on the inert mechanism as they are more meta-like. Conversely, there wouldn't be any change if they received inert, so at the end it doesn't change much.

dauriamarco (approved):
Thank you very much! LGTM :)

src/elements/core/controllers/inert-controller.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  public hostConnected(): void {
+    if (this._host.isOpen) {
+      this.activate();
+    }
+  }
+
+  public hostDisconnected(): void {
+    this.deactivate(true);
```
kyubisation, line 25: As discussed

kyubisation (approved):
LGTM 👍

PR #3018, opened 2024-08-22
feat(sbb-time-input): add size 's'

### LINKED ISSUE #2769: Time-Input: Size S
https://www.figma.com/file/9r6xSfNmEfCFxl1yFYedrj/Lyne-Components?type=design&node-id=29682%3A408302&mode=design&t=aBVGHDIkLF2bjfjV-1

TomMenga:
This PR Closes #2769 

Most of the task was accomplished by the form field. 

The only doubt I have is on a CSS rule in `core.scss` that sets the max-width for time inputs. 
The value seems arbitrary and I could not find it in Figma. Might be that we need to tune it when sizing 's'
@jeripeierSBB Perhaps you know something more?

image

### DIFF
--- a/src/elements/time-input/time-input.visual.spec.ts
+++ b/src/elements/time-input/time-input.visual.spec.ts
@@ -21,11 +21,17 @@ describe(`sbb-time-input`, () => {
     withError: [false, true],
   };
 
+  const sizeCases = {
+    size: ['s', 'm', 'l'],
+    noIcons: [false, true],
+  };
+
   type ParamsType = { [K in keyof typeof cases]: (typeof cases)[K][number] } & {
+    [K in keyof typeof sizeCases]: (typeof sizeCases)[K][number];
+  } & {
     readonly?: boolean;
     borderless?: boolean;
     disabled?: boolean;
-    noIcons?: boolean;
   };
   const template = (args: Partial<ParamsType>): TemplateResult => html`
     <sbb-form-field ?borderless=${args.borderless} ?negative=${args.negative} width="collapse">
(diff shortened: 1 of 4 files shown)


jeripeierSBB:
@TomMenga for the size s, it would be 51px and from medium 58px. So we have a little problem here :D

TomMenga:
> @TomMenga for the size s, it would be 51px and from medium 58px. So we have a little problem here 

Not the best looking but something like this works. Ps. can I ask you where did you get the value?
image

jeripeierSBB:
@TomMenga  From Figma by clicking on the input:
Bildschirmfoto 2024-08-22 um 16 09 15
Bildschirmfoto 2024-08-22 um 16 09 10


Maybe it's more clean, if we create a separate var for it and then only do the assignment in the block you showed? Would align a little bit more into the current solution

TomMenga:
> @TomMenga From Figma by clicking on the input: Bildschirmfoto 2024-08-22 um 16 09 15 Bildschirmfoto 2024-08-22 um 16 09 10
> 
> Maybe it's more clean, if we create a separate var for it and then only do the assignment in the block you showed? Would align a little bit more into the current solution

Don't know how I missed it... I blame the holidays 😐

DavideMininni-Fincons (approved):
LGTM, the readme should be updated with a small note, as done in the sbb-select

src/elements/time-input/time-input.visual.spec.ts
```diff
@@ -21,11 +21,17 @@ describe(`sbb-time-input`, () => {
...
+
   type ParamsType = { [K in keyof typeof cases]: (typeof cases)[K][number] } & {
+    [K in keyof typeof sizeCases]: (typeof sizeCases)[K][number];
+  } & {
     readonly?: boolean;
     borderless?: boolean;
     disabled?: boolean;
-    noIcons?: boolean;
   };
   const template = (args: Partial<ParamsType>): TemplateResult => html`
```
jeripeierSBB, line 36: size needs to be respected. right now, all visual regression tests have the same size

jeripeierSBB (changes requested):
Generally looks good, but screenshots are all using the same size.

jeripeierSBB (approved):
LGTM

PR #3027, opened 2024-08-26
fix(sbb-select): fix display value in SSR context

### LINKED ISSUE #3021: bug(sbb-select): SSR test fails providing a value to sbb-select
https://github.com/sbb-design-systems/lyne-components/tree/fix/bug-ssr-select-value

Adding a value in SSR tests as in commit 890d8e2 breaks the tests.

Both the connectedCallback and the willUpdate hooks call the `_onValueChanged` method, which asks for the `sbb-option`s to match the provided value against them. The querySelectorAll fails in SSR, so the mismatch in the rendered `displayValue`.

Fix needed to resolve comment https://github.com/sbb-design-systems/lyne-components/pull/2982#discussion_r1728513517

jeripeierSBB:
Closes #3021


src/elements/select/select.ssr.spec.ts
```diff
@@ -9,20 +9,48 @@ import '../option.js';
 describe(`sbb-select ssr`, () => {
   let root: SbbSelectElement;
 
-  beforeEach(async () => {
+  it('renders', async () => {
     root = await ssrHydratedFixture(
       html`
-        <sbb-select placeholder="Placeholder">
+        <sbb-select placeholder="Placeholder" value="1">
```
DavideMininni-Fincons, line 15: Should this be reverted (test with no value)?

DavideMininni-Fincons (approved):
LGTM, thanks!

src/elements/select/select.ssr.spec.ts
jeripeierSBB: Thanks! Will fix this

kyubisation (approved):
LGTM 👍

PR #3040, opened 2024-08-29
feat: keep action elements focusable when disabled

### LINKED ISSUE #2862: [Bug]: focus on disabled buttons
### Preflight Checklist

### Bug type

Accessibility

Components

1.2.0

### Operating system

Windows 10

### Browser / Browser version

Chrome  126.0.6478.127

### Input mode

Mouse/Keyboard

### Additional settings

_No response_

### What happened?

At this moment disabled Buttons are not focusable for blind users. You can overwrite this with the tabIndex but this is not working in the DialogActions.
They should be focusable aswell, so that blind users can get information about the status. 
I think there should be  a prop to make this possible.

1. disabled button on view
2. try to tab to a button
you cant tab to the disabled button

### Relevant log output

_No response_

--------------------------

## Implementation Details

- Remove tabindex logic for disabled buttons (all types of buttons, but not form elements)
- Add accessibility documentation for button components to recommend adding aria references to describe why the button is disabled (from consumer side)

TomMenga:
## Notes to reviewer

I believe that not all the button-like components that previously extended the `SbbDisabledTabIndexActionMixin` should be part of this development. 
The components that, imho, should be excluded are:
- `sbb-popover-trigger`
- `sbb-slider`

Not sure about:
- `sbb-mini-button` => In some cases it should be focusable (mini-button-group) and in others it should not (e.g. when it's part of a form field)
- link and button-links

## TODO
- Decide what to do with arrow navigation (especially for the sbb-menu)

## Preflight Checklist

## Issue

This PR Closes #2862 

## Pull request checklist

See [Review Guidelines](../docs/REVIEW.md) for more information what is checked during review process.

## Changes

Changes in this pull request:

- For a11y purposes, keep button-like elements focusable when disabled

## Browsers

- [ ] Firefox Desktop
- [x] Chrome Desktop
- [ ] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [ ] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [ ] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [ ] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


TomMenga:
> A topic that came to my mind was that we have several components where we implemented arrow key navigation. Normally (as I remember) we exclude the disabled ones (e.g. menu.ts line 153). Maybe this still makes sense, but it would not be the same as arriving with `tab` there. Especially for the menu where both navigations are possible. But there are other components I think.

I believe the menu is the only component allowing both arrow and tab navigation. In general, if there is arrow navigation, you cannot tab between them

jeripeierSBB:
> > A topic that came to my mind was that we have several components where we implemented arrow key navigation. Normally (as I remember) we exclude the disabled ones (e.g. menu.ts line 153). Maybe this still makes sense, but it would not be the same as arriving with `tab` there. Especially for the menu where both navigations are possible. But there are other components I think.
> 
> I believe the menu is the only component allowing both arrow and tab navigation. In general, if there is arrow navigation, you cannot tab between them

Ok, this can be. But also with only arrow-navigation components, should disabled buttons be reached now? Or don't we have any such case?

TomMenga:
> > > A topic that came to my mind was that we have several components where we implemented arrow key navigation. Normally (as I remember) we exclude the disabled ones (e.g. menu.ts line 153). Maybe this still makes sense, but it would not be the same as arriving with `tab` there. Especially for the menu where both navigations are possible. But there are other components I think.
> > 
> > 
> > I believe the menu is the only component allowing both arrow and tab navigation. In general, if there is arrow navigation, you cannot tab between them
> 
> Ok, this can be. But also with only arrow-navigation components, should disabled buttons be reached now? Or don't we have any such case?

I'm not sure about it. Because in lists and groups, like the stepper or tab-group, screen readers allow users to navigate between items (even when disabled). Maybe we could ask for advice from an a11y expert

jeripeierSBB (approved):
LGTM, thank you!

kyubisation (approved):
LGTM 👍 
Nice work! 😃

PR #3055, opened 2024-09-03
fix(option-base): use data-active instead of active attribute

jeripeierSBB:


DavideMininni-Fincons (approved):
LGTM, thanks!

src/elements/option/option/option-base-element.ts
```diff
@@ -44,8 +44,20 @@ export abstract class SbbOptionBaseElement extends SbbDisabledMixin(
     return this.getAttribute('value') ?? '';
   }
 
-  /** Whether the option is currently active. */
-  @property({ reflect: true, type: Boolean }) public active?: boolean;
+  /**
+   * Whether the option is currently active.
+   * @internal
+   */
+  public set active(value: boolean) {
```
kyubisation, line 51: Should this not be deprecated for now?

kyubisation (approved):
LGTM 👍

PR #3080, opened 2024-09-12
test: respect state element for visual focus tests

jeripeierSBB:


src/elements/core/testing/private/visual-regression-snapshot.ts
```diff
@@ -113,8 +113,19 @@ export const visualDiffFocus: VisualDiffState = {
   with(setup: (setup: VisualDiffSetupBuilder) => void | Promise<void>): Mocha.Func {
     return async function (this: Mocha.Context) {
       const builder = await runSetupWithViewport(setup, this.test?.ctx?.['requestViewport']);
-      builder.snapshotElement.focus();
+
+      // We create a focusable element (<a>) right before the state element.
+      // We can tab once to land on the desired element.
+      const link = document.createElement(`a`);
+      link.href = '#';
+      link.classList.add('sbb-screen-reader-only');
```
TomMenga, line 121:
```suggestion
```
Not sure if it's needed since we remove the element before the snapshot is taken

TomMenga (approved):
LGTM 👍 
just a minor thing

src/elements/core/testing/private/visual-regression-snapshot.ts
jeripeierSBB: You are right; it was a consideration to be sure that there is no unwanted layout shift. For 99%, we could remove it, but also after discussing it with Lukas, we decided to keep it.

PR #3081, opened 2024-09-12
fix(sbb-icon): ensure sbb-angular compatibility

kyubisation:
This PR adds svgicon to the observed attributes.


jeripeierSBB (approved):
LGTM

PR #3084, opened 2024-09-13
refactor: replace [rtl] and [ltr] selectors with :dir

### LINKED ISSUE #3050: CSS: Replace [rtl] and [ltr] selectors with :dir
- Replace our CSS [rlt] and [ltr] selector rules with :dir: https://developer.mozilla.org/en-US/docs/Web/CSS/:dir
- Deprecate getDocumentWritingMode and remove usage
- Also replace programmatic usage with css lookup (`const rtl = this.matches(':dir(rtl)')`)

dauriamarco:
Closes #3050


src/elements/core/a11y/arrow-navigation.spec.ts
TomMenga, line 1: If the feedback  on `function getPrevAndNextKeys` is accepted, this file should be fine as it was before

TomMenga (changes requested):
Good job, some minor fixes

TomMenga (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM, thank you

PR #3085, opened 2024-09-13
feat(sbb-file-selector): implement native form support

DavideMininni-Fincons:
Depends from #3101 

NOTE to reviewers: 
The `form-associated-mixin` code is cherry-picked from the `sbb-select` native form support development #3101

TODO:
✅ Create an issue marked as 2.0 to change the `files` property type from `File[]` to a read-only array


src/elements/file-selector/readme.md
```diff
@@ -93,24 +96,31 @@ It's suggested to have a different value for each variant, e.g.:
...
+| -------------------------- | ------- | ----------- | ------------------------------------------------------------- | -------- | -------------- |
+| `formResetCallback`        | public  |             |                                                               | `void`   |                |
+| `formStateRestoreCallback` | public  |             | `state: FormRestoreState \| null, _reason: FormRestoreReason` | `void`   |                |
+| `getFiles`                 | public  |             |                                                               | `File[]` |                |
 
 ## Events
 
 | Name          | Type                  | Description                                                | Inherited From |
 | ------------- | --------------------- | ---------------------------------------------------------- | -------------- |
+| `change`      | `Event`               |                                                            |                |
```
jeripeierSBB, line 121: not sure whether documentation should be added?

src/elements/file-selector/readme.md
```diff
@@ -93,24 +96,31 @@ It's suggested to have a different value for each variant, e.g.:
...
 | `fileChanged` | `CustomEvent<File[]>` | An event which is emitted each time the file list changes. |                |
+| `input`       | `Event`               |                                                            |                |
```
jeripeierSBB, line 123: not sure whether documentation should be added?

src/elements/file-selector/readme.md
```diff
@@ -84,6 +84,9 @@ It's suggested to have a different value for each variant, e.g.:
 ></sbb-file-selector>
 ```
 
+<!-- Override
+  @type value => File[] \| null
```
jeripeierSBB, line 88: needs logic to have effect (only reminder)

src/elements/file-selector/file-selector.ts
```diff
@@ -1,21 +1,25 @@
-import type { CSSResultGroup, TemplateResult } from 'lit';
-import { LitElement, nothing } from 'lit';
-import { customElement, property, state } from 'lit/decorators.js';
+import { type CSSResultGroup, LitElement, nothing, type TemplateResult } from 'lit';
```
jeripeierSBB, line 1: I detected that setting name on file selector is not reflected in form data. But so far I don't know why.

```html
<form>
  <label for="fil">Hello</label>
  <sbb-file-selector
    id="fil"
    name="test"
    data-testid="sbb-file-selector"
    variant="default"
    size="m"
    title-content="Title"
    multiple-mode="default"
    accessibility-label="Select from hard disk"
  ></sbb-file-selector>
</form>
```


While $0 is the form:
```ts
for(const foo of (new FormData($0)).entries()) {console.log(foo)}
```

jeripeierSBB (changes requested):
Great work! Not sure whether I thought of everything, but I have some comments :-)

src/elements/file-selector/file-selector.ts
```diff
@@ -61,22 +67,22 @@ export class SbbFileSelectorElement extends SbbDisabledMixin(LitElement) {
   /** This will be forwarded as aria-label to the native input element. */
   @property({ attribute: 'accessibility-label' }) public accessibilityLabel: string | undefined;
 
-  /** The list of selected files. */
-  @state() private _files?: File[];
```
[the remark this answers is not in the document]
TomMenga: Because `_files` is a mirror of the value property. If `file` is set then `value` is set, which triggers the render

src/elements/file-selector/file-selector.ts
```diff
@@ -61,22 +67,22 @@ export class SbbFileSelectorElement extends SbbDisabledMixin(LitElement) {
...
+  /**
+   * The list of selected files.
+   */
+  public get files(): File[] {
+    return this.value || [];
+  }
+  private set _files(value: File[]) {
```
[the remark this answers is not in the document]
TomMenga: I guess it slightly improves readability since we use the getter a lot. 
It seems bad to read from `files` and setting `value` (the main truth is that it was the easiest way to avoid refactoring all the occurrences)

src/elements/file-selector/file-selector.ts
TomMenga: To appear in the FormData the element needs to have a value (I guess it was empty in your test)

jeripeierSBB: hmm I added a file, which was in the form data, but the name was missing

src/elements/file-selector/readme.md
TomMenga: I'll add a reminder in the description 👍

src/elements/file-selector/file-selector.ts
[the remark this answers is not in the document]
jeripeierSBB: thank you! should have get that :P

src/elements/file-selector/readme.md
TomMenga: I had no idea that this was enough for the event to be added to the doc :|
image

src/elements/file-selector/readme.md
jeripeierSBB: @kyubisation brought the idea of creating a class decorator where we could define a kind of map with overrides instead of doing it in the readme files. What do you think?

src/elements/file-selector/readme.md
jeripeierSBB: oh yeah it does :D. btw. we have long-term plans to also document change and input events on every form component, so it's not wrong at all ;-)

src/elements/file-selector/file-selector.ts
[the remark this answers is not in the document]
TomMenga: Another fair question would be, should we keep the `files` property public?
We could deprecate it in favor of `value` (and just use value everywhere in the code)

src/elements/file-selector/file-selector.ts
TomMenga: Maybe we can try together since I might doing something different

src/elements/file-selector/readme.md
TomMenga: Inspired by MDN, I added a doc which should be fine for every form element

src/elements/file-selector/file-selector.ts
jeripeierSBB: Lets do it in the call

src/elements/file-selector/file-selector.ts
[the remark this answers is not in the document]
TomMenga: Copy the native behavior 
- Add a setter for files, the getter returns hiddenInput.files
- The value returned hiddenInput.value (which is the fake path)

src/elements/file-selector/readme.md
TomMenga: We go for the decorator approach (probably in a following pr)

src/elements/file-selector/file-selector.ts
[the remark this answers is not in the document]
TomMenga: @jeripeierSBB @kyubisation I know we said that we should comply with the native API where possible but I kept the type of the `files` property to `File[]` instead of `FileList`. FileList is read-only and does not have an initializer (would make the component inconvenient for consumers to use)

src/elements/file-selector/file-selector.ts
```diff
@@ -69,15 +67,37 @@ export class SbbFileSelectorElement extends SbbDisabledMixin(
...
+    this._hiddenInput.files = dt.files;
+
+    this.updateFormValue();
   }
-  private set _files(value: File[]) {
-    this.value = value;
+  public get files(): File[] {
+    return this._files;
   }
+  private _files: File[] = [];
```
jeripeierSBB, line 100: Maybe I didn't understand during the call, but is it really possible to hold the files as field? wouldn't it be sufficient to return this._hiddenInput.files in the getter?

src/elements/file-selector/readme.md
```diff
@@ -84,6 +84,9 @@ It's suggested to have a different value for each variant, e.g.:
 ></sbb-file-selector>
 ```
 
+<!-- Override
```
jeripeierSBB, line 87: should we have this here although the logic is not in this PR?

TomMenga: I would like to wait for the select to be ready anyway

src/elements/file-selector/file-selector.ts
```diff
@@ -61,21 +67,43 @@ export class SbbFileSelectorElement extends SbbDisabledMixin(LitElement) {
   /** This will be forwarded as aria-label to the native input element. */
   @property({ attribute: 'accessibility-label' }) public accessibilityLabel: string | undefined;
 
-  /** The list of selected files. */
-  @state() private _files?: File[];
+  /** The path of the first selected file. '' no file is selected */
+  @property({ attribute: false })
+  public override set value(value: string | null) {
+    this._hiddenInput.value = value ?? '';
```
jeripeierSBB, line 73: Just for my understanding: what happens to the form value if assigning a string here (not empty)? Do the files get updated or cleared? In case it gets updated, probably not possible as there is no real file, but only a string?

jeripeierSBB (commented):
LGTM, two questions

src/elements/file-selector/file-selector.ts
TomMenga: If a not-empty string is set, the native input throws an error:
`VM198:1 Uncaught InvalidStateError: Failed to set the 'value' property on 'HTMLInputElement': This input element accepts a filename, which may only be programmatically set to the empty string.`

So essentially, it's an invalid input

src/elements/file-selector/file-selector.ts
```diff
@@ -61,21 +67,43 @@ export class SbbFileSelectorElement extends SbbDisabledMixin(LitElement) {
   /** This will be forwarded as aria-label to the native input element. */
   @property({ attribute: 'accessibility-label' }) public accessibilityLabel: string | undefined;
 
-  /** The list of selected files. */
-  @state() private _files?: File[];
+  /** The path of the first selected file. '' no file is selected */
```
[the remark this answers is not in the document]
TomMenga: There was a typo there. The message was `The path of the first selected file. '' if no file is selected`
Your comment is also correct, not sure which one is preferable though. Any preference?

src/elements/file-selector/file-selector.ts
jeripeierSBB: Ok so we don't additionally have to throw an error as the native element does itself?

src/elements/file-selector/file-selector.ts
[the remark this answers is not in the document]
jeripeierSBB: Maybe take yours but with "empty string ('')". Otherwise it can look a little bit weird I think :)

src/elements/file-selector/file-selector.ts
TomMenga: Yes, that was the idea

jeripeierSBB: ok, I resolve the conversation

TomMenga (approved):

PR #3121, opened 2024-09-30
fix: switch to standard decorators and tighten property types

kyubisation:
This PR switches from experimental decorators to standard decorators. It additionally tightens the types of properties and applies `forceType` where necessary.


src/visual-regression-app/src/components/test-case/image-diff/image-diff.ts
```diff
@@ -23,16 +24,19 @@ const getImageDimension = (img: HTMLImageElement): string =>
...
-export class ImageDiff extends LitElement {
+class ImageDiff extends LitElement {
   public static override styles: CSSResultGroup = style;
 
-  @property() public screenshotFiles?: ScreenshotFiles;
+  @property() public accessor screenshotFiles: ScreenshotFiles | null = null;
 
-  @state() private _baselineDimension?: string;
-  @state() private _failedDimension?: string;
+  @state() private accessor _baselineDimension!: string;
```
jeripeierSBB, line 34: same topic, could we still go with "?" for @ state ?

src/visual-regression-app/src/components/test-case/test-case.ts
```diff
@@ -35,16 +35,17 @@ interface Filter {
...
   public static override styles: CSSResultGroup = style;
 
-  @property() public params?: { componentName: string; testCaseName: string };
+  @property() public accessor params: { componentName: string; testCaseName: string } | null = null;
 
-  @state() private _testCase?: ScreenshotTestCase;
-  @state() private _testCaseIndex: number = -1;
-  @state() private _filter: Filter = {};
-  @state() private _showGlobalDiff = true;
+  @state() private accessor _testCase!: ScreenshotTestCase;
```
jeripeierSBB, line 45: and the last time the same question ;). would `?` still work?

jeripeierSBB (commented):
Great work. Thanks a lot for your effort, especially for linting everything :). Looks good, I have some comments which are basically questions :)

src/elements/core/decorators/omit-empty-converter.ts
```diff
@@ -0,0 +1,15 @@
+import { defaultConverter, type ComplexAttributeConverter } from 'lit';
```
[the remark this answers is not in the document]
kyubisation: Discussed as acceptable.

PR #3124, opened 2024-09-30
build: avoid compiling CSS comments

jeripeierSBB:
SCSS `/*` comments are getting compiled into CSS. Having such comments in mixins generates a lot of redundant output. With the stylelint rule, we can avoid it.


kyubisation (approved):
LGTM 👍

PR #3130, opened 2024-10-01
feat(sbb-paginator): add disabled property

DavideMininni-Fincons:


src/elements/paginator/paginator.scss
```diff
@@ -146,7 +147,32 @@
...
+  &:disabled {
+    --sbb-paginator-page-color: var(--sbb-color-smoke);
+    --sbb-paginator-page-border-width: var(--sbb-border-width-1x);
+    --sbb-paginator-page-border-style: dashed;
+    --sbb-paginator-page-border-color: var(--sbb-color-smoke);
+    --sbb-paginator-page-background-color: var(--sbb-color-milk);
+    --sbb-paginator-page-cursor: default;
+
+    &[data-selected] {
+      --sbb-paginator-page-border-width: var(--sbb-border-width-2x);
```
jeripeierSBB, line 159: hmm I think border-width should only belong to data-selected. would we need an !important? or what is the main problem here?

src/elements/paginator/paginator.scss
```diff
@@ -146,7 +147,32 @@
...
+    }
+
+    @include sbb.if-forced-colors {
+      --sbb-paginator-page-border-color: GrayText;
+      --sbb-paginator-page-color: GrayText;
+      --sbb-paginator-page-background-color: Canvas;
+    }
+
+    :host([negative]) & {
```
jeripeierSBB, line 168: can you move negative before forced-colors?

src/elements/paginator/paginator.scss
DavideMininni-Fincons: Since the disabled rule come after the data-selected, it overrides it
i can add a !important at line 134 to remove this line

jeripeierSBB: I'ts only a little bad design for me because the key value (width-2x) is assigend at two different places. otherwise an additional variable could be used. but maybe important solves it best, or a restructure ?

DavideMininni-Fincons: I possibly fixed it by negating the condition for the -1x border

src/elements/paginator/paginator.scss
DavideMininni-Fincons: Done!

jeripeierSBB (approved):
LGTM :)

mcilurzo:
LGTM. Thank you.

PR #3139, opened 2024-10-02
build: inline fonts during tests

kyubisation:
Similar to what we already do, this PR downloads the fonts before test runs and inlines them in the initial HTML. This avoids flakey font resolution in tests.

### DIFF
--- a/web-test-runner.config.ts
+++ b/web-test-runner.config.ts
@@ -110,7 +110,22 @@ const testRunnerHtml = (
 ): string => `
 <!DOCTYPE html>
 <html lang='en'>
-  <head>
+  <head>${
+    // Although we provide the fonts as base64, we preload the original
+    // files which prevents a bug in Safari rendering special characters.
+    ['Roman', 'Bold', 'Light']
+      .map(
+        (type) => `
+    <link
+      rel="preload"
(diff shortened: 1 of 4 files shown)


web-test-runner.config.ts
```diff
@@ -108,20 +110,22 @@ const testRunnerHtml = (
...
-    )
-    .join('')}
+  <head>
     <link rel="modulepreload" href="/src/elements/core/testing/test-setup.ts" />
     <style type="text/css">${renderStyles()}</style>
+    <style type="text/css">
+      ${preloadedFonts
+        .map(
+          (f) => `
+      @font-face {
```
jeripeierSBB, line 120: In the standard theme we already have font face declarations. Are you sure that those ones don't get loaded beside the ones which are preloaded now?

jeripeierSBB (commented):
Great that you had a look at it!

web-test-runner.config.ts
TomMenga: In theory, the preloaded rule should win since it comes after the standard theme.
Anyway, it's worth a test just to be sure

jeripeierSBB: Already verified, it only loads the right one :)

PR #3143, opened 2024-10-04
fix(sbb-teaser): css variable typo

DavideMininni-Fincons:


kyubisation (approved):
LGTM 👍 
Nice find!

jeripeierSBB:
@kyubisation @DavideMininni-Fincons In this case, it seemed to work also without `#{}`. Otherwise, we would have recognized it much earlier :P.

PR #3144, opened 2024-10-04
fix(sbb-loading-indicator): center component into his box

DavideMininni-Fincons:
NOTE: where does the original `-3em` value come from?


MarioCastigliano (approved):
👍

PR #3169, opened 2024-10-23
build(docs-generate): override type in docs generation

DavideMininni-Fincons:
When developing the native form support for the `sbb-select`, the `SbbFormAssociatedMixinType` received a generic `T` type for `value` property, since it accepts `string | string[]`, while other components accepts only `string`.
This breaks the doc during the generation of the custom-element-manifest ([doc here](https://custom-elements-manifest.open-wc.org/analyzer/plugins/intro/)).
During the `analyzePhase`, the type is correctly set on the manifest, however during the `packageLinkPhase` the `SbbFormAssociatedCheckboxMixin` is resolved and any type in the derived class is overridden with the type in the superclass, as it could be seen [here](https://github.com/open-wc/custom-elements-manifest/blob/94650b64f992c216ff770ac9e398db7c4491c949/packages/analyzer/src/features/post-processing/apply-inheritance.js#L57).
Moreover, if the `value` property is overridden in the derived class, as in the sbb-slider, its member is available in the `moduleDoc.declarations` object during the `analyzePhase`, otherwise it appears only in the `packageLinkPhase`.

Proposal
- a new annotation named `@overrideType` can be added in the component jsDoc
- it's possibile to add more than one annotation per class
- the annotation accepts the form `@overrideType <property to override> - <override type>`
- during the `analyzePhase` an object is added on the class declaration, which holds the information for the override taken by the doc
- during the `packageLinkPhase`, the overwriting is performed and the temporary object is removed from the class declaration; this overrides any change done by the lib based on the superclass because the custom plugins are executed after the build-in ones

Open point
- where to document the new annotation
- annotation syntax 
    - `@overrideType <property to override> - <override type>` 
    - `@overrideType {<property to override>} <override type>` as in event type
    - ??


kyubisation (approved):
LGTM 👍

tools/manifest/custom-elements-manifest.config.js
```diff
@@ -34,6 +36,25 @@ export function createManifestConfig(library = '') {
               classNode = classNode.parent;
             }
           }
+
+          if (node.kind === ts.SyntaxKind.ClassDeclaration) {
+            node.jsDoc?.forEach((doc) => {
```
TomMenga, line 41:
```suggestion
          // A summary comment of what we do
          if (node.kind === ts.SyntaxKind.ClassDeclaration) {
            node.jsDoc?.forEach((doc) => {
```

tools/manifest/custom-elements-manifest.config.js
```diff
@@ -56,6 +77,22 @@ export function createManifestConfig(library = '') {
               if (declaration.name.includes('Base')) {
                 delete declaration.customElement;
               }
+
+              if (declaration[overrideTypeKey]) {
+                declaration[overrideTypeKey].forEach((overrideObj) => {
```
TomMenga, line 82:
```suggestion
              // A summary comment of what we do
              if (declaration[overrideTypeKey]) {
                declaration[overrideTypeKey].forEach((overrideObj) => {
```

TomMenga (approved):
LGTM 👍 
I would just add some comments to avoid forgetting what it does the next time we have to read that

tools/manifest/custom-elements-manifest.config.js
```diff
@@ -78,6 +86,14 @@ export function createManifestConfig(library = '') {
...
               }
 
+              /**
+               * When a generic T type is used in a superclass declaration, it overrides the type defined in derived class
+               * during the doc generation (as the `value` property in the `SbbFormAssociatedMixinType`).
+               * Using the `@overrideType` annotation in the jsDoc's derived class allows to override the type with the correct one.
+               *
+               * In this phase, the script looks for all the `classDeclaration` which have the `overrideTypeKey` property,
+               * and it cycles through them updating the provided property with the provided type.
+               */
```
TomMenga, line 96:
```suggestion
              // Appy the @overrideType
```
Since you extensively explained it in the precedent section, I believe that just a simple connective thought is enough here

PR #3174, opened 2024-10-24
build: fix logic for SSR in React wrapper

DavideMininni-Fincons:
The introduction of @slotstate broke https://github.com/sbb-design-systems/lyne-components/pull/2306 due to the missing call of `new SbbSloteStateController`, so a check on the class decorator has been added with the same logic.


TomMenga (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM. thank you

PR #3210, opened 2024-11-18
test: remove obsolete waitForEvent() method

jeripeierSBB:
BREAKING CHANGE: waitForEvent() method was removed in favor of using EventSpy class


kyubisation (approved):
LGTM 👍

PR #3220, opened 2024-11-20
fix(sbb-option,sbb-autocomplete-grid-option): remove deprecated `active` property

jeripeierSBB:
BREAKING CHANGE: removed `active` property from `sbb-option` and `sbb-autocomplete-grid-option`


kyubisation (approved):
LGTM 👍

PR #3221, opened 2024-11-20
fix(sbb-form-field): remove deprecated `getInputElement()` method

jeripeierSBB:
BREAKING CHANGE: The `getInputElement()` method of the `sbb-form-field` has been removed. Use `inputElement` property as alternative.


kyubisation (approved):
LGTM 👍

PR #3247, opened 2024-11-26
fix(sbb-datepicker): remove deprecated methods and properties

jeripeierSBB:
BREAKING CHANGE: This refactoring introduces multiple breaking changes to the datepicker:
- DateAdapter: return value for invalid dates changed from undefined to null
- Datepicker: removed functions `getAvailableDate()` and `isDateAvailable()`
- Datepicker: moved functions `findPreviousAvailableDate()` and `findNextAvailableDate()` into `SbbDatepickerElement` and removed all params but `date`
- Datepicker: removed properties `dateParser` and `format`, as alternative use custom DateAdapter
- Datepicker: `now` property newly accepts `null` instead of `undefined`
- Datepicker: removed methods `getValueAsDate()` and  `setValueAsDate()`. Use getter/setter `valueAsDate` instead.


kyubisation (approved):
LGTM 👍 
Nice work!

PR #3251, opened 2024-11-26
fix(sbb-tab-group): correctly select a new tab if it is 'active'

### LINKED ISSUE #3212: [Bug]: Tabs not working properly with @for loop
### Preflight Checklist

### Bug type

Functionality

Components

1.14.0

### Operating system

MacOS

### Browser / Browser version

Google Chrome 130.0.6723.117

### Input mode

Touch

### Additional settings

_No response_

### What happened?

We're having a tabgroup to collect data from inputfields within a dialog. The tabgroups are representing languages. As soon as we change the language for the first time, the active tab doesn't change. Instead it displays both contents. If you change the tabs from now on, the tabs will work as expected.
But as soon as you navigate to a different page and come back, the behaviour will be the same as mentioned before.
We discovered, that it only occurs if the tabs are rendered inside a for loop.
We are using angular and figured out it doesn't matter if the for loop is using a signal or not.

An example can be found [here](https://stackblitz.com/edit/stackblitz-starters-oafrye?file=src%2Fapp%2Ffirst-page%2Ffirst-page.component.html).

1. Open the StackBlitz link
2. Open the dialog
3. Click on the FR tab
4. Now you can see both contents are being displayed
5. If you close the dialog you have to refresh or navigate to reproduce the bug again

### Relevant log output

_No response_

TomMenga:
This PR Closes #3212 

The problem is connected to using the `@for` flow in Angular.
Essentially, the `sbb-tab-group` content is instantiated after the group init (which makes `initial-selected-index` unusable). 
It correctly reacts in the `slotChange` but forgets to mark the new eventual `active` tab as the one selected

TODO
- Add tests for late initialization ✅


kyubisation (approved):
LGTM 👍 
Thanks!

PR #3257, opened 2024-11-27
fix: remove useless base class

DavideMininni-Fincons:
The `SbbMiniButtonBaseElement` is used only in the mini-button-component, so it can be safely removed to avoid redundand code.


kyubisation (approved):
LGTM 👍 
Could you add a BREAKING CHANGE notice about removing the base class?

jeripeierSBB (approved):
LGTM

PR #3258, opened 2024-11-27
build: define scss api configuration

kyubisation:
This should facilitate the Vite 6 upgrade.


jeripeierSBB (approved):
LGTM. what do we have to do to get modern? ;)

PR #3284, opened 2024-12-10
fix: improve handling of animation events for zero duration

jeripeierSBB:
This PR enables updating Playwright and therefore provides a solution for the chromium bug https://issues.chromium.org/issues/373506511

Only for autocomplete and select, some code improvements were needed to support zero animation duration.


src/elements/select/select.ts
```diff
@@ -350,8 +355,18 @@ class SbbSelectElement extends SbbUpdateSchedulerMixin(
     this.addEventListener(
       'click',
       (e: MouseEvent) => {
-        this._onOptionClick(e);
-        this._toggleOpening();
+        // Avoid bubbling click to form field, where it would trigger this click again
```
kyubisation, line 358: Same thought

kyubisation (commented):
Looks good so far. Some concerns about events.

src/elements/autocomplete-grid/autocomplete-grid/autocomplete-grid.spec.ts
```diff
@@ -57,6 +57,7 @@ describe(`sbb-autocomplete-grid`, () => {
     `);
     input = formField.querySelector<HTMLInputElement>('input')!;
     element = formField.querySelector<SbbAutocompleteGridElement>('sbb-autocomplete-grid')!;
+    element = formField.querySelector<SbbAutocompleteGridElement>('sbb-autocomplete-grid')!;
```
TomMenga, line 60:
```suggestion
```
little typo?

TomMenga (approved):
Generally lgtm 👍 
Just an error slip and a general discussion

src/elements/alert/alert-group/alert-group.spec.ts
```diff
@@ -17,6 +17,10 @@ describe(`sbb-alert-group`, () => {
     const accessibilityTitle = 'Disruptions';
     const accessibilityTitleLevel = '3';
 
+    const alertOpenedEventSpy = new EventSpy(SbbAlertElement.events.didOpen, null, {
+      capture: true,
+    });
```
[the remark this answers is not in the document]
jeripeierSBB: For the moment we use capture as a workaround for two cases: 
 - Wherever we need to create a spy before the fixture creation (e.g. notification and alert)
 - When there was a parent listener to capture multiple events

While the second could be refactored to single listeners, the first is a little bit more tricky. We could add `@click=` listeners in the template in the hope they are ready early enough (probably yes). With that we couldn't use the EventSpy as is.

As said, it's more like a workaround and we are absolutely open for changes regarding that :)

PR #3285, opened 2024-12-11
feat: provide scrollbar and table styles as CSS classes

### LINKED ISSUE #3271: [Feature request]: Improve scrollbar usability
### Description
I tried using the scrollbar styles, navigated [here](https://lyne-storybook.app.sbb.ch/?path=/docs/styles-scrollbar--docs) and copied an example. Didnt work. After some trial an error I found [this file](https://github.com/sbb-design-systems/lyne-components/blob/main/src/storybook/styles/scrollbar/scrollbar-internal.scss) which seems to contain the classes used in the examples, but they arent publicly exposed, so its not possible to just copy an example.
### Request
Make these classes publicly available (that would be my preferred solution)
or
improve/fix the examples in the storybook. Not too familiar with storybook, but maybe its possible to show multiple source files for an example comparable to how it is in sbb-angular. Or add an example in the general description of how to use the mixin:
```scss
@use '@sbb-esta/lyne-elements' as sbb;
...
.scrollbar-thick-track-visible {
  @include sbb.scrollbar($size: thick, $track-visible: true);
}
```

jeripeierSBB:
Closes #3271


kyubisation (approved):
LGTM 👍

PR #3288, opened 2024-12-11
fix: fix mixin due to wrong manifest generation

DavideMininni-Fincons:
During the generation of the Angular classes in the new repo, the SbbFormAssociatedMixinType is marked with `customElement: true`, which breaks some logic.


kyubisation (approved):
LGTM 👍

TomMenga (approved):
LGTM 👍

PR #3292, opened 2024-12-12
feat: add class generic type param in manifest

DavideMininni-Fincons:
The datepicker and the calendar components use generic type parameter to allow custom implementation of Date, eg.

```
export class SbbDatepickerElement<T = Date> {
    private _min?: T | null;
    // and so on
}
```

This needs to be taken in account in Angular classes generation, so a new param has been added to the manifest during its generation, eg.

```
{
    "kind": "javascript-module",
    "path": "datepicker/datepicker.js",
    "declarations": [
      {
          "kind": "class",
          "description": "Combined with a native input, it displays the input's value as a formatted date.",
          "name": "SbbDatepickerElement",
          "members": [],
          "events": [],
          "attributes": [],
          "superclass": {},
          "classGenerics": "T = Date",        <--------------------------------- new property
          "tagName": "sbb-datepicker",
          "customElement": true
        }
    ],
    exports: []
```

### DIFF
--- a/tools/manifest/custom-elements-manifest.config.js
+++ b/tools/manifest/custom-elements-manifest.config.js
@@ -1,4 +1,5 @@
 const overrideTypeKey = 'overrideType';
+const classGenericsTypeKey = 'classGenerics';
 
 /**
  * Docs: https://custom-elements-manifest.open-wc.org/analyzer/getting-started/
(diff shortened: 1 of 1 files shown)


jeripeierSBB (approved):
LGTM

PR #3294, opened 2024-12-12
fix(sbb-message): support the use of `figure` as image

TomMenga:


src/elements/core/styles/core.scss
```diff
@@ -196,6 +196,14 @@ sbb-lead-container {
   }
 }
 
+// TODO: Move back to the sbb-message components when the global css refactoring happens
```
jeripeierSBB, line 199: Nitpicky 100%, but see my other open PR :D
```suggestion
// TODO: Move back to the sbb-message components when the global CSS refactoring happens
```

src/elements/message/message.scss
```diff
@@ -26,7 +27,8 @@
 }
 
 ::slotted([slot='image']) {
-  margin-block: var(--sbb-message-image-margin-block);
+  display: block;
+  margin-block: var(--sbb-message-image-margin-block) !important; // overrides '.sbb-figure' margin
```
jeripeierSBB, line 31: hmm. In theory the consumer could set a custom margin block which would be overridden from our definitions. otherwise, the design gives this margin as given and therefore shouldn't be overridden. A potential alternative would be to use :where() on the figure styles. But not convinced as well. So my comment is probably completely useless :D

jeripeierSBB (approved):
LGTM, thank you!

src/elements/message/message.scss
TomMenga: The only 'viable' alternative to remove the '!important' is to add in core
```css
  > figure[slot='image'] {
    margin-block: var(--sbb-message-image-margin-block)
  }
```

PR #3295, opened 2024-12-12
fix(sbb-teaser-hero): fix image size on custom width

TomMenga:


jeripeierSBB (approved):
LGTM, thanks for fixing it

PR #3297, opened 2024-12-13
fix: remove any type for Angular wrapper

DavideMininni-Fincons:


TomMenga (approved):
LGTM 👍 
Is this considered a breaking change?

src/visual-regression-app/src/components/test-case/image-diff/fullscreen-diff/fullscreen-diff.ts
```diff
@@ -11,6 +11,8 @@ import '@sbb-esta/lyne-elements/chip-label.js';
 import '@sbb-esta/lyne-elements/radio-button.js';
 import type { SbbRadioButtonGroupElement } from '@sbb-esta/lyne-elements/radio-button/radio-button-group/radio-button-group.js';
 
+type DiffFileType = 'baselineFile' | 'failedFile' | 'diffFile';
```
kyubisation, line 14: We should probably export types we publicly use
```suggestion
export type DiffFileType = 'baselineFile' | 'failedFile' | 'diffFile';
```

kyubisation (approved):
LGTM 👍 
We have done stricter types in the past without marking it as a breaking change, if the expectations mostly aligns. I feel that is the case here as well, so we can skip marking it as a breaking change.

PR #3300, opened 2024-12-16
fix(sbb-select): update displayed value on option label change

### LINKED ISSUE #3298: [Bug]: `sbb-select`'s current value doesn't always properly update
### Preflight Checklist

### Bug type

Functionality

Components

2.0.1

### Operating system

any

### Browser / Browser version

any

### Input mode

None

### Additional settings

_No response_

### What happened?

We have a select in which the user can select a data version (similar to a git branch) they want to load data from. The default branch is called "default" by a different system, which we would like to change to "Hauptversion" or the other translations based on the browser language. If it's not the default version, it should just display the version's name (not internationalized).

Simplified it looks like this:
```
<sbb-select>
    @for (v of versions; track v.name) {
        <sbb-option
          [value]="v.name"
          [selected]="currentVersion() === v.name"
          >{{
            v.name === 'default'
              ? 'Hauptversion-' + currentLanguage()
              : v.name
          }}</sbb-option
        >
    }
</sbb-select>
```

The problem: if we change the language (and the `currentLanguage`-signal updates), it will actually NOT change. Though the option element inside the dropdown actually updates!
I'm honestly not sure if this is a Lyne problem or just some weird Angular change detection stuff happening. Sorry in case it's the latter. :)

Full Stackblitz repro: https://stackblitz.com/edit/angular-at-czthv9sx?file=src%2Fmain.ts
Click the de/fr/it buttons - the text in the paragraph will update as opposed to the text in the select, which is unexpected.

### Relevant log output

_No response_

jeripeierSBB:
Closes #3298

### DIFF
--- a/src/elements/autocomplete-grid/autocomplete-grid-option/autocomplete-grid-option.ts
+++ b/src/elements/autocomplete-grid/autocomplete-grid-option/autocomplete-grid-option.ts
@@ -29,6 +29,7 @@ class SbbAutocompleteGridOptionElement extends SbbOptionBaseElement {
   public static readonly events = {
     selectionChange: 'autocompleteOptionSelectionChange',
     optionSelected: 'autocompleteOptionSelected',
+    optionLabelChanged: 'optionLabelChanged',
   } as const;
 
   protected optionId = autocompleteGridOptionId;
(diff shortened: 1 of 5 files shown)


kyubisation (approved):
LGTM 👍

PR #3301, opened 2024-12-16
build: use custom action for docker slim

kyubisation:


.github/actions/docker-slim/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Docker Slim
+description: Create slim versions of an existing container images
+
+inputs:
+  target:
+    description: The target image
+    required: true
+  tag:
+    description: The target image
```
jeripeierSBB, line 9: description not fitting?

.github/actions/docker-slim/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Setup docker slim
+      run: |
+        mkdir /tmp/ds
+        cd /tmp/ds
+        curl -L -o ds.tar.gz https://github.com/slimtoolkit/slim/releases/download/1.40.11/dist_linux.tar.gz
```
jeripeierSBB, line 18: does a fixed version make sense? or could renovate handle that?

.github/workflows/continuous-integration.yml
```diff
@@ -11,6 +11,28 @@ env:
   IMAGE_REPO_VISUAL_REGRESSION: ghcr.io/${{ github.repository }}/visual-regression
 
 jobs:
+  slim-test:
```
jeripeierSBB, line 14: As you migrated all the usages, is this test still needed?

jeripeierSBB: Ah was probably to see some results in the PR itself. :)

.github/actions/docker-slim/action.yml
kyubisation: Potentially. We will need to see, if Renovate will update this.

PR #3302, opened 2024-12-16
fix: improve null/undefined type handling for Angular wrapper

DavideMininni-Fincons:
Changes are related to Angular wrapper's classes generation.
- `sbb-calendar`: normalized null / undefined types for getters and setters; needs some fix for the `sbb-datepicker-toggle`
- `sbb-navigation-action-common`: updated mixin type
- `sbb-slider`: normalized type
- `sbb-stepper`: some minor fix to normalize null values, which are already considered in the existing logic but not in typings
- `sbb-tag-group`: normalized type
- `sbb-time-input`: add null as type as in the getter
- properties of `SbbStepValidateEventDetails` are now nullable and they are not optional anymore

### DIFF
--- a/src/elements/navigation/common/navigation-action-common.ts
+++ b/src/elements/navigation/common/navigation-action-common.ts
@@ -21,7 +21,7 @@ export declare class SbbNavigationActionCommonElementMixinType {
   public accessor size: SbbNavigationActionSize;
   public get marker(): SbbNavigationMarkerElement | null;
   public get section(): SbbNavigationSectionElement | null;
-  public connectedSection: SbbNavigationSectionElement | null;
+  public connectedSection: SbbNavigationSectionElement | undefined;
 }
 
 // eslint-disable-next-line @typescript-eslint/naming-convention
--- a/src/elements/slider/slider.ts
+++ b/src/elements/slider/slider.ts
@@ -56,8 +56,8 @@ class SbbSliderElement extends SbbDisabledMixin(SbbFormAssociatedMixin(LitElemen
 
   /** Numeric value for the inner HTMLInputElement. */
   @property({ attribute: 'value-as-number', type: Number })
-  public set valueAsNumber(value: number) {
-    this.value = value?.toString();
+  public set valueAsNumber(value: number | null) {
+    this.value = value?.toString() || null;
   }
   public get valueAsNumber(): number | null {
     return Number(this.value);
--- a/src/elements/slider/slider.ts
+++ b/src/elements/slider/slider.ts
@@ -57,7 +57,7 @@ class SbbSliderElement extends SbbDisabledMixin(SbbFormAssociatedMixin(LitElemen
   /** Numeric value for the inner HTMLInputElement. */
   @property({ attribute: 'value-as-number', type: Number })
   public set valueAsNumber(value: number | null) {
-    this.value = value?.toString() || null;
+    this.value = value?.toString() ?? null;
   }
   public get valueAsNumber(): number | null {
     return Number(this.value);
--- a/src/elements/stepper/stepper/stepper.ts
+++ b/src/elements/stepper/stepper/stepper.ts
@@ -37,16 +37,16 @@ class SbbStepperElement extends SbbHydrationMixin(LitElement) {
 
   /** Overrides the behaviour of `orientation` property. */
   @property({ attribute: 'horizontal-from', reflect: true })
-  public set horizontalFrom(value: SbbHorizontalFrom | undefined) {
-    this._horizontalFrom = value && breakpoints.includes(value) ? value : undefined;
+  public set horizontalFrom(value: SbbHorizontalFrom | null) {
+    this._horizontalFrom = value && breakpoints.includes(value) ? value : null;
     if (this._horizontalFrom && this._loaded) {
       this._checkOrientation();
     }
   }
-  public get horizontalFrom(): SbbHorizontalFrom | undefined {
+  public get horizontalFrom(): SbbHorizontalFrom | null {
     return this._horizontalFrom;
   }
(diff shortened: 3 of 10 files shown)


src/elements/tag/tag-group/tag-group.ts
```diff
@@ -91,7 +91,7 @@ class SbbTagGroupElement extends SbbNamedSlotListMixin<SbbTagElement, typeof Lit
         ? this.tags.filter((t) => t.checked).map((t) => t.value)
         : (this.tags.find((t) => t.checked)?.value ?? null);
   }
-  private _value: string | string[] | null = null;
+  private _value: string | (string | null)[] | null = null;
```
jeripeierSBB, line 94: Can that be happen, an array consisting of null?

src/elements/tag/tag-group/tag-group.ts
```diff
@@ -58,7 +58,7 @@ class SbbTagGroupElement extends SbbNamedSlotListMixin<SbbTagElement, typeof Lit
    * If set multiple to true, the value is an array.
    */
   @property()
-  public set value(value: string | string[] | null) {
+  public set value(value: string | (string | null)[] | null) {
```
jeripeierSBB, line 61: Looks weird, do we expect that?

src/elements/stepper/stepper/stepper.ts
```diff
@@ -37,8 +37,8 @@ class SbbStepperElement extends SbbHydrationMixin(LitElement) {
 
   /** Overrides the behaviour of `orientation` property. */
   @property({ attribute: 'horizontal-from', reflect: true })
-  public set horizontalFrom(value: SbbHorizontalFrom) {
-    this._horizontalFrom = breakpoints.includes(value) ? value : undefined;
+  public set horizontalFrom(value: SbbHorizontalFrom | undefined) {
```
jeripeierSBB, line 40: Can't we do it with null (for all changes in this file)?

src/elements/navigation/common/navigation-action-common.ts
```diff
@@ -21,7 +21,7 @@ export declare class SbbNavigationActionCommonElementMixinType {
   public accessor size: SbbNavigationActionSize;
   public get marker(): SbbNavigationMarkerElement | null;
   public get section(): SbbNavigationSectionElement | null;
-  public connectedSection: SbbNavigationSectionElement | null;
+  public connectedSection: SbbNavigationSectionElement | undefined;
```
jeripeierSBB, line 24: What was the problem with null?

src/elements/slider/slider.ts
```diff
@@ -56,8 +56,8 @@ class SbbSliderElement extends SbbDisabledMixin(SbbFormAssociatedMixin(LitElemen
 
   /** Numeric value for the inner HTMLInputElement. */
   @property({ attribute: 'value-as-number', type: Number })
-  public set valueAsNumber(value: number) {
-    this.value = value?.toString();
+  public set valueAsNumber(value: number | null) {
+    this.value = value?.toString() || null;
```
jeripeierSBB, line 60: Could we use ?? ?
```suggestion
    this.value = value?.toString() ?? null;
```

src/elements/tag/tag-group/tag-group.ts
DavideMininni-Fincons: `value` comes from the form-associated mixin, and it's nullable, so technically yes if all the tags have no value

src/elements/tag/tag-group/tag-group.ts
DavideMininni-Fincons: I just kept the same type of the getter because it seems that the manifest takes the prop types from the getter (see the other comment on the tag-group about nullability)

src/elements/stepper/stepper/stepper.ts
DavideMininni-Fincons: change to null would require more changes (eg. default value for horizontal-from, change the SbbStepValidateEventDetails to match the nullable and so on)

src/elements/navigation/common/navigation-action-common.ts
DavideMininni-Fincons: On the manifest the type is `SbbNavigationSectionElement | undefined` because is taken from the mixin getter, while the generated component has null because uses the MixinType;
so when i create the setter in Angular I have a type mismatch

jeripeierSBB: I'd suggest to keep null, but change
`public connectedSection?: SbbNavigationSectionElement` to
`public connectedSection: SbbNavigationSectionElement | null = null;`
would that work?

src/elements/stepper/stepper/stepper.ts
jeripeierSBB: @kyubisation Should we change the API to null (as its the basic goal) now or should we do it later?

src/elements/navigation/common/navigation-action-common.ts
kyubisation: Careful. As this is not a `@property`, it will not be retained during an upgrade, meaning the `= null` could overwrite a previously set value.

src/elements/stepper/stepper/stepper.ts
kyubisation: Ideally we do it now. @DavideMininni-Fincons How much effort would this be?

DavideMininni-Fincons: Minimal, I just have to touch more lines. See commit [cc89dae](https://github.com/sbb-design-systems/lyne-components/pull/3302/commits/cc89daea6f2edc2152c4672688b02c55e9ab4677)

kyubisation (approved):
LGTM 👍

PR #3307, opened 2024-12-17
docs: override prop in manifest

DavideMininni-Fincons:
As already experienced with generics (see 5702592fd40b9382702cee1a28c1729835092afc), the manifest generation has some issue when working with superclasses and mixin.
This time, i noticed that if an element extends a base class, and the base class implements a mixin and overrides some of its property, the overriding is not considered during the manifest generation, because the inheritance chain is traversed from the outer to the inner element and not vice versa as expected.

Eg. all the buttons extend the `SbbButtonBaseElement` class, which implements the `SbbFormAssociatedMixin`.
The `SbbButtonBaseElement` defines via getter/setter the `form` property by overriding the default `form` getter from the mixin; however, the override is not reflected in the manifest, which mark the `form` member as `readonly`, as does with getters. Also the `inheritedFrom` prop refers to the mixin and not to the base class.

This behavior is problematic when the Angular classes are generated, so a manual override must be done; I noticed that the corresponding attribute field is not changed when the inheritance chain it traversed, so it can be used as a 'baseline'.
As a consequence, the `form` jsDoc is overridden in all the button components, taking the correct description from the `SbbButtonBaseElement`.

NOTE: even if this PR fixes the Angular classes generation, it is not possible to automatically fix the type mismatch between get/set, so this should be done manually in the Angular repo.

```ts
  // button-base-element.ts
  
  @property()
  public override set form(value: string) {
    ...
  }
  public override get form(): HTMLFormElement | null {
     ... 
   }
```


src/elements/button/accent-button/readme.md
```diff
@@ -85,7 +85,7 @@ guard against such cases in your component.
...
 | `disabled`            | `disabled`             | public  | `boolean`                 | `false`            | Whether the component is disabled.                                                                                               |
 | `disabledInteractive` | `disabled-interactive` | public  | `boolean`                 | `false`            | Whether the button should be aria-disabled but stay interactive.                                                                 |
-| `form`                | `form`                 | public  | `HTMLFormElement \| null` |                    | Returns the form owner of the internals of the target element.                                                                   |
+| `form`                | `form`                 | public  | `HTMLFormElement \| null` |                    | The <form> element to associate the button with.                                                                                 |
```
kyubisation, line 88: This will break in documentation, as the `<form>` will be rendered as HTML. You can escape this by wrapping `<form>` in backticks: ``` `<form>` ```

kyubisation (approved):
Generally LGTM, apart from the issue with the formatting

jeripeierSBB (approved):
Nice! LGTM

PR #3314, opened 2024-12-19
refactor: move event handlers to constructors

kyubisation:
This is the recommended approach for event handlers for web components.

### DIFF
--- a/tools/generate-component/boilerplate/component.ts
+++ b/tools/generate-component/boilerplate/component.ts
@@ -1,8 +1,6 @@
-import type { CSSResultGroup, PropertyValues, TemplateResult } from 'lit';
-import { html, LitElement, nothing } from 'lit';
-import { customElement, property, state } from 'lit/decorators.js';
-
-import { EventEmitter } from '../core/eventing.js';
+import type { CSSResultGroup, TemplateResult } from 'lit';
+import { html, LitElement } from 'lit';
+import { customElement } from 'lit/decorators.js';
 
 import style from './__noPrefixName__.scss?lit&inline';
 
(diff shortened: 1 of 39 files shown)


tools/generate-component/boilerplate/component.ts
```diff
@@ -36,13 +34,6 @@ export class __nameUpperCase__ extends LitElement {
     this._myEvent.emit();
   }
 
-  public override connectedCallback(): void {
-    super.connectedCallback();
-    const signal = this._abort.signal;
-    this.addEventListener('click', () => this._onClickFn(), { signal });
```
jeripeierSBB, line 42: I suggest to add the click event listener also here to the constructor. We still have the _onClickFn() function which is unused now.

jeripeierSBB (approved):
LGTM, thanks!

tools/generate-component/boilerplate/component.ts
kyubisation: I think I would rather remove it from the template entirely (and maybe more). We have enough examples which can be used as reference.
What do you think?

jeripeierSBB: ok as well. boilerplate usage is rare anyways :)

tools/generate-component/boilerplate/component.ts
```diff
@@ -16,40 +14,12 @@ import style from './__noPrefixName__.scss?lit&inline';
 export class __nameUpperCase__ extends LitElement {
```
jeripeierSBB, line 14: jsdoc of myEventName should be removed to be consistent

PR #3323, opened 2024-12-19
fix(sbb-radio-button-group): sync radios synchronously

### LINKED ISSUE #3316: [Bug]: SbbRadioButtonGroup does not reset old checked radio button after checking another one
### Preflight Checklist

### Bug type

Functionality

Components

_No response_

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

If a sbb-radio-button-group is used and the first sbb-radio-button in group is checked, any selection of the following buttons does not change the value of the sbb-radio-button-group. The reason is the missing reset of the old checked button and the getter returns in every case the old value, because first checked button in radios list is still the initial checked button.

1. use sbb-radio-button-group
2. add 2 sbb-radio-buttons as child
3. check the first one
4. check the second one
5. value of the sbb-radio-button-group returns still the value of the first button in group

### Relevant log output

_No response_

jeripeierSBB:
Closes #3316

### DIFF
--- a/src/elements/radio-button/radio-button-group/radio-button-group.spec.ts
+++ b/src/elements/radio-button/radio-button-group/radio-button-group.spec.ts
@@ -71,13 +71,13 @@ import '../radio-button-panel.js';
         expect(element.value).to.be.equal(radio.value);
       });
 
-      it('should respect disabled radio in value getter', async () => {
+      it('should ignore disabled radios', async () => {
         const radio = radios[0];
         radio.checked = true;
         radio.disabled = true;
         await waitForLitRender(element);
 
-        expect(element.value).to.be.equal('Value one');
+        expect(element.value).to.be.null;
       });
 
       it('should update disabled on children', async () => {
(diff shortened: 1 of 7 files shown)


src/elements/radio-button/radio-button-group/radio-button-group.spec.ts
```diff
@@ -71,13 +71,13 @@ import '../radio-button-panel.js';
...
 
-      it('should ignore disabled radios', async () => {
+      it('should respect disabled radio in value getter', async () => {
         const radio = radios[0];
         radio.checked = true;
         radio.disabled = true;
         await waitForLitRender(element);
 
-        expect(element.value).to.be.null;
+        expect(element.value).to.be.equal('Value one');
```
TomMenga, line 80: Do you think this is the desired behavior? Natively, the FormData would be empty in this case

TomMenga (changes requested):

src/elements/radio-button/radio-button-group/radio-button-group.spec.ts
jeripeierSBB: Is it the same? Here it is the group while native there is no group?

TomMenga: My doubt is, should the `group.value` consider disabled radios?

In this case, both the native and our implementation would return an empty FormData (as if nothing were selected).
Meanwhile the `sbb-radio-group` returns `Value one`. I feel it is inconsistent :/

PR #3340, opened 2025-01-13
build: move icon templates into dedicated div

kyubisation:
This improves usage in test debugging, as the body element is no longer cluttered with template tags.

### DIFF
--- a/web-test-runner.config.ts
+++ b/web-test-runner.config.ts
@@ -147,12 +147,13 @@ const testRunnerHtml = (
       globalThis.testRunScript = '${testFramework}';
     </script>
   </head>
-  <body class="sbb-disable-animation">${preloadedIcons
-    .map(
-      (i) => `
+  <body class="sbb-disable-animation">
+    <div hidden>${preloadedIcons
+      .map(
+        (i) => `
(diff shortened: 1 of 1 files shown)


TomMenga (approved):
Much appreciated  👍

jeripeierSBB (approved):
LGTM, thanks!

web-test-runner.config.ts
```diff
@@ -147,12 +147,13 @@ const testRunnerHtml = (
       globalThis.testRunScript = '${testFramework}';
     </script>
   </head>
-  <body class="sbb-disable-animation">${preloadedIcons
-    .map(
-      (i) => `
+  <body class="sbb-disable-animation">
+    <div>${preloadedIcons
```
jeripeierSBB, line 151: div seems not to work with testing framework. Also, to prevent possible problems with visual tests, I'd recommend to use "hidden" on it. (I think some scrolling tests sometimes are influences by other parts of the page, not only the actual div where it is tested)

kyubisation: No, it works, I just forgot a closing >. I also added the hidden attribute.

PR #3341, opened 2025-01-13
feat(sbb-menu): improvements on arrow navigation

DavideMininni-Fincons:


jeripeierSBB (approved):
Thank you for improving here!

PR #3349, opened 2025-01-14
feat: improve stacking context for overlay components

jeripeierSBB:
By using the Popover API internally, an overlay is always on top, regardless of the stacking context. Currently, we only use the Popover API for the stacking context and are not taking advantage of other possible benefits.


docs/GETTING_STARTED.md
```diff
@@ -348,11 +348,15 @@ to all contained interactive elements.
...
 ### Stacking
 
-As we can't use popover API yet, stacking of overlay context is done manually.
-However, this can interfere with the z-index of your components.
-Therefore, every overlay component provides a CSS variable to override its z-index.
-Additionally, there is the global CSS variable `--sbb-overlay-default-z-index` that has a default z-index of 1000.
-With this, developers have the chance to change the z-index either globally or on component level.
+Our overlay components internally use the Popover API to take advantage of positioning elements on the top layer.
+However, since the Popover API is not fully supported by our list of supported browsers (about 90% as of January 25th),
+stacking the overlay context should still be done manually for older browsers.
```
kyubisation, line 353:
```suggestion
As of January 2025, the browser support for the Popover API is at
[90%](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover#browser_compatibility)
and available for all [our supported browsers](https://github.com/sbb-design-systems/lyne-components?tab=readme-ov-file#-browser-and-screen-reader-support).
If you need to support older browsers, you can either attempt to use a polyfill or
manually configure the stacking order of the overlays.
```

kyubisation (approved):
LGTM 👍 
Nice work! 😄

PR #3355, opened 2025-01-16
fix: display non bubbling events in storybook actions

### LINKED ISSUE #3354: did/will + open/close events not logged into action tab even if the action listener is set
Components that have didOpen/didClose/willOpen/willClose event are not logging this events in the storybook's Action tab, even if the listener is set.
Example: sbb-autocomplete, sbb-accordion
other events are logged, eg. the change and the option-selected in the autocomplete-stories.ts

jeripeierSBB:
Closes #3354

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -97,10 +97,47 @@ const parameters: Parameters = {
   },
 };
 
-const preview: Preview = {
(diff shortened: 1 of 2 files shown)


DavideMininni-Fincons (approved):
LGTM, thanks!

PR #3358, opened 2025-01-17
test(sbb-navigation): fix popover visual spec

TomMenga:
I do not have a proper explanation of for the problem we're having.

In general, it seems that the `visuaDiff` tool is not able to take a screenshot of an element market as `popover` (It says that it is not attached to the DOM)

For the `sbb-navigation` the case is the exact opposite. If I take a snapshot of the default element (which is the parent of the template) I get the "not attached to DOM" error. 
On the contrary, if I set the snapshotElement as the `sbb-navigation` it does not complain

Test made locally on `navigation.visual.spec` and `dialog.visual.spec`

### DIFF
--- a/src/elements/navigation/navigation-section/navigation-section.visual.spec.ts
+++ b/src/elements/navigation/navigation-section/navigation-section.visual.spec.ts
@@ -44,9 +44,12 @@ describe(`sbb-navigation-section`, () => {
           `,
           { padding: '0' },
         );
-        setup.withPostSetupAction(() =>
-          setup.snapshotElement.querySelector<SbbNavigationElement>('sbb-navigation')!.open(),
+
+        setup.withSnapshotElement(
+          setup.snapshotElement.querySelector<SbbNavigationElement>('sbb-navigation')!,
         );
+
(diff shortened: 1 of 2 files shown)


jeripeierSBB:
Hi @TomMenga 
Thanks for your investigation :). I already came to the a conclusion and created the PR #3356 on Thursday.

The problem is not the popover, but element itself did not have an intrinsic size. So the host elements did have a height of 0px. I fixed this width "height: auto". The good stuff, for full overlays like navigation, dialog and overlay we don't have to use 'minHeight' anymore and we can just define the snapshotElement. Maybe you can check my PR as well :)

TomMenga:
> Hi @TomMenga Thanks for your investigation :). I already came to the a conclusion and created the PR #3356 on Thursday.
> 
> The problem is not the popover, but element itself did not have an intrinsic size. So the host elements did have a height of 0px. I fixed this width "height: auto". The good stuff, for full overlays like navigation, dialog and overlay we don't have to use 'minHeight' anymore and we can just define the snapshotElement. Maybe you can check my PR as well :)

Oh, somehow we missed your PR :/
I'll close this as a duplicate of yours

TomMenga:
Duplicate of #3356

PR #3369, opened 2025-01-24
fix: avoid moving content when blocking scrolling

jeripeierSBB:
With previously introduced changes to scroll block, in certain situation the content is slightly moving. This happens when there is a scroll bar visible and some content is aligned next to the scrollbar.
Setting the height to 100dvh ensures that the body is always taking the full available height which is necessary e.g. for the sidebar. 
Setting `position: relative` reverts the previous change where this was removed. This change mainly avoids moving content.


kyubisation (approved):
LGTM 👍

PR #3371, opened 2025-01-27
fix(sbb-radio-button-group): prevent didChange emit from nested group

kyubisation:
Currently if there is a `change` from a radio button inside a radio button group that is nested in another radio button group, the outer radio button group also emits a `didChange` event. This fix checks whether the target of an event is actually directly contained in a group.


jeripeierSBB (approved):
LGTM, cool

TomMenga (approved):
LGTM, nice catch 👍

PR #3382, opened 2025-02-03
feat(sbb-chip-group): initial implementation

### LINKED ISSUE #2761: sbb-chip-group: initial implementation
## Feature Description

We want to implement the `sbb-chip-group` component. It holds a list of `sbb-chip` that the user can add or remove.
It is mainly meant to be used inside an `sbb-form-field`, connected to an `input`.

[Figma](https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?node-id=7491-22&node-type=canvas&m=dev)
[Figma 2](https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?node-id=7575-651925&node-type=instance&m=dev)

## Design Specification
- The user can add new chips by typing and pressing the "enter" key;
- Only one chip is focusable at any moment (generally the last one);
- If the input is empty and the  "delete" key is pressed, the last chip is focused;
- Chips can be removed unless they are `readonly` or `disabled`
- It should be possible to also connect an `sbb-autocomplete` to the same `input` element. When an option is selected, it should also add the corresponding `sbb-chip` (as if the user had pressed `enter`)
- the `sbb-chip-group` is a flex container that wraps horizontally. The `input` should wrap with it which means that it needs to be the last child of the `sbb-chip-container` (ensure the consistency of the `sbb-form-field` logic and style)

### Keyboard interaction
- You can navigate chips through arrow keys;
- Each chip has two navigation steps, one for the chip and the second for the delete button
- If the input is focused (and empty), the 'detete' key moves the focus to the last chip;
- If a chip is focused, the 'detete' key removes it;

## Technical Specification

- `sbb-chip-container` is a form-associated element
- When the user adds a new chip, the `sbb-chip-group` creates a new `sbb-chip` and appends it to the Light DOM
- the `value` is a get/set property that queries for non-disabled `sbb-chip` and returns an array of their values
- `sbb-chip-group` optionally connects to an input through the `for` property. If possible, it connects automatically if in an `sbb-form-field`

## A11y
[Angular material](https://material.angular.io/components/chips/overview) uses 
- `role="grid"` for the container
- `role="row"` for the chip
- Two `role="gridcell"`, one for the label container and one for the delete button

### HTML Usage

```html
<sbb-form-field>
  ...
  <sbb-chip-group name="...">
    <sbb-chip value="value">Label value</sbb-chip>
    <sbb-chip [disabled] [readonly]>Label</sbb-chip>
    ...
    <input />
  </sbb-chip-group>
</sbb-form-field>
```

TomMenga:
## Open topics
- [x] Get translation for [i18nChipDelete](https://github.com/sbb-design-systems/lyne-components/pull/3382/files#diff-4d7532494ddf07bd72f76a2fb3588d4c36384b382940bfd0055606ebefdfbed9)
- [x] To style the `input` slotted into the `sbb-chip-group`, I had to move to `core.scss` some style from the form-field. The `sbb-paginator` uses the `sbb-select` in its shadow-dom (which is not affected by `core.scss`)

## Preflight Checklist

## Issue

This PR Closes #2761

## Pull request checklist

See [Review Guidelines](../docs/REVIEW.md) for more information on what is checked during the review process.

## Changes

Changes in this pull request:

- Created `sbb-chip-group`
- Created `sbb-chip`
- Integration of `sbb-chip-group` and `sbb-autocomplete`

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [ ] Edge Desktop
- [x] Safari Desktop
- [ ] Chrome Mobile
- [ ] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [x] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No

## Jeri's feedback

Some general topics:

- [x] Did you test with the various screen readers?
- [x] If you navigate by keyboard to an "x" and press it, the focus gets lost. I think it should land on an element next to it
- [x] When having the floating-label attribute set on the form field, the form field doesn't correctly recognize the empty state and therefore the label visually collides with the chips. Please add some visual tests as well. Maybe also a story?
- [x] Adding hidden-label on the form-field results in not centered chip. Please add some visual tests as well. Maybe also a story?
- [x] It should be prevented, that a disabled chip can be removed (like in readonly)
- [x] In readonly (maybe also if disabled), navigating between chips seems blocked. maybe discuss what's best, I don't know how it should be best. maybe check Material?
- [x] Comparing to Material, it should probably be possible to set an aria label for the grid element (check Material)
- [x] In Chromium the label seems not correctly centered? https://lyne-components-visual-regression-diff-pr3382.app.sbb.ch/compare/sbb-chip/negative=false_active
- [x] Could it be that the focus outline is missing in the focus state? https://lyne-components-visual-regression-diff-pr3382.app.sbb.ch/compare/sbb-chip/negative=false_delete_focus
- [x] I think we should add stories with formfields of other sizes (especially size s), also check with @mcilurzo if it looks ok.
- [x] When there are long chips (a lot of text) which don't fit into the form field for example, the text breaks into multiple lines. I think we should avoid that and add ellipsis (check with @mcilurzo). Also, please add some visual tests for this case.

Discussion topics:

- [x] In order to provide full screen reader support maybe we have to create a construct like in the autocomplete where the row has its own component. I feel like if we mix roles in shadow DOM and light DOM, there are potential problems for accessibility. I remember some testing which lead to this conclusion. Also looking at snapshot tests of the chip group, you see that not all roles in both browsers are appearing.
- [x] Should adding and removing be handled like in material/sbb-angular? It would probably be more flexible. Like the consumer himself has to build the grid of chips, and we provide an event like in material: (matChipInputTokenEnd)="addTemplateKeyword($event)" (https://material.angular.io/components/chips/overview). See also that they use aria-live messages to announce adding and removal.
Also for what value is shown in the chip: currently you take the "value" attribute, but maybe someone prefers something different. With the Material solution the control is on consumers side.
- [x] Should we allow to configure separators like in sbb-angular? (I think yes to provide feature parity)
- [x] In Angular it should be possible to use with a Set and not only Array. But maybe this could be solved entirely in the angular wrapper.


src/elements/chip-group/chip/chip.ts
```diff
@@ -0,0 +1,111 @@
...
+import { SbbDisabledMixin, SbbNegativeMixin } from '../../core/mixins.js';
+
+import '../../button/mini-button.js';
+
+import style from './chip.scss?lit&inline';
+
+import { i18nChipDelete } from '@sbb-esta/lyne-elements/core/i18n/i18n';
+
+/**
+ * Describe the purpose of the component with a single short sentence.
```
DavideMininni-Fincons, line 17: Remove this placeholder text

DavideMininni-Fincons (changes requested):
Generally looks good, some detail to fix

src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -30,6 +30,11 @@ import style from './autocomplete-base-element.scss?lit&inline';
  */
 const ariaRoleOnHost = isSafari;
 
+/**
```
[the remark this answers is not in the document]
TomMenga: It is a peculiar situation because that event is emitted on the 'input' element, similar to the 'input/change' events.
Plus, it's an event we use internally. Probably, we don't want to expose it to consumers

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,332 @@
...
+
+    // Subtract from the new 'value' what was already present
+    // The result are the new chips to add (handle duplicates)
+    const toAdd = [...value];
+    for (const c of oldValue) {
+      if (toAdd.includes(c)) {
+        toAdd.splice(toAdd.indexOf(c), 1);
+      }
+    }
+    toAdd.forEach((c) => this._createChipElement(c));
```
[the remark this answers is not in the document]
TomMenga: Unfortunately not :/ 
Users can add duplicates and the `Set` does not allow them by definition

jeripeierSBB (changes requested):
Great work! I think we need to discuss some basic questions. So with this first review I did not dive deep into the code but I have some feedbacks on a higher level. If we are going to discuss them, I'd like to involve @kyubisation as well.

Some general topics:
 - [ ] Did you test with the various screen readers?
 - [ ] If you navigate by keyboard to an "x" and press it, the focus gets lost. I think it should land on an element next to it
 - [ ] When having the `floating-label` attribute set on the form field, the form field doesn't correctly recognize the empty state and therefore the label visually collides with the chips. Please add some visual tests as well. Maybe also a story?
 - [ ] Adding `hidden-label` on the form-field results in not centered chip. Please add some visual tests as well. Maybe also a story?
 - [ ] It should be prevented, that a disabled chip can be removed (like in readonly)
 - [ ] In readonly (maybe also if disabled), navigating between chips seems blocked. maybe discuss what's best, I don't know how it should be best. maybe check Material?
 - [ ] Comparing to Material, it should probably be possible to set an aria label for the grid element (check Material)
 - [ ] In Chromium the label seems not correctly centered? https://lyne-components-visual-regression-diff-pr3382.app.sbb.ch/compare/sbb-chip/negative=false_active
 - [ ] Could it be that the focus outline is missing in the focus state? https://lyne-components-visual-regression-diff-pr3382.app.sbb.ch/compare/sbb-chip/negative=false_delete_focus
 - [ ] I think we should add stories with formfields of other sizes (especially size s), also check with @mcilurzo if it looks ok.
 - [ ] When there are long chips (a lot of text) which don't fit into the form field for example, the text breaks into multiple lines. I think we should avoid that and add ellipsis (check with @mcilurzo). Also, please add some visual tests for this case.
 
Discussion topics:
- [ ] In order to provide full screen reader support maybe we have to create a construct like in the autocomplete where the row has its own component. I feel like if we mix roles in shadow DOM and light DOM, there are potential problems for accessibility. I remember some testing which lead to this conclusion. Also looking at snapshot tests of the chip group, you see that not all roles in both browsers are appearing.
- [ ] Should adding and removing be handled like in material/sbb-angular? It would probably be more flexible. Like the consumer himself has to build the grid of chips, and we provide an event like in material: (matChipInputTokenEnd)="addTemplateKeyword($event)" (https://material.angular.io/components/chips/overview). See also that they use aria-live messages to announce adding and removal.
Also for what value is shown in the chip: currently you take the "value" attribute, but maybe someone prefers something different. With the Material solution the control is on consumers side.
- [ ] Should we allow to configure separators like in sbb-angular? (I think yes to provide feature parity)
- [ ] In Angular it should be possible to use with a Set and not only Array. But maybe this could be solved entirely in the angular wrapper.

src/elements/form-field/form-field/form-field.ts
```diff
@@ -203,10 +203,12 @@ class SbbFormFieldElement extends SbbNegativeMixin(SbbHydrationMixin(LitElement)
...
    * It is used internally to assign the attributes of `<input>` to `_id` and `_input` and to observe the native readonly and disabled attributes.
    */
-  private _onSlotInputChange(event: Event): void {
-    const newInput = (event.target as HTMLSlotElement)
-      .assignedElements()
-      .find((e): e is HTMLElement => this._supportedInputElements.includes(e.localName));
+  private _onSlotInputChange(): void {
+    // Find the slotted 'supportedInputElement', even if it's nested
+    const newInput = this.querySelector(
+      `:not(slot):where(${this._supportedInputElements.join(',')})`,
```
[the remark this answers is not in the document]
TomMenga: I added it to exclude input elements slotted in other sections of the form field.
But now that I think of it, this would not work in cases where the input is nested in a slot different from the main one.
I believe that, since the low chance of the first case to happen, we can remove the `:not(slot)`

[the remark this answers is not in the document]
jeripeierSBB: But the element e.g. the input is never the slot element itself? so :not(slot) doesn't do anything because it's always true. But yes, if you remove it anyways, then the discussion has its end :)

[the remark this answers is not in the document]
TomMenga: you're right 😅  It should have been `:not([slot])`

src/elements/form-field/form-field/form-field.ts
```diff
@@ -203,10 +203,12 @@ class SbbFormFieldElement extends SbbNegativeMixin(SbbHydrationMixin(LitElement)
   /**
    * It is used internally to assign the attributes of `<input>` to `_id` and `_input` and to observe the native readonly and disabled attributes.
    */
-  private _onSlotInputChange(event: Event): void {
-    const newInput = (event.target as HTMLSlotElement)
-      .assignedElements()
-      .find((e): e is HTMLElement => this._supportedInputElements.includes(e.localName));
+  private _onSlotInputChange(): void {
+    // Find the slotted 'supportedInputElement', even if it's nested
+    const newInput = this.querySelector(
```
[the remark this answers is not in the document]
TomMenga: I need the casting for the ` | undefined` part :/

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,332 @@
...
+ *
+ * @event {CustomEvent<void>} change - Notifies that the component's value has changed.
+ * @event {CustomEvent<void>} input - Notifies that the component's value has changed.
+ * @slot - Use the unnamed slot to add `sbb-chip` elements.
+ * @overrideType value - string[] | null
+ */
+export
+@customElement('sbb-chip-group')
+@hostAttributes({
+  tabindex: '0',
```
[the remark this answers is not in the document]
TomMenga: The nice thing about this implementation is that I don't need to update the tabindex on the chips. They are always focusable but not in the tab order. 
When the `sbb-chip-group` is focused, I can easily decide to which element forward the focus

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,406 @@
...
+    const oldValue = this.value;
+
+    // Subtract from 'oldValue' the new 'value'
+    // The result are the chips to remove (handle duplicates)
+    const toRemove = [...oldValue];
+    for (const c of value) {
+      if (toRemove.includes(c)) {
+        toRemove.splice(toRemove.indexOf(c), 1);
+      }
+    }
```
DavideMininni-Fincons, line 71: Could this be replaced by:
```suggestion
   const toRemove = oldValue.filter(item => !newValue.includes(item));
```

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,406 @@
...
+    );
+
+    // Subtract from the new 'value' what was already present
+    // The result are the new chips to add (handle duplicates)
+    const toAdd = [...value];
+    for (const c of oldValue) {
+      if (toAdd.includes(c)) {
+        toAdd.splice(toAdd.indexOf(c), 1);
+      }
+    }
```
DavideMininni-Fincons, line 85: Could this be replaced by:
```suggestion
    const toAdd = newValue.filter(item => !oldValue.includes(item));
```

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,406 @@
...
+      const eventDetail: SbbChipInputTokenEndEventDetails = {
+        origin: origin,
+        value: token,
+        label: undefined,
+        setValue: (value: string) => (eventDetail.value = value),
+        setLabel: (label: string) => (eventDetail.label = label),
+      };
+
+      if (!this._chipInputTokenEnd.emit(eventDetail)) {
+        return; // event prevented; do nothing (the consumer has to create the chip)
```
DavideMininni-Fincons, line 315: Question: 
if returned here, the clearing of the input and the event emission is on consumers too, right?
i feel this should be clarified in the readme

src/elements/chip-group/chip-group/readme.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+## Slots
+
+Use the unnamed slot to provide the `sbb-chip` and the `input` field.
+
+## States
+
+The `sbb-chip-group` has a `disabled` and a `readonly` state and reacts to the respective `input` properties.
+The `disabled`/`readonly` property is proxied to the slotted `sbb-chip`.
```
DavideMininni-Fincons, line 24:
```suggestion
The `disabled`/`readonly` properties are proxied to the slotted `sbb-chip`s.
```

src/elements/chip-group/chip/chip.ts
```diff
@@ -0,0 +1,121 @@
...
+          aria-label=${this.accessibilityLabel || nothing}
+          @click=${() => this._chipLabel().focus()}
+        >
+          <span class="sbb-chip__label">
+            <slot>${this.value}</slot>
+          </span>
+        </div>
+        <div role="gridcell">
+          <sbb-mini-button
+            tabindex=${!this.disabled ? '-1' : nothing}
```
DavideMininni-Fincons, line 104: Can this be removed, considering the override done in firstUpdated method?

src/elements/chip-group/chip/readme.md
```diff
@@ -0,0 +1,37 @@
+The `sbb-chip` is a component meant to be used in combination with the [sbb-chip-group](/docs/elements-sbb-chip-group-sbb-chip-group--docs)
```
DavideMininni-Fincons, line 1:
```suggestion
The `sbb-chip` is a component meant to be used in combination with the [sbb-chip-group](/docs/elements-sbb-chip-group-sbb-chip-group--docs) to display . . . . 
```

DavideMininni-Fincons: Also states and negative are missing

DavideMininni-Fincons (commented):
Looks good, some comments about minor questions and details

src/elements/chip-group/chip-group/chip-group.ts
TomMenga: Same as previous feedback

src/elements/chip-group/chip-group/chip-group.ts
TomMenga: Yes, all the default behaviors are prevented if the event is
I'll add a note in the readme

src/elements/chip-group/chip/chip.ts
TomMenga: Nope, the override only works during initialization. I need the binding for eventual post-init changes

src/elements/form-field/form-field/form-field.ts
```diff
@@ -399,7 +401,9 @@ class SbbFormFieldElement extends SbbNegativeMixin(SbbHydrationMixin(LitElement)
   }
 
   private _isInputEmpty(): boolean {
-    if (this._input instanceof HTMLInputElement) {
+    if (this.querySelector('sbb-chip-group')) {
```
jeripeierSBB, line 404:
```suggestion
Because two times used, I'd create a var.
    const chipGroup = this.querySelector('sbb-chip-group')
    if (chipGroup) {
```

src/elements/form-field/form-field/form-field.ts
```diff
@@ -399,7 +401,9 @@ class SbbFormFieldElement extends SbbNegativeMixin(SbbHydrationMixin(LitElement)
...
+      return this._isInputValueEmpty() && this.querySelector('sbb-chip-group')!.value.length === 0;
```
jeripeierSBB, line 405: with my suggestion, the "!" should be obsolete.
```suggestion
      return this._isInputValueEmpty() && chipGroup.value.length === 0;
```

jeripeierSBB, line 405: Do we have a test for the new logic? Not sure

jeripeierSBB (changes requested):
Excellent PR quality! Almost done I think. Only a few minor things. A test needs to be fixed. After that I'd also like to check the visual Diff.

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,391 @@
...
+  public static override styles: CSSResultGroup = style;
+  public static readonly events = {
+    input: 'input',
+    change: 'change',
+    chipInputTokenEnd: 'chipInputTokenEnd',
+  } as const;
+
+  /** Value of the form element. */
+  @property({ type: Array })
+  public override set value(value: string[] | null) {
```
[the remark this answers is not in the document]
TomMenga: It doesn't hurt, I guess. 
The thing is that there might be FE framework that threats object binding in templates differently (they might set the attribute with a json string)

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,391 @@
...
+  }
+
+  /** Return the list of chip elements **/
+  private _chipElements(): SbbChipElement[] {
+    return Array.from(this.querySelectorAll?.('sbb-chip') ?? []);
+  }
+
+  /** Return the list of enabled chip elements **/
+  private _enabledChipElements(): SbbChipElement[] {
+    return Array.from(this.querySelectorAll?.('sbb-chip:not([disabled])') ?? []);
```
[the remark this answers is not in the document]
TomMenga: Since the `sbb-chip` is not a form-associated element it does not work

src/elements/chip-group/chip/chip.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  protected override render(): TemplateResult {
+    return html`
+      <div class="sbb-chip">
+        <div
+          class="sbb-chip__label-wrapper"
+          role="gridcell"
+          tabindex=${!this.disabled ? '-1' : nothing}
+          aria-label=${this.accessibilityLabel || nothing}
+          @click=${() => this._chipLabel().focus()}
```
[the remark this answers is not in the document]
TomMenga: Strange, i thought it was needed but, apparently, not anymore. Maybe it was necessary in a previous version

src/elements/chip-group/chip/chip.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  /** The value of chip. Will be used as label. */
+  @forceType() @property() public accessor value: string = '';
+
+  /** Whether the component is readonly */
+  @forceType()
+  @property({ type: Boolean, reflect: true })
+  public accessor readonly: boolean = false;
+
+  /** This will be forwarded as aria-label to the inner anchor element. */
```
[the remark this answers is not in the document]
TomMenga: whops 😅

src/elements/chip-group/chip-group/chip-group.visual.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+
+    it(
+      `floating-label=true`,
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(template({ floatingLabel: true }));
+      }),
+    );
+
+    it.only(
```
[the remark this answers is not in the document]
TomMenga: I noticed it too :/
I often use 'only' to focus on a test and it might happen to slip. (I'll write a lint rule when I have time)

src/elements/chip-group/chip-group/chip-group.visual.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+const template = (
+  args: Partial<{
+    negative: boolean;
+    disabled: boolean;
+    readonly: boolean;
+    longLabel: boolean;
+    hiddenLabel: boolean;
+    floatingLabel: boolean;
+    size: string;
```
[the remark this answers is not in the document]
TomMenga: Nice! I never thought about that

src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -81,6 +82,15 @@ abstract class SbbAutocompleteBaseElement extends SbbNegativeMixin(
   protected abstract panelRole: string;
   /** @deprecated No longer used internally. */
   protected abort = new SbbConnectedAbortController(this);
+  private _originResizeObserver = new ResizeController(this, {
```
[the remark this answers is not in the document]
TomMenga: would it be okay if I added a visual spec for it?
I feel like that simple assertion is just a 'half-check'

src/elements/form-field/form-field/form-field.ts
TomMenga: For my taste, I would not but I'll do it anyway

src/elements/form-field/form-field/form-field.ts
TomMenga: we do have the visual spec for it but I'm often wondering if a spec test is also needed for cases covered by visual tests.
Do you have an opinion on that?

src/elements/chip-group/chip/chip.ts
```diff
@@ -0,0 +1,121 @@
...
+          <span class="sbb-chip__label">
+            <slot>${this.value}</slot>
+          </span>
+        </div>
+        <div role="gridcell">
+          <sbb-mini-button
+            tabindex=${!this.disabled ? '-1' : nothing}
+            class="sbb-chip__delete"
+            icon-name="cross-tiny-small"
+            aria-label=${`${i18nChipDelete[this._language.current]} ${this.value}`}
```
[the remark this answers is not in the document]
TomMenga: True, nice catch!

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,391 @@
...
+      this._deleteChip(ev.target as SbbChipElement),
+    );
+
+    this.addEventListener('focus', () => this._focusChip());
+    this.addEventListener('keydown', (ev) => this._onChipKeyDown(ev));
+  }
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    this.setAttribute('data-size', this.closest('sbb-form-field')?.size ?? 'm');
```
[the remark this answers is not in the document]
TomMenga: somehow I missed this new thing :/

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -0,0 +1,391 @@
...
+ * @slot - Use the unnamed slot to add `sbb-chip` elements.
+ * @overrideType value - string[] | null
+ */
+export
+@customElement('sbb-chip-group')
+@hostAttributes({
+  tabindex: '0',
+  role: 'grid',
+})
+class SbbChipGroupElement extends SbbDisabledMixin(
```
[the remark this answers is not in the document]
TomMenga: Added both the required and the validation logic :)

src/elements/chip-group/chip-group/chip-group.ts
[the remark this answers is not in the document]
TomMenga: Reading it today, I might have misinterpreted the feedback.
To answer your question: no, Lit accepts and converts JSON objects for `Array` and `Object` type

src/elements/autocomplete/autocomplete-base-element.ts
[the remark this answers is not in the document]
jeripeierSBB: for sure. At the end it should just somehow be covered :)

src/elements/form-field/form-field/form-field.ts
jeripeierSBB: no, not a fixed opinion. I myself decide case by case. So I let it up to you. (visual specs aren't considered for test coverage, that's a little drawback, but not important)

src/elements/chip-group/chip-group/chip-group.ts
```diff
@@ -213,6 +238,7 @@ class SbbChipGroupElement extends SbbDisabledMixin(
     this.toggleAttribute('data-empty', this.value.length === 0);
     this._reactToInputChanges();
     this.updateFormValue();
+    this._setValidity();
```
jeripeierSBB, line 241: @kyubisation changed the logic for validation and now there is a shouldValidate() method to override. Can you adapt those changes?
the "requestUpdate" method should not be used anymore.

src/elements/chip-group/chip-group/chip-group.visual.spec.ts
```diff
@@ -69,14 +70,14 @@ describe('sbb-chip-group', () => {
       it(
         `size=${size}`,
         visualDiffDefault.with(async (setup) => {
-          await setup.withFixture(template({ size: size }));
+          await setup.withFixture(template({ size: size as SbbFormFieldElement['size'] }));
```
jeripeierSBB, line 73: Nitpicky: Instead of casting two times the usages, it could also be casted on line 69 where it is created.

src/elements/chip-group/chip-group/readme.md
```diff
@@ -114,6 +114,16 @@ In this scenario, selecting an option will create a new chip using the option va
 </sbb-form-field>
 ```
 
+### Custom 'separator-keys'
+
+By default, the `sbb-chip-group` create a new chip on `Enter` key press.
```
jeripeierSBB, line 119:
```suggestion
By default, the `sbb-chip-group` creates a new chip on `Enter` key press.
```

src/elements/chip-group/chip-group/readme.md
```diff
@@ -114,6 +114,16 @@ In this scenario, selecting an option will create a new chip using the option va
...
+
+Consumers can customize the array of [keys](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#special_values) that will trigger the chip creation by using the `SeparatorKeys` property.
```
jeripeierSBB, line 121: I think we should use the real property name if talking about property
```suggestion
Consumers can customize the array of [keys](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#special_values) that will trigger the chip creation by using the `separatorKeys` property.
```

src/elements/option/option/option.ssr.spec.ts
```diff
@@ -33,7 +33,7 @@ describe(`sbb-option ssr`, () => {
         html`
           <div>
             <input id="autocomplete-input" />
-            <sbb-autocomplete trigger="autocomplete-input">
+            <sbb-autocomplete trigger="autocomplete-input" origin="autocomplete-input">
```
jeripeierSBB, line 36: Is it required to work? Because it shouldn't I think

jeripeierSBB (changes requested):
Still almost done ;-)

src/elements/chip-group/chip-group/chip-group.ts
TomMenga: Nice improvement!

jeripeierSBB (approved):
LGTM! 🎉

src/elements/core/i18n/i18n.ts
```diff
@@ -768,3 +768,17 @@ export const i18nDateMax = (max: string): Record<string, string> => ({
...
+
+export const i18nChipDelete: Record<string, string> = {
+  de: 'TODO',
+  en: "Press the 'Delete' button to remove the chip",
+  fr: 'TODO',
+  it: "Premi il tasto 'Canc' per rimuovere questa chip",
+};
+
+export const i18nChipGroupInputDescription: Record<string, string> = {
+  de: 'TODO',
```
jeripeierSBB, line 780:
```suggestion
  de: 'Selektierte Elemente:',
```

src/elements/core/i18n/i18n.ts
```diff
@@ -768,3 +768,17 @@ export const i18nDateMax = (max: string): Record<string, string> => ({
   fr: `La date ne doit pas être postérieure au ${max}.`,
   it: `La data non deve essere successiva al ${max}.`,
 });
+
+export const i18nChipDelete: Record<string, string> = {
+  de: 'TODO',
+  en: "Press the 'Delete' button to remove the chip",
+  fr: 'TODO',
```
jeripeierSBB, line 775: Please verify somehow
```suggestion
  fr: 'Appuyez sur la touche 'Del' pour supprimer cette puce',
```

jeripeierSBB (changes requested):
Very nice!

src/elements/chip-group/chip/chip.ts
```diff
@@ -20,7 +21,7 @@ import style from './chip.scss?lit&inline';
 export
 @customElement('sbb-chip')
 @hostAttributes({
-  role: 'row',
+  role: 'option',
```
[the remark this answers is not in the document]
jeripeierSBB: See e.g. sidebar (which is not a form control as well)

```ts
  public constructor() {
    super();
    const internals: ElementInternals = this.attachInternals();
    /** @internal */
    internals.role = 'navigation';
  }
```

src/elements/chip-group/chip/chip.ts
```diff
@@ -36,86 +37,66 @@ class SbbChipElement extends SbbNegativeMixin(SbbDisabledMixin(LitElement)) {
...
   private _language = new SbbLanguageController(this);
 
   public override click(): void {
     if (this.disabled) {
       return;
     }
-    const chipLabel = this._chipLabel();
-    chipLabel.focus();
-    chipLabel.click();
+    this.focus();
```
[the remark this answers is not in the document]
TomMenga: it was part of the two-focus-step implementation. We can remove it now

src/elements/core/i18n/i18n.ts
TomMenga: Did you forgot the german one?

src/elements/core/i18n/i18n.ts
```diff
@@ -770,15 +770,15 @@ export const i18nDateMax = (max: string): Record<string, string> => ({
 });
 
 export const i18nChipDelete: Record<string, string> = {
-  de: 'TODO',
-  en: "Press the 'Delete' button to remove the chip",
-  fr: 'TODO',
-  it: "Premi il tasto 'Canc' per rimuovere questa chip",
+  de: ', Drücken Sie die Entfernen-Taste um den Chip zu löschen',
```
jeripeierSBB, line 773: Shouldn't we add the comma in the code instead of the strings?

jeripeierSBB (approved):
I think we are done :-) congratulations!

PR #3385, opened 2025-02-05
fix: close overlay stack with Escape key

### LINKED ISSUE #3272: [Bug]: Esc closes Dialog, when Autocomplete drop-down is focused
### Preflight Checklist

### Bug type

Functionality

Components

_No response_

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

If a sbb-autocomplete is used in a sbb-dialog and Esc-key is pressed while sbb-autocomplete option is focused the complete sbb-dialog closes not only the sbb-autocomplete drop-down. Maybe the keyboard event is not stopped?

1. open a sbb-autocomplete in a sbb-dialog
2. press Esc-key while focusing sbb-autocomplete option
3. dialog is closed instead of closing only the sbb-autocomplete

### Relevant log output

_No response_

DavideMininni-Fincons:
Closes #3272 

The proposed implementation revolves around a new base class which maintains a registry of overlays.

A controller has been added and each time an overlay is opened, its instance is added to registry;
each time the overlay is closed is removed.
To avoid dead elements in the registry, eg. changing page, the disconnect is called also in the 'hostDisconnected' hook.


src/elements/core/base-elements/open-close-escapable-element.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (overlayStack.length) {
+      overlayStack.pop();
+    }
+  }
+
+  private _cleanStack(): void {
+    overlayStack
+      .map((overlay: WeakRef<SbbOpenCloseEscapableElement>) => overlay.deref()?.id)
+      .forEach((overlayId: string | undefined, index: number) => {
+        if (overlayId && !document.getElementById(overlayId)) {
```
jeripeierSBB, line 52: Is it really necessary to use an Id lookup? This would break shadow DOM usage. Couldn't we just use the element which is hold in overlayStack?

src/elements/core/base-elements/open-close-escapable-element.ts
```diff
@@ -0,0 +1,57 @@
...
+      overlayStack.at(-1)?.deref()?.close();
+    }
+  });
+}
+
+/**
+ * Class for overlay components which extends SbbOpenCloseBaseElement
+ * and adds the functionality to close the overlay pressing the `Escape` button.
+ */
+export abstract class SbbOpenCloseEscapableElement extends SbbOpenCloseBaseElement {
```
jeripeierSBB, line 32: I'm generally insecure whether I consider it a good idea to have a base class for this logic. In our overlay components we use a lot of stuff, like InertController, FocusTrap, ... All of this stuff is implemented as controller or similar which manually gets called at the right moment (like here the call to super.open(). Wouldn't a controller fit better into our existing solutions?

src/elements/dialog/dialog/dialog.stories.ts
```diff
@@ -373,14 +375,25 @@ const NestedTemplate = ({
...
+        <p>
+          Nested dialog content. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+          eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
+          nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute
+          irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla
+          pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia
+          deserunt mollit anim id est laborum.
+        </p>
+        <sbb-form-field>
+          <label>Testing 'Escape' keydown with multiple overlay</label>
```
jeripeierSBB, line 388: If we keep it, maybe modify phrase to have something without "testing"?

src/elements/core/base-elements/open-close-escapable-element.ts
DavideMininni-Fincons: The first implementation was with a controller, I could easily revert it

src/elements/core/base-elements/open-close-escapable-element.ts
DavideMininni-Fincons: I added this code when i noticed that opening a dialog and changing to a different story maintains the opened element in the stack (since the close method is not called and adding hooks like disconnectedCallback on unload won't work); 
so technically if you do this several time, the stack would be a bit bloated, and there would be several elements 'on hold'

my idea was that when you open another dialog in the new page, the old element should not be in the dom anymore and it would be removed;
but as you mentioned and as mentioned in the pr description, it breaks shadowDOM and if the id is reused it won't work, so i would appreciate any other solution :)

DavideMininni-Fincons: Apparently, using the 'hostDisconnected' works, so after some testing I reverted to a more basic implementation

src/elements/dialog/dialog/dialog.stories.ts
```diff
@@ -368,14 +370,25 @@ const NestedTemplate = ({
...
       )}
-      <sbb-dialog-content
-        >Nested dialog content. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
-        eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis
-        nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute
-        irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.
-        Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit
-        anim id est laborum.</sbb-dialog-content
-      >
+      <sbb-dialog-content>
```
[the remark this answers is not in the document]
DavideMininni-Fincons: All the involved components had existing spec test for closing with Esc, which not fail now without changing them;
should i add a specific one for the controller, eg. as done for inert-controller?

src/elements/core/controllers/overlay-controller.ts
```diff
@@ -0,0 +1,39 @@
...
+if (!isServer) {
+  window.addEventListener('keydown', (event: KeyboardEvent) => {
+    if (event.key === 'Escape' && !event.defaultPrevented) {
+      event.preventDefault();
+      overlayStack.at(-1)?.close();
+    }
+  });
+}
+
+export class SbbOverlayController implements ReactiveController {
```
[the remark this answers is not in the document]
DavideMininni-Fincons: I agree; what about `SbbOverlayEscapeClosableController` ?
@kyubisation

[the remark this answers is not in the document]
kyubisation: I think `SbbOverlayCloseController` or `SbbOverlayEscapeClosableController` would work.
My suggestion would be `SbbEscapableOverlayController`.

src/elements/dialog/dialog/dialog.spec.ts
```diff
@@ -495,4 +500,110 @@ describe('sbb-dialog', () => {
...
+      const input = root.querySelector<HTMLInputElement>('input')!;
+      const autocomplete = root.querySelector<SbbAutocompleteElement>('sbb-autocomplete')!;
+
+      // the overlays are all closed
+      expect(dialog).to.have.attribute('data-state', 'closed');
+      expect(dialog).not.to.match(':popover-open');
+      expect(nestedDialog).to.have.attribute('data-state', 'closed');
+      expect(nestedDialog).not.to.match(':popover-open');
+      expect(input).to.have.attribute('aria-expanded', 'false');
+      expect(autocomplete).not.to.match(':popover-open');
```
jeripeierSBB, line 556: very nitpicky, I think asserting only the data-state is enough (you only want to be sure that you are in the right state)

jeripeierSBB (approved):
LGTM thank you! Maybe @kyubisation, can you check if the logic really fits our needs now?

kyubisation (commented):
Nice work!

src/elements/core/controllers/escapable-overlay-controller.ts
```diff
@@ -0,0 +1,39 @@
+import { isServer, type ReactiveController, type ReactiveControllerHost } from 'lit';
+
+import type { SbbOpenCloseBaseElement } from '../base-elements/open-close-base-element.js';
+
+export const overlayStack = new Array<SbbOpenCloseBaseElement>();
```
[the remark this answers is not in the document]
jeripeierSBB: maybe just passing the var into the constructor? and from the test you can pass in your own stack. see inertcontroller as possible reference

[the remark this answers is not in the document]
DavideMininni-Fincons: it should be fixed by https://github.com/sbb-design-systems/lyne-components/pull/3385/commits/ef95e0d7b5bf2b586361add84c5561b9272f7bb8

PR #3386, opened 2025-02-05
fix(sbb-datepicker-toggle): fix popover style

DavideMininni-Fincons:
Reset popover style in datepicker-toggle
image


kyubisation (approved):
LGTM 👍

PR #3389, opened 2025-02-06
chore: remove puppeteer dependency

TomMenga:
We kept the puppeteer configuration just for the possibility of debugging tests.

Since we found a way to do it with Playwright, it does not add any value anymore. 
Plus, tests do not run consistently on it (I was probably the only one using it)


kyubisation (approved):
LGTM 👍 
Thanks! 😄

PR #3394, opened 2025-02-12
fix(buttons): avoid moving surrounding context on hover

jeripeierSBB:


TomMenga (dismissed):
LGTM 👍

src/elements/button/common/button-common.scss
```diff
@@ -229,7 +226,6 @@ $active: ':active, [data-active]';
 .sbb-action-base {
   // Reset for link variant
   text-decoration: none;
-  display: flex;
   white-space: nowrap;
 }
```
TomMenga, line 230: I think we can merge this set of rules with the class declaration in line 177

TomMenga (approved):
LGTM 👍 
Just a little improvement

src/elements/button/common/button-common.scss
jeripeierSBB: Originally it was planned to extract the button styles, but the reset for the link we wanted to keep here. So this is the reason for two blocks. But with this PR, we loose again the possibility to extract the styles into one single class. So I can merge them together again.

PR #3399, opened 2025-02-14
docs: added warning about necessary import in type-script file

flgubler-ergon:
It is - in my opinion - not obvious that I need to make a JavaScript import in the TypeScript file of my component to use a lyne-component in its template.

(also, that was not necessary in the "older" lyne-components dependency)

### DIFF
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -76,6 +76,8 @@ Select your technology to get started.
 
 4. In order to use web components with Angular, you have to import `CUSTOM_ELEMENTS_SCHEMA` from the `@angular/core` package.
 
+5. In each component, import the lyne components, which you want to use, in the typescript file: e.g. `import '@sbb-esta/lyne-elements/button.js';`
+
 ### Example app
 
 ```ts


kyubisation:
Hello @flgubler-ergon 
Thank you for the PR. Could you please adapt the PR name according to the guidelines in CONTRIBUTING.md?
Also, this is already displayed in the example just below. But if you feel that this provides more visibility for this topic, then we'll go ahead with this.

flgubler-ergon:
thanks @kyubisation 
I think the extra comment helps because I (and probably also other developers) often skip reading the import-statements in code because 90% of the time the IDE handles them correctly on its own. 
I first forgot the add the import statement and there is no error-message whatsoever, the styling is just not applied but neither the IDE nor the browser gave any hint what the problem was.

jeripeierSBB:
@flgubler-ergon With the upcoming `@sbb-esta/lyne-angular` package (prerelease available, https://www.npmjs.com/package/@sbb-esta/lyne-angular), this problem should have gone, as either auto import should work and also angular compiler would detect missing imports.

flgubler-ergon:
Thanks everyone - and thanks for the nice work with the component library :-)

PR #3404, opened 2025-02-14
feat: implement native validation for form elements

kyubisation:
This PR adds validation via ElementInternals to form controls.


src/elements/core/mixins/form-associated-validation-mixin.ts
```diff
@@ -0,0 +1,133 @@
...
+      const customFlags = Object.keys(ValidityState.prototype).filter(
+        (f) => !(f in validityKeys) && f !== 'valid',
+      );
+      for (const flag of customFlags) {
+        const value = flag in flags ? flags[flag as keyof ValidityStateFlags] : false;
+        Object.defineProperty(this.internals.validity, flag, { value, configurable: true });
+        if (value) {
+          // If any custom errors are provided, we need to set customError to true,
+          // as this is the only custom property browsers accept.
+          flags.customError = true;
```
jeripeierSBB, line 122: Something to discuss so I probably better understand it: if we delegate extending the prototype to the consumer, shouldn't this part not also be in responsibility of the consumer? So we don't do any custom stuff ourself?

jeripeierSBB (commented):
Interesting work, thanks for addressing this!

src/elements/core/mixins/form-associated-checkbox-mixin.ts
```diff
@@ -145,6 +152,22 @@ export const SbbFormAssociatedCheckboxMixin = <T extends Constructor<LitElement>
       }
     }
 
+    public override requestUpdate(
+      name?: PropertyKey,
+      oldValue?: unknown,
+      options?: PropertyDeclaration,
+    ): void {
+      super.requestUpdate(name, oldValue, options);
+      if (this.hasUpdated && (name === 'checked' || name === 'required' || !name)) {
```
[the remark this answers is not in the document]
kyubisation: `!name` should solve this.

docs/GETTING_STARTED.md
```diff
@@ -384,3 +384,53 @@ p {
...
+
+#### `readonly validity: ValidityState`
+
+https://developer.mozilla.org/en-US/docs/Web/API/ValidityState
+
+The `validity` property return the current validity state of an element.
+
+#### `readonly validationMessage: string`
+
+The `validationMessage` propert returns the currently applicable validation
```
jeripeierSBB, line 414:
```suggestion
The `validationMessage` property returns the currently applicable validation
```

docs/GETTING_STARTED.md
```diff
@@ -384,3 +384,53 @@ p {
...
+
+The `checkValidity()` method returns true if the element has no validity
+problems; false otherwise. Fires an invalid event at the element in the
+latter case.
+
+#### `setCustomValidity(message: string): void`
+
+The `setCustomValidity(message: string)` method sets the custom validity
+message for the element. Use the empty string to indicate that the element
+does not have a custom validity error.
```
jeripeierSBB, line 436: Nitpicky, but for me it should be somehow clear that a consumer can set a message. Reading the existing phrase it seems not very clear for me.
```suggestion
By using `setCustomValidity(message: string)` it's possible to define a custom validation
message which is set on the element. Use the empty string to indicate that the element
does not have a custom validity error.
With this method, the validation state of the element can be controlled from outside.
```

jeripeierSBB (approved):
LGTM! thank you for addressing everything.

PR #3409, opened 2025-02-18
feat(sbb-toggle): implement form association

### LINKED ISSUE #3364: sbb toggle

TomMenga:
This PR Closes #3364

**Note for reviewers**
I added a few comments to clarify the decisions I took during development


src/elements/toggle/toggle-option/toggle-option.ts
```diff
@@ -79,19 +70,10 @@ class SbbToggleOptionElement extends SbbIconNameMixin(LitElement) {
...
   private _handleDisabledChange(): void {
-    // Enforce disabled state from parent.
-    if (!this._toggle) {
-      // Ignore illegal state. Our expectation  is that a sbb-toggle-option
-      // always has a parent sbb-toggle.
-    } else if (this._toggle.disabled && !this.disabled) {
-      this.disabled = true;
-    } else if (!this._toggle.disabled && this.disabled) {
-      this.disabled = false;
-    }
```
TomMenga, line 94: I removed this logic to simplify the dependency between the `toggle` and the `toggle-option`.
I decided to do it for two reasons:
1 - The `sbb-toggle` is also disabled when `formDisabled`. Which is a protected property that is not accessible from the toggle option.
2 - It was a protection against a misuse case (which we generally do not cover)

src/elements/toggle/toggle-option/toggle-option.ts
```diff
@@ -21,28 +22,18 @@ export
...
+  @forceType()
   @property()
-  public set value(value: string) {
-    this._value = `${value}`;
-  }
-  public get value(): string {
-    return this._value;
-  }
-  private _value: string = '';
+  public accessor value: string = '';
```
jeripeierSBB, line 36: nice improvement :)

src/elements/toggle/toggle-option/toggle-option.ts
```diff
@@ -21,28 +22,18 @@ export
   role: 'radio',
 })
 @slotState()
-class SbbToggleOptionElement extends SbbIconNameMixin(LitElement) {
+class SbbToggleOptionElement extends SbbDisabledMixin(SbbIconNameMixin(LitElement)) {
```
jeripeierSBB, line 25: nice improvement

src/elements/toggle/toggle/toggle.ts
```diff
@@ -97,8 +97,55 @@ class SbbToggleElement extends LitElement {
     this.addEventListener?.('keydown', (e) => this._handleKeyDown(e));
   }
 
-  /** @internal */
-  public updatePillPosition(resizing: boolean): void {
```
jeripeierSBB, line 101: to avoid a breaking change, we should keep the method and mark it deprecated

jeripeierSBB (changes requested):
Nice and clean work, should hopefully be okay :D one change about a breaking change needed.

src/elements/toggle/toggle/toggle.ts
TomMenga: Even though it was marked as internal?

jeripeierSBB: I think yes, @kyubisation your opinion?

kyubisation: Yes, we should be very conservative with API available in TypeScript types.
It's annoying, but we want to be as safe as possible.

jeripeierSBB (approved):
LGTM, thank you!

PR #3423, opened 2025-02-25
fix(panels): reflect colors

### LINKED ISSUE #3418: [Bug]: Prop `color` not working on Panels since React 19
### Preflight Checklist

### Bug type

Functionality

Components

2.2.0

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

Since we updated our app to Next.js 15 and React 19 the `color` setting of panels did work work anymore by simply setting `color="milk"` on respective Lyne components.

Rendering the following JSX displayed a panel with milk-colored background before the update:

```jsx
<SbbSelectionExpansionPanel color="milk">
  ...
</SbbSelectionExpansionPanel>
```

Since the update the component is rendered like this:

Image

This seems to affect all kinds of panels, including `SbbSelectionExpansionPanel`, `SbbCheckboxPanel` and `SbbRadioButtonPanel`.

We are currently working around this issue by setting the `color` attribute using a Ref callback. This works as expected.

```tsx
function MyComponent(): React.JSX.Element {
  // Workaround: Since React 19 the prop "color" seems to be set as DOM property
  // and not as attribute which is not respected by the expansion panel.
  const expansionPanelRefCallback = useCallback((element: SbbSelectionExpansionPanelElement | null) => {
    if (element) element.setAttribute('color', 'milk');
  }, []);

  return (
    <SbbSelectionExpansionPanel ref={expansionPanelRefCallback} color="milk">
      ...
    </SbbSelectionExpansionPanel>
  );
}
```

One important thing to note is that React since version 19 handles Web Components differently. Props are not always set as attributes as before. Now it sets props as DOM properties if applicable.

> Client Side Rendering: props that match a property on the Custom Element instance will be assigned as properties, otherwise they will be assigned as attributes.

See https://react.dev/blog/2024/12/05/react-19#support-for-custom-elements

1. Render React version of a panel, e.g. `SbbSelectionExpansionPanel` with prop `color="milk"`.
2. Notice that color is not set as expected.

Get in touch with Team Volantis for a demonstration in our app.

### Relevant log output

```shell

```

jeripeierSBB:
Closes #3418

### DIFF
--- a/src/elements/accordion/__snapshots__/accordion.snapshot.spec.snap.js
+++ b/src/elements/accordion/__snapshots__/accordion.snapshot.spec.snap.js
@@ -4,6 +4,7 @@ export const snapshots = {};
 snapshots["sbb-accordion renders DOM"] = 
 `<sbb-accordion size="l">
   <sbb-expansion-panel
+    color="white"
     data-accordion=""
     data-accordion-first=""
     size="l"
@@ -33,6 +34,7 @@ snapshots["sbb-accordion renders DOM"] =
     </sbb-expansion-panel-content>
   </sbb-expansion-panel>
   <sbb-expansion-panel
+    color="white"
     data-accordion=""
     data-accordion-last=""
     size="l"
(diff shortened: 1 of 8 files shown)


kyubisation (approved):
LGTM 👍

PR #3427, opened 2025-03-03
fix(sbb-checkbox-group, sbb-radio-button-group): fix panel detection and styling

### LINKED ISSUE #3426: bug: radio & check panel within a group don't expand full width (Angular)
### Feature Description

Noticed in stories in Angular repo:
when several `sbb-checkbox-panel` are slotted in a `sbb-checkbox-group`, they don't occupy the page full width;
it seems that the `data-has-panel` attribute, which sets a 100% width, is not set on the group. 

Same for sbb-radio-button-group/panel.

jeripeierSBB:
Closes #3426

### DIFF
--- a/src/elements/checkbox/checkbox-group/checkbox-group.ts
+++ b/src/elements/checkbox/checkbox-group/checkbox-group.ts
@@ -1,5 +1,5 @@
 import type { CSSResultGroup, PropertyValues, TemplateResult } from 'lit';
-import { LitElement, html } from 'lit';
+import { html, LitElement } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 
 import { getNextElementIndex, interactivityChecker, isArrowKeyPressed } from '../../core/a11y.js';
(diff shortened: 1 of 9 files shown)


src/elements/checkbox/checkbox-group/checkbox-group.ts
```diff
@@ -60,10 +60,7 @@ class SbbCheckboxGroupElement extends SbbDisabledMixin(LitElement) {
 
   public override connectedCallback(): void {
     super.connectedCallback();
-    this.toggleAttribute(
-      'data-has-panel',
-      !!this.querySelector?.('sbb-selection-expansion-panel, sbb-checkbox-panel'),
-    );
+    this.toggleAttribute('data-has-panel', !!this.querySelector?.('sbb-checkbox-panel'));
```
DavideMininni-Fincons, line 63: Is the `selection-expansion-panel` not needed anymore? same for radio-group

DavideMininni-Fincons (approved):
LGTM!

src/elements/checkbox/checkbox-group/checkbox-group.ts
jeripeierSBB: I think not, as you never use a selection expansion panel without a containing checkbox or radio panel. with queryselector you will also find the underlaying checkbox panel or radio panel.

jeripeierSBB: do you agree?

PR #3430, opened 2025-03-05
fix(sbb-autocomplete): fix floating label flickering

### LINKED ISSUE #3410: bug(autocomplete, autocomplete-grid): floating label flickers on selection
Reproduction:
1. https://lyne-components-pr3312.app.sbb.ch/?path=/story/elements-sbb-autocomplete-grid-sbb-autocomplete-grid--floating-label
2. Select an option
3. See that label gets bigger and smaller in a short amount of time

Expected behavior:
Should not flicker.

TomMenga:
This PR Closes #3410

### DIFF
--- a/src/elements/autocomplete/autocomplete-base-element.ts
+++ b/src/elements/autocomplete/autocomplete-base-element.ts
@@ -107,6 +107,7 @@ abstract class SbbAutocompleteBaseElement extends SbbNegativeMixin(
 
     this.showPopover?.();
     this.state = 'opening';
+    this._triggerElement?.toggleAttribute('data-expanded', true);
     this._setOverlayPosition();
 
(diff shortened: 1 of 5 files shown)


jeripeierSBB (approved):
Thanks! wondering if it would be possible to just use data-state, but it's ok like you did as well :) Did you think of a test to cover the fix? at least that the expanded stuff is set (the flickering itself would be difficult :) )

TomMenga:
> Thanks! wondering if it would be possible to just use data-state, but it's ok like you did as well :) Did you think of a test to cover the fix? at least that the expanded stuff is set (the flickering itself would be difficult :) )

I could use the `data-state` but for the autocomplete is a bit messier since the `state` is not on the `input`. I thought that we could simplify the dependencies from other components and define an easy way to handle this case for future form components with overlays.

PR #3442, opened 2025-03-11
feat(sbb-stepper): fix wrong stepper height

DavideMininni-Fincons:
The bug arise from the `display:block` of non-selected step.
However, the animation doesn't look very good.


src/elements/dialog/dialog/dialog.stories.ts
```diff
@@ -406,7 +406,7 @@ const StepperTemplate = ({
   <sbb-dialog id="stepper" ${sbbSpread(args)}>
     ${dialogTitle(level, backButton, hideOnScroll, accessibilityCloseLabel, accessibilityBackLabel)}
     <sbb-dialog-content>
-      <sbb-stepper linear orientation="horizontal" size="m">
+      <sbb-stepper orientation="horizontal" size="m">
```
jeripeierSBB, line 409: Should we re-add?

src/elements/stepper/step/step.scss
```diff
@@ -59,13 +59,21 @@
...
   color: var(--sbb-step-color);
   transition: {
-    property: opacity, visibility;
+    property: display, opacity;
     duration: var(--sbb-step-animation-duration);
     delay: var(--sbb-step-animation-delay);
     timing-function: var(--sbb-animation-easing);
   }
+
+  transition-behavior: allow-discrete;
```
jeripeierSBB, line 72: maybe add a note about known compatibility issues

src/elements/stepper/step/step.scss
```diff
@@ -59,13 +59,21 @@
...
+}
+
+@starting-style {
```
jeripeierSBB, line 75: Can it be placed directly in .sbb-step?

src/elements/dialog/dialog/dialog.stories.ts
DavideMininni-Fincons: I added parameters in the story for both linear and orientation, in order to test also the animation in vertical mode

TomMenga (approved):
LGTM 👍 
Pls have a look at the visual specs which are failing

PR #3476, opened 2025-03-19
build: update eslint plugin configuration

kyubisation:


jeripeierSBB (approved):
LGTM, thanks! Are we sure that the rules still run? ;-)

kyubisation:
No, because they were annoying anyway 😆 
Yes, everything still works as expected 👍

PR #3477, opened 2025-03-19
build: update Node.js version

kyubisation:
We were previously stuck with version 22.2.0 of Node.js, as that had Node.js workers inherit hooks from the calling context.
This had been removed with 22.3.0. However, we extended the functionality of the Lit SSR Testing implementation and can now load the hooks in the worker directly (https://github.com/lit/lit/pull/4730).


jeripeierSBB (approved):
LGTM

PR #3481, opened 2025-03-21
build: add size check for GitHub CI

kyubisation:
This PR adds functionality for performing a size check for pull requests. As this requires the relevant code to be on main, this PR cannot yet be used to test the functionality.


scripts/calculate-stats.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+}
+
+const dist = fileURLToPath(import.meta.resolve('../dist/'));
+const stats = {
+  size: 0,
+  cssSize: 0,
+  files: {} as Record<string, { size: number; cssSize?: number }>,
+};
+const themePath = 'elements/standard-theme.css';
```
jeripeierSBB, line 24: What about looking for files with "-theme.css" and also using a the named key (`standardCss`) below? It would be more future proof

scripts/calculate-stats.ts
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+
+    if (cssSize) {
+      stats.cssSize += cssSize;
+      stats.files[key].cssSize = cssSize;
+    }
+  }
+}
+
```
jeripeierSBB, line 71: I think we should also save the stats for brotlified files. this would give a more accurate feeling. about how good it is. e.g. the css has a lot of stuff which can be nicely compressed.

jeripeierSBB (commented):
Very nice that you bring this idea to life :-)

TomMenga (approved):
Very nice! :D

scripts/calculate-stats.ts
kyubisation: I added this in general, but I cannot do this for the css in js case, as there is no reliable way to figure out what is js and what is css in the compressed file

jeripeierSBB (approved):
Very cool! Let's merge and try this :)

PR #3482, opened 2025-03-21
test: increase timeout for flaky ssr test

kyubisation:


src/elements-experimental/timetable-row/timetable-row.ssr.spec.ts
```diff
@@ -13,7 +13,10 @@ describe(`sbb-timetable-row ssr`, () => {
     });
   });
 
-  it('renders', () => {
```
jeripeierSBB, line 16: I saw also another test with the same error: the journey summary ssr. do we need in general a slightly higher timeout for ssr tests?

PR #3485, opened 2025-03-24
test(sbb-icon-sidebar): improve test stability

TomMenga:


src/elements/sidebar/icon-sidebar/icon-sidebar.visual.spec.ts
```diff
@@ -8,6 +8,7 @@ import '../../link/block-link.js';
 import '../../link-list.js';
 import '../../logo.js';
 import '../../sidebar.js';
+import { waitForLitRender } from '@sbb-esta/lyne-elements/core/testing/wait-for-render';
```
jeripeierSBB, line 11: import should be relative and containing ".js"

src/elements/sidebar/icon-sidebar/icon-sidebar.visual.spec.ts
```diff
@@ -121,9 +122,10 @@ describe('sbb-icon-sidebar', () => {
           { minHeight: '400px' },
         );
 
-        setup.withPostSetupAction(() => {
+        setup.withPostSetupAction(async () => {
           const sidebar = setup.snapshotElement.querySelector('sbb-sidebar')!;
           sidebar.open();
+          await waitForLitRender(sidebar);
```
jeripeierSBB, line 128: basically in the test setup everything runs synchronously when opening. Following the code path I don't see anything which could improve the sidebar rendering output when awaiting the lit render.
what exactly did flicker?

jeripeierSBB (commented):
Thanks for addressing this issue :)

src/elements/sidebar/icon-sidebar/icon-sidebar.visual.spec.ts
TomMenga: In one of Lukas PR i saw this [visual spec](https://lyne-components-visual-regression-diff-pr3481.app.sbb.ch/compare/sbb-icon-sidebar/nested) failing which looked like a time-sensitive case.

I didn't deep down the code but, in any case, if the fix were to work, I expected to see the same test failing here

PR #3487, opened 2025-03-24
fix(sbb-select): update displayed value on option slot change

jeripeierSBB:


TomMenga (approved):
LGTM 👍 
Should we consider removing the `state-change` event as a breaking change?

jeripeierSBB:
> LGTM 👍 Should we consider removing the `state-change` event as a breaking change?

Good input, but discussed this one during development with @kyubisation. We decided to continue with the change.

PR #3505, opened 2025-03-27
fix: synchronize id references

jeripeierSBB:
We have several components that rely on an id connection. Before this change, when the id was not found when setting the id, the referenced element was never found. With this change we introduce an id reference observer which allows finding id's in the document (rootNode) whenever they appear or vanish. This imitates the browser behavior for id connections.


src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -108,7 +125,13 @@ abstract class SbbAutocompleteBaseElement extends SbbNegativeMixin(
     this.showPopover?.();
     this.state = 'opening';
     this._triggerElement?.toggleAttribute('data-expanded', true);
-    this._setOverlayPosition();
+    const originElement = this.originElement;
+    if (!originElement) {
+      throw new Error(
+        'Cannot find the origin element. Please specify a valid element or read the "origin" prop documentation',
```
DavideMininni-Fincons, line 131:
```suggestion
        'Cannot find the origin element. Please specify a valid element or check the usage of the "origin" property from the documentation',
```

DavideMininni-Fincons (approved):
LGTM

note: there are some TODOS left: will they be fixed here or in a following PR?

kyubisation (approved):
LGTM 👍

PR #3506, opened 2025-03-27
refactor(sbb-date-input): feature detect plaintext-only support

kyubisation:
Currently we do not use `contenteditable="plaintext-only"` on Firefox, as it was not supported. Now that support has landed, we are switching to feature detection.


jeripeierSBB (approved):
LGTM

PR #3508, opened 2025-03-31
fix(sbb-step): fix stepper height when content changes

DavideMininni-Fincons:
Changes in this PR:
- I added a ResizeObserver on the `sbb-step`; it emits a `resizeChange` event which is caught by the the `sbb-stepper` to correctly set the stepper's height
- the mentioned ResizeObserver triggers its logic with a `setTimeout` set to 0, to avoid to have the "ResizeObserver loop completed with undelivered notifications" error in test; this behavior has been copied from the `sbb-dialog`


src/elements/stepper/step/step.ts
```diff
@@ -41,9 +49,28 @@ class SbbStepElement extends LitElement {
...
 
+  // We use a timeout as a workaround to the "ResizeObserver loop completed with undelivered notifications" error.
+  // For more details:
+  // - https://github.com/WICG/resize-observer/issues/38#issuecomment-422126006
+  // - https://github.com/juggle/resize-observer/issues/103#issuecomment-1711148285
+  private _stepResizeObserver = new ResizeController(this, {
+    target: null,
+    skipInitial: true,
+    callback: () => setTimeout(() => this._onStepElementResize()),
+  });
```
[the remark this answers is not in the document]
DavideMininni-Fincons: It's exactly what i did.

src/elements/stepper/step/step.ts
```diff
@@ -134,6 +168,12 @@ class SbbStepElement extends LitElement {
     this._label = this._getStepLabel();
   }
 
+  protected override async firstUpdated(changedProperties: PropertyValues<this>): Promise<void> {
```
jeripeierSBB, line 171: we should not alter the modifiers of a inherited method. However, I see a couple of other occurrences :D One simple possibility is to use ".then"

src/elements/stepper/stepper/stepper.ts
```diff
@@ -307,6 +313,14 @@ class SbbStepperElement extends SbbHydrationMixin(LitElement) {
     this._loaded = true;
     this.selectedIndex = this.linear ? 0 : Number(this.getAttribute('selected-index')) || 0;
     this._observer.observe(this);
+    this._resizeAbortController = new AbortController();
+    this.steps.forEach(
```
jeripeierSBB, line 317: Am I right, that if a new step is slotted that there will be no listener on it?
I propose to refactor the logic as following:
1. make sure the event bubbles
2. listen to the event on stepper level
3. get the step from the event target.

With that, the above problem should be solved.

jeripeierSBB (changes requested):

src/elements/stepper/stepper/stepper.ts
jeripeierSBB: You could also remove the abort controller and just in the controller add the listener

PR #3514, opened 2025-04-01
fix: avoid throwing on parsing invalid date

kyubisation:
Currently we use `createDate` inside `parse`, which throws on invalid date. However, `parse` should never throw, just return `null` for an invalid date, which this PR addresses.


jeripeierSBB (approved):
LGTM :)

PR #3524, opened 2025-04-02
style(sbb-table): add support to header filters

TomMenga:
Propaedeutic to the angular table example

Implementation of https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?node-id=11340-100695&m=dev

### DIFF
--- a/src/elements/core/styles/mixins/table.scss
+++ b/src/elements/core/styles/mixins/table.scss
@@ -28,8 +28,12 @@
   }
 
   thead {
-    & > tr > th {
-      @include table-header-cell;
+    & > tr {
+      @include table-header-row;
+
+      & > th {
+        @include table-header-cell;
+      }
     }
   }
 
@@ -45,6 +49,13 @@
 
   caption {
     @include table-caption;
+  }
+}
+
+@mixin table-header-row {
+  // If there are more header rows, only the last one has the bottom border
+  &:not(:last-of-type) > th {
+    border-bottom: none;
   }
 }
 
@@ -91,6 +102,10 @@
   color: var(--sbb-table-caption-color, var(--sbb-color-granite));
   margin-block-start: var(--sbb-table-caption-margin-block-start, var(--sbb-spacing-fixed-4x));
   text-align: left;
+}
+
+@mixin table-filter {
+  padding-block-start: 0 !important;
 }
 
 @mixin table--striped {
--- a/src/elements/core/styles/mixins/table.scss
+++ b/src/elements/core/styles/mixins/table.scss
@@ -18,9 +18,6 @@
     --sbb-table-caption-color: var(--sbb-color-granite);
     --sbb-table-caption-margin-block-start: var(--sbb-spacing-fixed-4x);
 
-    background-color: var(--sbb-table-background-color);
-    border: var(--sbb-table-border);
-    border-radius: var(--sbb-table-border-radius);
     border-spacing: 0;
     caption-side: bottom;
     color: var(--sbb-table-color);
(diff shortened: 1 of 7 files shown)


src/elements/core/styles/table.scss
```diff
@@ -60,6 +64,11 @@ sbb-table-wrapper[negative] .sbb-table,
   @include table.table-caption;
 }
 
+.sbb-table-filter {
+  @include table.table-filter;
+}
+
+// TODO In future, move to the 'sbb-lean' theme
```
jeripeierSBB, line 71:
```suggestion
// TODO: In future, move to the 'sbb-lean' theme
```

src/elements/core/styles/mixins/table.scss
```diff
@@ -48,6 +52,13 @@
   }
 }
 
+@mixin table-header-row {
+  // If there is more than one header row, only the last one has the bottom border
+  &:not(:last-of-type) > th {
+    border-bottom: none;
```
jeripeierSBB, line 58: possible to use logical?
```suggestion
    border-block-end: none;
```

src/elements/core/styles/mixins/table.scss
```diff
@@ -48,6 +52,13 @@
   }
 }
 
+@mixin table-header-row {
+  // If there is more than one header row, only the last one has the bottom border
+  &:not(:last-of-type) > th {
```
jeripeierSBB, line 57: would it be easier to invert the definition like only a bottom border for the last of type? just a question, your solution is fine.

src/elements/table/table-wrapper/table-wrapper.scss
```diff
@@ -6,14 +6,15 @@
...
-  display: block;
+  display: contents;
 }
 
 ::slotted(.sbb-table) {
   width: 100%;
 }
 
 .sbb-table-wrapper {
+  height: 100%;
```
jeripeierSBB, line 17: I'm a little bit worried about the height: 100% because this would possibly change the height unwanted? Because with the combination of display: contents, everything is delegated into here. how could you control the height from outside?

TomMenga: Good point
To be conservative, I'll leave the host as `display: block`. This way, the internal height definition does not have an effect unless a height is set on the host.
The height is necessary for the .sbb-table-wrapper, otherwise, you cannot achieve overflow

src/elements/core/styles/mixins/table.scss
TomMenga: I chose this because there is a more general rule that always adds the border. I need to subtract it in the exception case

src/elements/core/styles/mixins/table.scss
TomMenga: It doesn't really add any value since the borders are independent of the writing mode. 
I feel like it only adds the burden to 'test' the table composition in more cases. What do you think?

jeripeierSBB: It's more our base rule: always use logical properties and only if not working use not logical ones and comment it

src/elements/core/styles/mixins/table.scss
jeripeierSBB: Just from reading the comment ('If there is more than one header row, only the last one has the bottom border') I'd expect a rule like setting the border for the last one. but as said, I can live with your solution :D

src/elements/core/styles/mixins/table.scss
TomMenga: I'm still not thoroughly convinced but I don't feel like it's worth enough to discuss it 😅 
I'll change other occurrences too

jeripeierSBB (approved):
LGTM, thanks! Do we need a @mcilurzo review?

TomMenga:
> LGTM, thanks! Do we need a @mcilurzo review?

Not sure, the impact is relatively low in the main library. It was more of a refactor.
We will surely need Manuel review for the angular examples

jeripeierSBB (approved):
LGTM again ;-)

PR #3527, opened 2025-04-02
build: refactor entry points logic

kyubisation:
This PR changes our previous entry point detection logic to explicitly define entry points via js doc.
e.g. 
```ts
/**
 * @entrypoint
 */
```
Additionally our build no longer moves the component logic down into the re-export but keeps it at the (renamed) location of the original source file.
However, our package.json exports field prevents direct access to these component files and forces consumers to still use the re-export files, as we intend.

**Important note:** All files that should not be included in the build need to be suffixed with `.spec.ts`, `.stories.ts`, `.private.ts` or be in a directory named `private`. This PR adapts the file name of files that should not be included with the `.private.ts` suffix.


src/elements/vite.config.ts
```diff
@@ -96,10 +87,15 @@ export default defineConfig((config) =>
       sourcemap: isProdBuild(config) ? false : 'inline',
       rollupOptions: {
         external: (source: string, importer: string | undefined) => {
+          if (!source.startsWith('.') && importer?.startsWith(fileURLToPath(packageRoot))) {
+            //console.log(`${source} <= ${importer}`);
```
jeripeierSBB, line 91: Should we keep the console.log comment?

DavideMininni-Fincons:
Generally looks good, i've got only a question: could the change on the custom-elements-manifest-config have an impact on the generation of files in the Angular project? or the manifest stays unchanged?

kyubisation:
@DavideMininni-Fincons That is possible, as the structure of the custom-elements.json manifest changes slightly. Once I merge this PR, I will have a look at the lyne-angular logic and adapt it, where necessary.

PR #3528, opened 2025-04-03
fix(sbb-clock): handle 'visibilityChange' when paused

### LINKED ISSUE #3520: bug(sbb-clock): time changes switching to another page
Steps to reproduce:
- open https://lyne-storybook.app.sbb.ch/?path=/story/elements-sbb-clock--paused
- clock is correctly set at 09:10:30
- switch to another tab or open a new one
- get back to the clock
- time get set at 12:10:00

TomMenga:
This PR Closes #3520


jeripeierSBB (approved):
Nice fix. thank you!

PR #3537, opened 2025-04-04
build: add lint rule for only using relative imports inside package

kyubisation:
This PR adds an eslint rule to only allow relative imports inside a package. It also implements a fixer to automatically fix failing cases.


TomMenga (approved):
LGTM 👍

PR #3544, opened 2025-04-07
refactor: remove async overrides to avoid changing signature

jeripeierSBB:
BREAKING CHANGE: removed async modifier of all firstUpdated() overrides.


kyubisation (approved):
LGTM 👍

PR #3552, opened 2025-04-08
fix: use correct id resolution for form

kyubisation:
Currently we use `this.ownerDocument`, which is basically the same as `document`. We however want to resolve to the given id in the current document fragment, for which we should use `this.getRootNode()`.


jeripeierSBB (approved):
LGTM

PR #3558, opened 2025-04-09
build: remove font preloading

kyubisation:
We tested font preloading for more consistent visual regression tests. However, it actually caused decreased reliability, so we are removing it.


jeripeierSBB (approved):
LGTM, thanks

PR #3583, opened 2025-04-15
test(sbb-table-wrapper): fix visual spec tests

TomMenga:


jeripeierSBB (approved):
LGTM, thanks for the quick fix

PR #3624, opened 2025-04-22
feat(sbb-table-wrapper): introduce focusable property

### LINKED ISSUE #3623: Move table wrapper focusable property into lyne-elements and set role in lyne-elements and also fix focus outline

jeripeierSBB:
Closes #3623

### DIFF
--- a/src/elements/table/table-wrapper/table-wrapper.component.ts
+++ b/src/elements/table/table-wrapper/table-wrapper.component.ts
@@ -2,18 +2,20 @@ import { ResizeController } from '@lit-labs/observers/resize-controller.js';
 import {
   type CSSResultGroup,
   html,
+  isServer,
   LitElement,
   type PropertyValues,
   type TemplateResult,
 } from 'lit';
-import { customElement } from 'lit/decorators.js';
+import { customElement, property } from 'lit/decorators.js';
 
+import { forceType } from '../../core/decorators.js';
 import { SbbNegativeMixin } from '../../core/mixins.js';
 
 import style from './table-wrapper.scss?lit&inline';
 
 /**
- * Wraps a table to enhance its functionality
+ * Wraps a table to enhance its functionality.
  *
  * @slot - Use the unnamed slot to add the table.
  */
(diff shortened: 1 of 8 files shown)


src/elements/table/table-wrapper/table-wrapper.scss
```diff
@@ -4,26 +4,18 @@
...
-  overflow: auto;
-  @include sbb.scrollbar($size: thick, $track-visible: true);
-
-  :host([negative]) & {
-    @include sbb.scrollbar-variables--color-negative;
-    @include sbb.scrollbar-variables--color-negative-track-visible;
-  }
+:host(:focus-visible) {
+  @include sbb.focus-outline;
+}
```
MarioCastigliano, line 17: In negative mode focus outline is still black, making it not visible

src/elements/table/table-wrapper/table-wrapper.component.ts
```diff
@@ -22,21 +24,68 @@ export
...
-  private _resizeObserver = new ResizeController(this, {
-    target: null,
-    skipInitial: true,
-    callback: () => this._checkHorizontalScrollbarOffset(),
-  });
-  private _tableWrapper!: HTMLElement;
+  /** Whether the table wrapper is focusable. */
+  @forceType()
+  @property({ reflect: true, type: Boolean })
+  public accessor focusable: boolean = false;
```
MarioCastigliano, line 30: It might be that I'm not understanding the scope completely, but at the moment the table is always focusable, whether `focusable` is set to `true` or not. Should it naturally prevent focus (maybe with a tabindex="-1") when it's set to false?

MarioCastigliano (commented):
Just a couple of questions

src/elements/table/table-wrapper/table-wrapper.scss
jeripeierSBB: That is on purpose, as the outline is around the negative table. the negative table can be used inside a positive context. the outline is depending on the background of the surrounding area and therefore is responsibility of consumer or outer component.

src/elements/table/table-wrapper/table-wrapper.visual.spec.ts
```diff
@@ -74,5 +75,25 @@ describe(`sbb-table-wrapper`, () => {
...
+              </tr>
+              <tr>
+                <td>Content</td>
+                <td>Content</td>
+              </tr>
+            </table>
+          </sbb-table-wrapper>`,
+        );
+      }),
+    );
```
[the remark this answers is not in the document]
jeripeierSBB: see other comment

src/elements/table/table-wrapper/table-wrapper.component.ts
jeripeierSBB: This is due to every scroll area (overflow: auto) is per definition focusable as long as there is no focusable element within. But I think this behavior should not be overridden by us. It could break accessibility. So if you expand your window, and there is no scroll area anymore, it shouldn't be focusable (when property turned off)?

MarioCastigliano: Makes sense, thanks

MarioCastigliano (approved):
👍

PR #3645, opened 2025-04-30
fix(sbb-date-input): handle null like the native text input

kyubisation:
Currently assigning null to `SbbDateInput.value` will result in `value` being `'null'`. This is in contrast to the native text input, where, if you assign `null`, it will result in `value` being `''`.
As our goal is to align as closely as possible to the native behavior, we also adapt this.


jeripeierSBB (approved):
LGTM, nice! :)

PR #3646, opened 2025-04-30
fix(sbb-chip-group): fix chip position when initialized with value

TomMenga:


jeripeierSBB (approved):
LGTM, thanks for the fix!

PR #3652, opened 2025-05-02
refactor: adapt id reference properties to only accept element references

kyubisation:
This change removes the string type from id reference attributes and adds the `idReference` decorator. This will dynamically resolve references in the same document fragment, both when set and when an id reference is established or destroyed due to changes in the DOM.

BREAKING CHANGE: Id reference properties no longer accept string as an argument. However, attribute usage will still work.


src/elements/menu/menu/menu.component.ts
```diff
@@ -312,11 +312,7 @@ class SbbMenuElement extends SbbNamedSlotListMixin<
 
   // Check if the trigger is valid and attach click event listeners.
   private _configureTrigger(): void {
-    const triggerElement =
-      this.trigger instanceof HTMLElement
-        ? this.trigger
-        : this._triggerIdReferenceController.find();
-
+    const triggerElement = this.trigger;
```
jeripeierSBB, line 315: extra const could probably be removed

src/elements/navigation/navigation-section/navigation-section.component.ts
```diff
@@ -218,10 +223,7 @@ class SbbNavigationSectionElement extends SbbUpdateSchedulerMixin(LitElement) {
 
   // Check if the trigger is valid and attach click event listeners.
   private _configureTrigger(): void {
-    const triggerElement = (
-      this.trigger instanceof HTMLElement ? this.trigger : this._triggerIdReferenceController.find()
-    ) as SbbNavigationButtonElement | null;
-
+    const triggerElement = this.trigger as SbbNavigationButtonElement | null;
```
jeripeierSBB, line 226: extra const could probably be removed, but casting then would be needed multiple times

jeripeierSBB (approved):
Nice refactoring!

Do you plan to remove the idReferenceController after datepicker refactoring? I think we should not have two ways for the same logic in the repo.

jeripeierSBB:
TODO: Test ergänzen von https://github.com/sbb-design-systems/lyne-components/pull/3661/files

PR #3661, opened 2025-05-05
fix: avoid throwing on utf8 id reference

jeripeierSBB:


src/elements/core/controllers/id-reference-controller.ts
```diff
@@ -95,7 +95,7 @@ export class SbbIdReferenceController<T extends LitElement> implements ReactiveC
       return null;
     } else {
       const id = this._host[this._idRef];
-      return typeof id == 'string' ? this._rootNode?.querySelector(`#${id}`) : null;
+      return typeof id == 'string' ? this._rootNode?.querySelector(`#${CSS.escape(id)}`) : null;
```
kyubisation, line 98: This can probably be replaced by getElementById, which is available on both Document and ShadowRoot.

kyubisation (approved):

TomMenga (approved):
LGTM 👍

PR #3664, opened 2025-05-06
fix(sbb-toggle): fix animation glitches

jeripeierSBB:
With the previous solution, on sbb.ch it causes problems when resizing the window. Furthermore, with the new solution, the correct pill can be shown earlier, preventing a rendering without the pill and therefore provides a smoother experience.


kyubisation (approved):
LGTM 👍 
Nice!

PR #3666, opened 2025-05-06
docs: fix broken docs

DavideMininni-Fincons:
The docs on the `sbb-datepicker` is broken

image


jeripeierSBB (approved):
Thanks for the fix :) one little fix needed

PR #3672, opened 2025-05-06
refactor: remove autocompleteOptionSelectionChange event

jeripeierSBB:
BREAKING CHANGE: Removed the following event and method
- removed autocompleteOptionSelectionChange event
- removed setSelectedViaUserInteraction method


DavideMininni-Fincons (approved):
Not a big fan of accessing `selectViaUserInteraction` via square brackets, but i understand there's no other way since it is protected.
LGTM

kyubisation (approved):
LGTM 👍

PR #3673, opened 2025-05-07
fix(sbb-autocomplete, sbb-autocomplete-grid): recalculate panel's position on slotchange

DavideMininni-Fincons:


src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -270,6 +270,13 @@ abstract class SbbAutocompleteBaseElement extends SbbNegativeMixin(
 
   private _handleSlotchange(): void {
     this._highlightOptions(this.triggerElement?.value);
+    /**
+     * In Angular, it's possible to filter out option with an opened panel on input change.
```
jeripeierSBB, line 274: As it is possible also without Angular, I suggest to remove this.
```suggestion
     * It's possible to filter out option with an opened panel on input change.
```

src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -270,6 +270,13 @@ abstract class SbbAutocompleteBaseElement extends SbbNegativeMixin(
...
+     * In this case, the panel's position must be recalculated considering the new option's list.
+     */
+    if (this.isOpen) {
+      this._setOverlayPosition();
```
jeripeierSBB, line 278: Please add a test checking updated positions (either spec or visual)

DavideMininni-Fincons: Test added

src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -270,6 +270,13 @@ abstract class SbbAutocompleteBaseElement extends SbbNegativeMixin(
 
   private _handleSlotchange(): void {
     this._highlightOptions(this.triggerElement?.value);
+    /**
+     * It's possible to filter out option with an opened panel on input change.
```
jeripeierSBB, line 274:
```suggestion
     * It's possible to filter out options with an opened panel on input change.
```

jeripeierSBB (approved):
LGTM, thanks for the fix! A little special way of configuring the test container, but should be ok :)

PR #3677, opened 2025-05-07
refactor: remove SbbDateLike

### LINKED ISSUE #3332: Remove SbbDateLike
We want to remove the SbbDateLike interface.
Any property using it should only support the generic date type or Date (and optionally null).
Attributes should be parsed according to use case (either date only or datetime).

kyubisation:
Closes #3332

BREAKING CHANGE: This change removes the `SbbDateLike` type and the possibility to remove UNIX timestamps in date attributes. It is now mandatory to use date objects for date properties and ISO 8601 (e.g. `2024-12-24`) strings for attributes.


src/elements/core/decorators/plain-date.ts
```diff
@@ -0,0 +1,69 @@
...
+    const { kind } = context;
+    if (kind === 'accessor') {
+      return {
+        set(this: C, value) {
+          const dateAdapter = readConfig().datetime?.dateAdapter ?? defaultDateAdapter;
+          value = dateAdapter.getValidDateOrNull(dateAdapter.deserialize(value));
+          target.set.call(
+            this,
+            value
+              ? dateAdapter.createDate(
```
jeripeierSBB, line 49: Isn't value already a valid date or null? why using createDate here? If it i to make a copy, could clone() be used?
Or is it to remove the time information? An explaining comment is maybe a good idea.

src/elements/core/decorators/plain-date.ts
```diff
@@ -0,0 +1,69 @@
...
+                  dateAdapter.getMonth(value),
+                  dateAdapter.getDate(value),
+                )
+              : null,
+          );
+        },
+        get(this: C) {
+          const dateAdapter = readConfig().datetime?.dateAdapter ?? defaultDateAdapter;
+          const value = target.get.call(this);
+          return value != null
```
jeripeierSBB, line 60: Should we maybe have a comment why "!=" (and not !==)

src/elements/date-input/date-input.component.ts
```diff
@@ -54,12 +54,9 @@ class SbbDateInputElement<T = Date> extends SbbFormAssociatedInputMixin(LitEleme
     return super.value ?? '';
   }
 
-  @forceType()
-  @property({ attribute: false, type: DateOnlyType })
+  @property({ attribute: false })
```
jeripeierSBB, line 57: Not sure whether you already told me :D, but why this change?

src/elements/date-input/date-input.component.ts
```diff
@@ -70,23 +67,37 @@ class SbbDateInputElement<T = Date> extends SbbFormAssociatedInputMixin(LitEleme
...
   private _valueAsDate?: T | null;
 
-  @forceType()
-  @property({ converter: dateConverter, reflect: true, type: DateOnlyType })
+  /**
+   * The minimum valid date. Accepts a date object or null.
+   * Accepts an ISO8601 formatted string (e.g. 2024-12-24) as attribute.
+   */
+  @plainDate()
+  @property({ converter: plainDateConverter, reflect: true })
```
jeripeierSBB, line 92: Why do we have to reflect the date?

src/elements/datepicker/datepicker-toggle/datepicker-toggle.component.ts
```diff
@@ -132,7 +132,7 @@ class SbbDatepickerToggleElement<T = Date>
       return;
     }
     this._calendarElement.wide = this.datepicker.wide;
-    this._calendarElement.now = this._nowOrNull();
+    this._calendarElement.now = this._nowOrNull()!;
```
jeripeierSBB, line 135: Looks strange to do this as it can be null. why not using T | null for dates?

jeripeierSBB (commented):
Looks great :-)
 - Could you search the project for "/ 1000" and maybe check if those occurrences should be migrated to not use timestamps anymore? Probably they mostly still work, but it would be kind of clean to not use timestamps. I also checked where we use the storybook date control, but this should be ok as far as I saw.

src/elements/core/decorators/plain-date.ts
kyubisation: Switched to dateAdapter.isValid, which is more correct

src/elements/date-input/date-input.component.ts
kyubisation: It is no longer necessary, as the setter handles most of the logic anyway.

src/elements/date-input/date-input.component.ts
kyubisation: Parity with the native input. Probably not important, but I prefer it this way.
Or what is your opinion?

src/elements/datepicker/datepicker-toggle/datepicker-toggle.component.ts
kyubisation: We do. It is just `now` that is different.
I also was not quite happy with this, but I'm not sure how much effort to invest in `now`.
Thoughts?

src/elements/date-input/date-input.component.ts
jeripeierSBB: I think I maybe understand it now, because the now property of the calendarElement has a fallback, we can assign null and it fallbacks to today(); but in theory now doesn't allow null.

src/elements/datepicker/datepicker-toggle/datepicker-toggle.component.ts
jeripeierSBB: I think I maybe understand it now, because the now property of the calendarElement has a fallback, we can assign null and it fallbacks to today(); but in theory now doesn't allow null.

src/elements/date-input/date-input.component.ts
jeripeierSBB: ah ok, then let's keep it. I just thought probably we never use the attribute.

src/elements-experimental/timetable-row/timetable-row.visual.spec.ts
```diff
@@ -85,7 +85,7 @@ describe(`sbb-timetable-row`, () => {
       ?loading-trip=${args.loadingTrip}
       ?a11y-footpath=${args.a11yFootpath}
       ?loading-price=${args.loadingPrice}
-      now=${(args.now ?? new Date('2022-12-01T12:11:00').valueOf()) / 1000}
+      now=${args.now ?? new Date('2022-12-01T12:11:00')}
```
jeripeierSBB, line 88:
```suggestion
      .now=${args.now ?? new Date('2022-12-01T12:11:00')}
```

jeripeierSBB (approved):
LGTM, one fix for visual regression tests needed

src/elements/datepicker/datepicker-toggle/datepicker-toggle.component.ts
kyubisation: Ok, no. I think this property will be removed in the refactoring anyway, so I'll leave it as it is.

src/elements-experimental/timetable-row/timetable-row.visual.spec.ts
kyubisation: Thanks! I was trying to figure out where I missed it. 😅

PR #3714, opened 2025-05-15
feat(sbb-sticky-bar): add size s

DavideMininni-Fincons:


src/elements/container/sticky-bar/sticky-bar.visual.spec.ts
```diff
@@ -56,7 +57,7 @@ describe(`sbb-sticky-bar`, () => {
   `;
 
   describeViewports({ viewports: ['zero', 'medium', 'ultra'] }, () => {
-    describeEach(cases, ({ color, containerExpanded, scrolled }) => {
+    describeEach(cases, ({ color, containerExpanded, scrolled, size }) => {
```
jeripeierSBB, line 60: It doesn't really hurt, but I think we could reduce the amount of screenshot drastically by only testing size s in a separate test case. as it only is the block padding so far, every other param here doesn't have any effect. what do you think?

jeripeierSBB (approved):
LGTM, maybe reduce visual snapshots. thanks!

mcilurzo:
LGTM! Thank you.

PR #3721, opened 2025-05-16
fix(sbb-skiplink-list): avoid invisible content block

kyubisation:
With our previous fix, we unintentionally caused the skiplink list to block the content behind it. This fix basically resets that behavior and prevents the skiplink list from taking up space.


TomMenga (approved):
LGTM 👍

PR #3724, opened 2025-05-19
feat: change 'readonly' properties to camelCase

DavideMininni-Fincons:
Not sure if more changes are needed (stories?)


src/elements/alert/alert/readme.md
```diff
@@ -69,7 +69,7 @@ As a base rule, opening animations should be active if an alert arrives after th
...
 | `iconName`     | `icon-name`     | public  | `string`                               | `'info'`           | Name of the icon which will be forward to the nested `sbb-icon`. Choose the icons from https://icons.app.sbb.ch. Styling is optimized for icons of type HIM-CUS. |
 | `isOpen`       | -               | public  | `boolean`                              |                    | Whether the element is open.                                                                                                                                     |
-| `readonly`     | `readonly`      | public  | `boolean`                              | `false`            | Whether the alert is readonly. In readonly mode, there is no dismiss button offered to the user.                                                                 |
+| `readOnly`     | `readOnly`      | public  | `boolean`                              |                    | Whether the alert is readonly. In readonly mode, there is no dismiss button offered to the user.                                                                 |
```
jeripeierSBB, line 72: Probably not solvable, but the attribute should be readonly (lower case)

jeripeierSBB (commented):
Basically LGTM!
Would it make sense to have a readonly mixin? like for required and disabled.
Breaking change notice is missing.

DavideMininni-Fincons:
@jeripeierSBB I thought about the mixin but I didn't add it since it was not requested explicitly, I'm adding it

src/elements/alert/alert/readme.md
DavideMininni-Fincons: I think I can add `attribute: 'readonly'` in the property, just for docs generation

src/elements/select/select.component.ts
```diff
@@ -115,11 +118,12 @@ class SbbSelectElement extends SbbUpdateSchedulerMixin(
...
   @forceType()
-  @handleDistinctChange((e: SbbSelectElement, newValue: boolean) =>
-    e._closeOnDisabledReadonlyChanged(newValue),
-  )
-  @property({ type: Boolean, reflect: true })
-  public accessor readonly: boolean = false;
+  @property({ type: Boolean, attribute: 'readonly' })
+  public override set readOnly(value: boolean) {
+    this._closeOnDisabledReadonlyChanged(value);
+    this.toggleAttribute('readonly', value);
```
jeripeierSBB, line 124: Can we call super here? So attribute creation is not duplicated

jeripeierSBB (approved):
LGTM, one little optimization possible

PR #3786, opened 2025-06-03
fix: avoid exceptions on storybook

jeripeierSBB:
The new storybook version seems to patch focus and blur functions readonly. With the previous implementation this leads to an error. With the new implementation it works.


TomMenga (approved):
Even better now! LGTM 👍

PR #3802, opened 2025-06-11
refactor(sbb-menu, sbb-menu-button, sbb-menu-link): assigned specific role

### LINKED ISSUE #3792: sbb-menu: Items/buttons be role menuitem

TomMenga:
This PR Closes #3792 

To test:
- [x] VoiceOver
- [x] NVDA
- [x] Jaws


TomMenga:
A11y review (translated):

Default:
The disabled option should also receive focus. (Normally, non-selectable elements shouldn't receive keyboard focus, but in this case, users have a hard time identifying the skipped list item. Of course, "unavailable" or "disabled" should be displayed when focusing.)

List:
The disabled option should also receive focus.
The number of list items is not displayed.

Why two implementation variants? Are both needed, or is only one of them ultimately available?

TomMenga:
> Default: The disabled option should also receive focus. (Normally, non-selectable elements shouldn't receive keyboard focus, but in this case, users have a hard time identifying the skipped list item. Of course, "unavailable" or "disabled" should be displayed when focusing.)

Isn't this true for any disabled element we have? Anyway, I believe screen readers are able to navigate through disabled elements (with VoiceOver using CTRL + Option + Arrow)

> List: The disabled option should also receive focus. **The number of list items is not displayed**.

That's a problem we already had, unfortunately :/ 
At least for VoiceOver, the situation slightly improved. Numbers are read if you navigate with  CTRL + Option + Arrow

> Why two implementation variants? Are both needed, or is only one of them ultimately available?

We render the menu as a list if it only contains links and/or buttons. Otherwise, the menu acts as a simple container


I think this is the best result we can achieve for now. @jeripeierSBB what do you think?

PR #3803, opened 2025-06-11
fix(sbb-seat-reservation): clean up exports

jeripeierSBB:


src/elements-experimental/seat-reservation/common/mapper.ts
```diff
@@ -1,2 +1,3 @@
 export * from './mapper/mapper.js';
 export * from './mapper/icon-mapper.js';
+export * from './mapper/seat-reservation-sample-data.js';
```
kyubisation, line 3: Should sample data be in the public API and bundled in production code?

src/elements-experimental/seat-reservation/seat-reservation/seat-reservation-base-element.ts
```diff
@@ -34,6 +34,8 @@ interface CoachScrollTriggerPoint {
   width: number;
 }
 
+export type SeatReservationSelectedPlacesEventDetails = SeatReservationPlaceSelection[];
```
kyubisation, line 37: Is this type really necessary?

kyubisation (approved):
LGTM 👍 
Nice work!

src/elements-experimental/seat-reservation/seat-reservation/seat-reservation-base-element.ts
jeripeierSBB: Probably not, but the react wrapper generator fails with an exception and this was the only work around I found :D. the same for another event details where I had to assign "number" which is even more strange. But looking at the generator script, I couldn't find a solution. Maybe you have an idea?

kyubisation: Good question. I think it is fine for now.

PR #3808, opened 2025-06-12
refactor: remove table border radius

jeripeierSBB:


jeripeierSBB:
Yeah originally I completely refactored it, but then had to revert it because of the sticky case 😅

Issue #3610, opened 2025-04-17
task: check translations


jeripeierSBB:
We have to normalize our translations for the T-V distinction. https://en.wikipedia.org/wiki/T%E2%80%93V_distinction

 - [x] EN: (should be just ok)
 - [x] IT: informal
 - [x] FR: formal
 - [x] DE: formal

PR #3817, opened 2025-06-15
fix: improve translations

jeripeierSBB:
Contributing to #3610


TomMenga (approved):
Can't judge for the DE and FR, but IT is LGTM

PR #3825, opened 2025-06-16
fix(sbb-perl-chain): handle partially canceled trips

christoph-bittmann:
closes #3821


jeripeierSBB:
@MarioCastigliano For your information, maybe needs a fix on new pearl chain too

PR #3830, opened 2025-06-17
fix: removal of ariaRoleOnHost [WIP]

DavideMininni-Fincons:
DO NOT REVIEW
deployment needed to test JAWS/NVDA on the VDI


DavideMininni-Fincons:
Tested with NVDA, optgroups are not read. Both issue and PR can be closed for now and tested again in the future.

Main branch vs this PR:
image
image

PR #3837, opened 2025-06-19
refactor: events renaming

### LINKED ISSUE #3793: Event renaming
For event names we initially adopted the naming schema from Stencil.js (will*, did*).
This worked fine but there was doubt about consistency with the web platform.

Now looking at the pattern of native events, it follows `beforeeventname` and `eventname`.
We want to adapt our event naming schema with this pattern in mind.

Examples 

- https://developer.mozilla.org/en-US/docs/Web/API/Element/beforeinput_event
- https://developer.mozilla.org/en-US/docs/Web/API/Element/beforematch_event
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/beforetoggle_event
- https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
- https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeprint_event

Also adapt the Angular Wrapper, after this change has been released.

TomMenga:
BREAKING CHANGE: Renamed the event names to lowercase to adhere to web standards. The following events have been renamed:
- `willOpen` => `beforeopen`
- `didOpen` => `open`
- `willClose` => `beforeclose`
- `didClose` => `close`
- `willStick` => `beforestick` 
- `didStick` => `stick`,
- `willUnstick` => `beforeunstick`
- `didUnstick` => `unstick`
- `optionSelectionChange` => `optionselectionchange`
- `optionSelected` => `optionselected`
- `dateSelected` => `dateselected`
- `chipInputTokenEnd` => `chipinputtokenend`
- `toggleExpanded` => `toggleexpanded`
- `fileChanged` => `filechanged`

This PR Closes #3793

List of changes:

### Internal (not included in the breaking change message)
‘displayValueChange’ => ‘displayvaluechange’
‘panelConnected’ => ‘panelconnected’
‘stateChange’ => ‘statechange
‘requestDelete’ => ‘requestdelete’
‘resizeChange’ => ‘resizechange’
‘trainSlotChange’ => ‘trainslotchange’
‘sectorChange’ => ‘sectorchange’


src/elements/select/select.component.ts
```diff
@@ -31,7 +31,7 @@ import {
   SbbUpdateSchedulerMixin,
 } from '../core/mixins.js';
 import { isEventOnElement, overlayGapFixCorners, setOverlayPosition } from '../core/overlay.js';
-import type { SbbOptGroupElement, SbbOptionElement } from '../option.js';
+import { type SbbOptGroupElement, SbbOptionElement } from '../option.js';
```
jeripeierSBB, line 34: Please avoid importing optionelement

src/elements/select/select.component.ts
```diff
@@ -192,7 +192,7 @@ class SbbSelectElement<T = string> extends SbbUpdateSchedulerMixin(
 
   public constructor() {
     super();
-    this.addEventListener?.('optionSelectionChange', (e: CustomEvent<void>) =>
+    this.addEventListener?.(SbbOptionElement.events.selectionchange, (e: CustomEvent<void>) =>
```
jeripeierSBB, line 195: please only use string here to avoid importing the component

src/elements/autocomplete/autocomplete.component.ts
```diff
@@ -3,7 +3,7 @@ import { customElement } from 'lit/decorators.js';
 import { getNextElementIndex } from '../core/a11y.js';
 import { isSafari } from '../core/dom.js';
 import { setAriaComboBoxAttributes } from '../core/overlay.js';
-import type { SbbOptGroupElement, SbbOptionElement } from '../option.js';
+import { type SbbOptGroupElement, SbbOptionElement } from '../option.js';
```
jeripeierSBB, line 6: Please avoid importing optionelement

src/elements/expansion-panel/expansion-panel/expansion-panel.component.ts
```diff
@@ -10,7 +10,7 @@ import type { SbbOpenedClosedState } from '../../core/interfaces.js';
 import { SbbHydrationMixin } from '../../core/mixins.js';
 import type { SbbTitleLevel } from '../../title.js';
 import type { SbbExpansionPanelContentElement } from '../expansion-panel-content.js';
-import type { SbbExpansionPanelHeaderElement } from '../expansion-panel-header.js';
+import { SbbExpansionPanelHeaderElement } from '../expansion-panel-header.js';
```
jeripeierSBB, line 13: Please avoid importing header element

src/elements/stepper/stepper/stepper.component.ts
```diff
@@ -13,7 +13,7 @@ import { forceType } from '../../core/decorators.js';
 import { breakpoints, isBreakpoint, isLean } from '../../core/dom.js';
 import type { SbbHorizontalFrom, SbbOrientation } from '../../core/interfaces.js';
 import { SbbHydrationMixin } from '../../core/mixins.js';
-import type { SbbStepElement, SbbStepValidateEventDetails } from '../step.js';
+import { SbbStepElement, type SbbStepValidateEventDetails } from '../step.js';
```
jeripeierSBB, line 16: Please avoid importing step element

src/elements/train/train-formation/train-formation.component.ts
```diff
@@ -12,8 +12,8 @@ import { SbbLanguageController } from '../../core/controllers.js';
 import { i18nSector, i18nSectorShort, i18nTrains } from '../../core/i18n.js';
 import { SbbNamedSlotListMixin, type WithListChildren } from '../../core/mixins.js';
 import type { SbbTrainBlockedPassageElement } from '../train-blocked-passage.js';
-import type { SbbTrainWagonElement } from '../train-wagon.js';
-import type { SbbTrainElement } from '../train.js';
+import { SbbTrainWagonElement } from '../train-wagon.js';
```
jeripeierSBB, line 15: Please avoid importing traingwagon and train

TomMenga:
> I think importing elements into others should be avoided, although you only did it where it's expected to have them loaded by consumer anyways. @kyubisation what is your opinion?

Yes, adding the dependency should be fine in this case, but if we want to avoid it, could we reconsider having the events as a static const?
It hurts manutenibility using event names as 'strings' :/

kyubisation:
I would like to avoid importing a component just for the event names.
I'm aware that this is a typing issue, for which we do not yet have a clean solution.
We could potentially create an eslint rule for this?

src/elements/autocomplete/autocomplete.spec.ts
```diff
@@ -589,7 +589,7 @@ describe(`sbb-autocomplete`, () => {
       input = formField.querySelector('input')!;
     });
 
-    it('should select value', async () => {
+    it.only('should select value', async () => {
```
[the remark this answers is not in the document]
TomMenga: I swear i'll write an eslit rule just for this :|

TomMenga:
> I would like to avoid importing a component just for the event names. I'm aware that this is a typing issue, for which we do not yet have a clean solution. We could potentially create an eslint rule for this?

I reverted to using strings for now. Let's discuss possibilities in the next refinement :)

src/elements/file-selector/common/file-selector-common.spec.ts
```diff
@@ -5,10 +5,9 @@ import type { SbbSecondaryButtonElement } from '../../button.js';
 import { fixture } from '../../core/testing/private.js';
 import { EventSpy, waitForLitRender } from '../../core/testing.js';
 import type { SbbFileSelectorDropzoneElement } from '../file-selector-dropzone.js';
-import type { SbbFileSelectorElement } from '../file-selector.js';
+import { SbbFileSelectorElement } from '../file-selector.js';
```
DavideMininni-Fincons, line 8: Same as other comments

DavideMininni-Fincons (commented):
LGTM, a couple of fix needed

src/elements/autocomplete-grid/autocomplete-grid/autocomplete-grid.component.ts
```diff
@@ -57,8 +57,9 @@ class SbbAutocompleteGridElement<T = string> extends SbbAutocompleteBaseElement<
 
   public constructor() {
     super();
-    this.addEventListener?.('autocompleteOptionSelected', (e: CustomEvent<void>) =>
-      this.onOptionSelected(e),
+    this.addEventListener?.(
+      SbbAutocompleteGridOptionElement.events.optionselected,
```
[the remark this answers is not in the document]
TomMenga: In this case, we already have to import the `SbbAutocompleteGridOptionElement` for other uses. We can use the event constant without drawbacks

src/elements/datepicker/datepicker/datepicker.spec.ts
```diff
@@ -156,17 +157,17 @@ describe(`sbb-datepicker`, () => {
...
 
     toggle.click();
-    await didOpenEventSpy.calledOnce();
+    await openSpy.calledOnce();
     expect(datepicker).to.have.attribute('data-state', 'opened');
 
     const calendar = datepicker.shadowRoot!.querySelector('sbb-calendar')!;
     calendar.dispatchEvent(
-      new CustomEvent('dateSelected', {
+      new CustomEvent(SbbCalendarElement.events.dateselected, {
```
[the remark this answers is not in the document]
TomMenga: It's not a problem to import components in a spec file (since it's not delivered).
If we can, we would rather always use the constants

src/elements/file-selector/common/file-selector-common.spec.ts
TomMenga: It's not a problem to import components in a spec file (since it's not delivered).
If we can, we would rather always use the constants

PR #3845, opened 2025-06-20
refactor: use aria properties where possible

kyubisation:
This PR refactors usage of aria attributes to use aria properties where possible. This reduces the amount of attributes visible in the DOM.

### DIFF
--- a/src/elements/form-field/form-field/form-field.spec.ts
+++ b/src/elements/form-field/form-field/form-field.spec.ts
@@ -87,7 +87,7 @@ describe(`sbb-form-field`, () => {
       await nextFrame();
 
       // Then input should be linked and sbb-form-error configured
-      expect(input.ariaErrorMessageElements).to.have.same.members([formError]);
+      expect(input.ariaDescribedByElements).to.have.same.members([formError]);
       expect(formError).to.have.attribute('role', 'status');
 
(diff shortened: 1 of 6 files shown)


src/elements/form-field/form-field/form-field.spec.ts
```diff
@@ -100,28 +97,6 @@ describe(`sbb-form-field`, () => {
       // Then aria-describedby should be removed
       expect(input).not.to.have.attribute('aria-describedby');
     });
-
-    it('should reference sbb-form-error with original aria-describedby', async () => {
```
jeripeierSBB, line 104: Would it be worth to replicate it?

src/elements/stepper/step/step.component.ts
```diff
@@ -70,9 +69,25 @@ class SbbStepElement extends SbbElementInternalsMixin(LitElement) {
   });
 
   /** The label of the step. */
+  // eslint-disable-next-line @typescript-eslint/naming-convention
+  private set label(value: SbbStepLabelElement | null) {
+    if (this._label && this.internals.ariaLabelledByElements?.length) {
```
jeripeierSBB, line 74: check if empty

jeripeierSBB (approved):
LGTM :-)

PR #3890, opened 2025-07-03
perf(sbb-datepicker): delay rendering of calendar

kyubisation:
Currently the `<sbb-calendar>` is rendered synchronously inside the `<sbb-datepicker>`. As the calendar is only needed once the datepicker needs to be opened, we can safely delay the rendering of the calendar.

Closes sbb-design-systems/lyne-angular#138


kyubisation:
@k-genov We have done some performance testing and this PR should improve the initial rendering significantly. However, it will never be as fast as a native input element.

jeripeierSBB (approved):
LGTM, thanks!

PR #3906, opened 2025-07-10
feat(sbb-form-field): add support for custom controls

kyubisation:
This PR introduces the possibility of defining custom form controls for `<sbb-form-field>`.
The basic functionality is based on implementing the `SbbFormFieldElementControl` and dispatching the `SbbFormFieldControlEvent` on the related `<sbb-form-field>` instance.

This is the basis for supporting sbb-design-systems/lyne-angular#55

### DIFF
--- a/src/elements/form-field/form-field/form-field.component.ts
+++ b/src/elements/form-field/form-field/form-field.component.ts
@@ -33,6 +33,34 @@ let nextId = 0;
 const patchedInputs = new WeakMap<HTMLInputElement, PropertyDescriptor>();
 const nativeInputElements = ['input', 'textarea', 'select'];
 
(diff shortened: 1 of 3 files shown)


TomMenga (approved):
Implementation looks good 👍 

The only potential problem I see is that I would have expected that the `custom-input-element` had to implement the `SbbFormFieldElementControl` interface (similar to the `ControlValueAccessor` in Angular).

Instead, we are opting to have a different object that connects the actual input and the form-field. 
I'm not saying it's worse, just that it's not what I would have expected. I would not extend my expectations to everyone, but I think that an explicit example is needed in the readme

jeripeierSBB (approved):
LGTM nr 2 :)

PR #3943, opened 2025-08-05
fix(sbb-autocomplete): update active option if autoActiveFirstOption is enabled and options change

### LINKED ISSUE #3940: [Bug]: autocomplete autoActiveFirstOption with lazy options + disabled default option
### Preflight Checklist

### Bug type

Functionality

Components

3.2.0 / 20.0.0-rc.1

### Operating system

win10

### Browser / Browser version

chrome latest

### Input mode

None

### Additional settings

_No response_

### What happened?

Under specific cirumstances the autoActiveFirstOption does produce errors (see below):

1. The options change when the autocomplete opens/closes
2. The options of the closed autocomplete are disabled

While this may look like an extreme corner case, I'd argue this is relevant: 
- lazyly computing the options only for opened autocompletes showed major peroformance improvments for our app, as it contained many autocompletes (in the same view) with non-trivial computation for the options (different groups depending on given state).
- the disabled option on the closed autocomple is only used as a placeholder, so that the openSignal fires and we can then compute the relevant options. It is disabled, because it should not be selectable.

To put it differently: the first option should be activated not only on open, but should be recalculated also when the options change (this is regularly happening e.g. when filtering the options).

1. open https://stackblitz.com/edit/lyne-angular-starter-e2yegqzt?file=src%2Fmain.ts
2. click in the input
3. observe the error in the developer console

### Relevant log output

```shell
ERROR TypeError: Cannot read properties of undefined (reading 'setActive')
    at _a.setNextActiveOption (autocomplete.component.js:59:24)
    at _a.open (autocomplete-base-element.js:448:14)
    at triggerElement.addEventListener.signal (autocomplete-base-element.js:582:64)
    at _ZoneDelegate.invokeTask (zone.js:431:33)
    at debug_node.mjs:16330:55
    at AsyncStackTaggingZoneSpec.onInvokeTask (debug_node.mjs:16330:36)
    at _ZoneDelegate.invokeTask (zone.js:430:38)
    at Object.onInvokeTask (debug_node.mjs:16653:33)
    at _ZoneDelegate.invokeTask (zone.js:430:38)
    at ZoneImpl.runTask (zone.js:161:47)
```

jeripeierSBB:
Closes #3940

### DIFF
--- a/src/elements/autocomplete-grid/autocomplete-grid/autocomplete-grid.component.ts
+++ b/src/elements/autocomplete-grid/autocomplete-grid/autocomplete-grid.component.ts
@@ -98,7 +98,10 @@ class SbbAutocompleteGridElement<T = string> extends SbbAutocompleteBaseElement<
       this.activeOption
         ?.closest('sbb-autocomplete-grid-row')
         ?.querySelectorAll('sbb-autocomplete-grid-button')
-        [this._activeColumnIndex - 1]?.click();
+        [
+          // We ignore the option in the selector. Therefore, we have to shift the activeColumnIndex by one.
+          this._activeColumnIndex - 1
+        ]?.click();
     } else {
       this.activeOption?.['selectViaUserInteraction'](true);
     }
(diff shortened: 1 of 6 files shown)


DavideMininni-Fincons (approved):
LGTM, thanks!

PR #3955, opened 2025-08-11
fix(sbb-autocomplete, sbb-autocomplete-grid): only opening on lazy options if count was 0

jeripeierSBB:
Closes [#lyne-angular/159](https://github.com/sbb-design-systems/lyne-angular/issues/159)


DavideMininni-Fincons (approved):
LGTM, thank you!

PR #3994, opened 2025-08-26
feat(sbb-dialog): expose announceTitle method to announce title at any time

### LINKED ISSUE #3982: [Feature request]: sbb-dialog should announce changes of accessibilityLabel in overlay live region
### Preflight Checklist

### Request type

Components

### Functionality

**Situation:**

We are using the `sbb-dialog` component in such a way that requires us to change the title text of the dialog while it is already open. Think of it as switching views inside the dialog.

Currently, the live region component `sbb-screen-reader-only` created by `sbb-dialog` only announces the dialog title right after the dialog is opened. Subsequent changes of the title - no matter if text content of `sbb-dialog-title` is updated or property `accessibilityLabel` is changed - are not announced which makes it hard for screen reader users to recognize that the view has changed.

**Proposal:**

Everytime the property `accessibilityLabel` is updated with a value that differs from the current one this new label text should be set in the live region to announce this new dialog title.

The reset of the live region can stay as is, i.e. on any interaction within the dialog the live region is cleared.

**Workaround:**

We have implemented a workaround in our Next.js project which involves updating the live region using a ref callback and some Shadow DOM manipulation. If someone is interested in that code, drop a comment.

_No response_

jeripeierSBB:
Closes #3982

### DIFF
--- a/src/elements/dialog/dialog/dialog.component.ts
+++ b/src/elements/dialog/dialog/dialog.component.ts
@@ -64,6 +64,13 @@ class SbbDialogElement extends SbbOverlayBaseElement {
     super.connectedCallback();
   }
 
+  /** Announce the accessibility label or dialog title for screen readers. */
+  public announceTitle(): void {
+    this.setAriaLiveRefContent(
+      this.accessibilityLabel || this.querySelector('sbb-dialog-title')?.innerText.trim(),
+    );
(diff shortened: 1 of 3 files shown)


kyubisation (approved):
LGTM 👍

PR #4039, opened 2025-09-17
fix(sbb-radio-button, sbb-radio-button-panel): remove label tag from Shadow DOM

DavideMininni-Fincons:


DavideMininni-Fincons:
> LGTM, thank you! Clicking on the text label does still toggle the state, right?

Yes, it does
related topic: in radios stories we don't have a control for the allow-empty-selection attribute, is it wanted or do we just forgot to add it?

jeripeierSBB:
> > LGTM, thank you! Clicking on the text label does still toggle the state, right?
> 
> Yes, it does related topic: in radios stories we don't have a control for the allow-empty-selection attribute, is it wanted or do we just forgot to add it?

Probably forgotten. What does @TomMenga  the ink?

PR #4053, opened 2025-09-19
feat(sbb-tooltip): add global delay attributes

DavideMininni-Fincons:


src/elements/tooltip/tooltip.component.ts
```diff
@@ -154,24 +154,50 @@ class SbbTooltipElement extends SbbDisabledMixin(SbbOpenCloseBaseElement) {
...
+          const target = mutation.target as HTMLElement;
+          if (mutation.attributeName === 'sbb-tooltip') {
+            this._handleTooltipTrigger(target);
+          } else if (delayAttributeMap.has(mutation.attributeName!)) {
+            this._setDelayProperty(
+              target,
+              mutation.attributeName!,
+              delayAttributeMap.get(mutation.attributeName!)!,
+            );
+          }
```
TomMenga, line 178: I think the right place to do this is inside the `_handleTooltipTrigger` method, after `tooltip.textContent = tooltipMessage;`. That point is called whenever there is a new tooltip or when the attributes are modified.

You could just do something like `tooltip.openDelay = triggerElement.getAttribute('sbb-tooltip-open-delay');`

src/elements/tooltip/tooltip.spec.ts
```diff
@@ -284,4 +284,54 @@ describe('sbb-tooltip', () => {
       expect(tooltipOutlet!.children[0].localName).to.equal('sbb-tooltip');
     });
   });
+
+  describe('delay attributes', () => {
```
TomMenga, line 288:
```suggestion
  describe('attribute usages - delay attributes', () => {
```

Good set of test, i would just change the name

TomMenga (changes requested):
GG 👍 
I propose a different approach, if you wish we can discuss it

src/elements/tooltip/readme.md
```diff
@@ -42,6 +42,14 @@ mergeConfig({
 });
 ```
 
+If the component is used via attribute usage, it's possible to set the delays using the `sbb-tooltip-open-delay` and the `sbb-tooltip-close-delay` attributes.
```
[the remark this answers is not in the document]
DavideMininni-Fincons: I rewrote it a bit; what do you think?

PR #4079, opened 2025-10-02
fix(sbb-paginator, sbb-compact-pagintor): check isUpdated to avoid the 'page' event emission

DavideMininni-Fincons:


kyubisation (commented):
LGTM 👍 
Is this testable? I guess it would be too difficult to get the timing right?

src/elements/paginator/common/paginator-common.ts
```diff
@@ -203,8 +203,13 @@ export const SbbPaginatorCommonElementMixin = <T extends AbstractConstructor<Lit
     }
 
     protected emitPageEvent(previousPageIndex: number): void {
+      // When the paginator is used in a routed component,
+      // the hasUpdated flag is true after the first property's set is called.
+      // Adding the check on `this.isUpdatePending` seems to solve the issue.
+      // https://github.com/sbb-design-systems/lyne-components/issues/4059
       if (
         !this.hasUpdated ||
+        this.isUpdatePending ||
```
TomMenga, line 212: I don't think this is a stable solution to the problem :/
You're just lucky that another property has been set before the index one and initiated an update cycle.
The real problem is that the inputs (pageIndex, pageSize and length) all trigger a recalculation of the pageIndex (on a state which is only partial)

If this happens after the 'firstUpdate', you find yourself in a case like the one described in the bug

I might have an idea on how to fix it. Let's have a call

jeripeierSBB (commented):
Thank you for having a look at this. I already spent hours and variants of solutions for that topic (but not the specific bug) :D. I think with the new approach we have again problems in the angular wrapper using it with sbbtabledatasource. So could you maybe bundle and test it manually in the lyne-angular repo? the datasource also needs the initial values.
I think the cleanest approach long term would be that we have a kind of internal event which emits always to work together with angular wrapper and a second, public one, which really only triggers on user change. but this would also need more discussion.

src/elements/paginator/paginator/paginator.stories.ts
```diff
@@ -223,7 +223,7 @@ const meta: Meta = {
   decorators: [withActions as Decorator],
   parameters: {
     actions: {
-      handles: [SbbPaginatorElement.events.page],
+      handles: [SbbPaginatorElement.events.page, SbbPaginatorElement.events.ɵpage],
```
jeripeierSBB: Should probably be hidden here

src/elements/paginator/common/paginator-common.ts
```diff
@@ -247,7 +281,7 @@ export const SbbPaginatorCommonElementMixin = <T extends AbstractConstructor<Lit
             icon-name="chevron-small-left-small"
             ?disabled=${this.disabled || !this.hasPreviousPage()}
             @click=${() => {
-              this.previousPage();
+              this.changeAndEmitPage(this.pageIndex - 1);
```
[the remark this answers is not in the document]
DavideMininni-Fincons: The pageIndex is changed before the emission;
I changed the story's template to 

```ts
const Template = ({ pageSizeOptions, ...args }: Args): TemplateResult => {
  return html` <sbb-paginator aria-label="Select page" .pageSizeOptions=${pageSizeOptions} ${sbbSpread(args)}
    @page=${(e) => {
      console.log((document.querySelector('sbb-paginator') as SbbPaginatorElement).pageIndex);
      console.log(e.detail);
  }}
  ></sbb-paginator>`;
};
```

and I see the pageIndex is correctly set, and the event.detail has the right properties.

PR #4085, opened 2025-10-06
feat(sbb-autocomplete-grid): enable `sbb-options-nowrap` class

jeripeierSBB:


src/elements/autocomplete-grid/autocomplete-grid-row/autocomplete-grid-row.scss
```diff
@@ -55,6 +55,9 @@
 ::slotted(sbb-autocomplete-grid-option) {
   flex: 1 1 auto;
   margin-right: calc(-1 * var(--sbb-spacing-fixed-2x));
+
+  // Used to enable ellipsis
+  overflow-x: hidden;
```
kyubisation, line 60: Does this need to be hardcoded here?

kyubisation (approved):
LGTM 👍 
Thanks!

src/elements/autocomplete-grid/autocomplete-grid-row/autocomplete-grid-row.scss
jeripeierSBB: As offline discussed, it should be ok for all contexts.

PR #4130, opened 2025-10-27
refactor(sbb-dialog): refactor dialog to support Angular overlay services

jeripeierSBB:
BREAKING CHANGE: removed slots `title-section`, `actions` and CSS part `scroll-container`.

### DIFF
--- a/src/elements/core/testing/private/visual-regression-snapshot.ts
+++ b/src/elements/core/testing/private/visual-regression-snapshot.ts
@@ -85,12 +85,10 @@ const runSetupWithViewport = async (
 ): Promise<VisualDiffSetupBuilder> => {
   const builder = new VisualDiffSetupBuilder();
   await setup(builder);
-
-  await builder.postSetupAction();
-
   if (viewport) {
     await setViewport(viewport);
   }
+  await builder.postSetupAction();
 
   return builder;
 };


src/elements/dialog/dialog/dialog.component.ts
```diff
@@ -146,6 +157,26 @@ class SbbDialogElement extends SbbOverlayBaseElement {
...
 
+  private _detectIntermediateElement(): void {
+    const potentialChildElements = ['sbb-dialog-title', 'sbb-dialog-content', 'sbb-dialog-actions'];
+
+    const hasOnlyOneDirectChild =
+      Array.from(this.children).filter((el) => el instanceof HTMLElement).length === 1;
+    const noExpectedElementAsDirectChild = potentialChildElements.every(
+      (name) => this.firstElementChild?.localName !== name,
+    );
+    const intermediateElementHasExpectedChildren = Array.from(
```
kyubisation: Are we sure that the children are already attached when the intermediate element is slotted?

src/elements/dialog/dialog/dialog.stories.ts
```diff
@@ -286,7 +286,7 @@ const StepperTemplate = ({
   <sbb-dialog trigger="dialog-trigger" ${sbbSpread(args)}>
     ${dialogTitle(level)}
     ${includeCloseButton ? html`<sbb-dialog-close-button></sbb-dialog-close-button>` : nothing}
-    <sbb-dialog-content>
+    <sbb-dialog-content style="padding-block-start: 1rem">
```
kyubisation: Why is this necessary? I worry about additional style attributes in stories, as they might be copied without understanding.

src/elements/dialog/dialog/dialog.component.ts
jeripeierSBB: only use noExpectedElementAsDirectChild

PR #4142, opened 2025-10-28
fix(sbb-menu): remove menu items as list

DavideMininni-Fincons:
BREAKING CHANGE: the sbb-menu doesn't render a list of sbb-menu-button/sbb-menu-link as ul/li anymore. This pattern uses 'list'/'listitem' roles which interfere with the assigned 'menu'/'menuitem' roles when screenreaders are used.

---

Accessibility review:
in the list implementation elements with role="menuitem" are nested inside elements with role="listitem"; both NVDA and JAWS don't read the list index (VoiceOver does).

A possible solution suggested by Anastasiia could be remove the roles on sbb-menu and sbb-menu-button/link and keep the ul /li implementation , but since they are readonly they can't be directly reassigned; 
for menu and menu-button, it could be possible to do `element['internals'].role = null`, but menu-link doesn't use internals and it uses a protected readonly prop that sets the role on the inner anchor tag instead than on the host.

So the other solution is to totally remove the list implementation; 
quoting Anastasiia, 'The implementation “menu-default” is announced correctly because the ARIA roles menu and menuitem are used properly.'

NOTE:
are there other components affected?

### DIFF
--- a/src/elements/menu/menu/readme.md
+++ b/src/elements/menu/menu/readme.md
@@ -70,6 +70,8 @@ Please note that nesting the menus in DOM is not supported. The `sbb-menu` eleme
 
 ## Accessibility
 
+The `sbb-menu` uses the roles `menu` and `menuitem` for children.
+
 As the menu opens, the focus will automatically be set to the first focusable
 item within the component (unless manually specified, see below).
 When using the `sbb-menu` as a select (e.g. language selection) it's recommended to use the `aria-pressed` attribute


jeripeierSBB (approved):
Would like to discuss also with @kyubisation . Please add also a BREAKING CHANGE: notice to the PR body.

src/elements/menu/menu/readme.md
```diff
@@ -68,11 +68,6 @@ Please note that nesting the menus in DOM is not supported. The `sbb-menu` eleme
...
 ```
 
-## Style
-
-If only `sbb-menu-button`/`sbb-menu-link` components are provided, the items are automatically grouped within a list
-using `<ul>` and `<li>` items, for more complex scenarios the grouping must be done manually.
-
 ## Accessibility
 
 As the menu opens, the focus will automatically be set to the first focusable
```
kyubisation:
```suggestion
The `sbb-menu` uses the role `menu` and `menuitem` for children.

As the menu opens, the focus will automatically be set to the first focusable
```

PR #4166, opened 2025-11-05
docs: improve events and properties JSDoc

DavideMininni-Fincons:
Changes:
- fix wrong jsdoc opening tags  - `/**` instead than `/*`;
- add missing jsdoc for input/change events.

Useful for the docs-app in the angular repo.


jeripeierSBB (approved):
LGTM, could also go into 3.x I think

PR #4200, opened 2025-11-17
refactor: use controller to sync ancestor properties

kyubisation:
This PR refactors the ancestor property sync functionality.
For this it implements the `SbbAncestorWatcherController` class to keep track of the property states and react on changes.


src/elements/button/common/mini-button-common.scss
```diff
@@ -1,6 +1,6 @@
 @use '../../core/styles' as sbb;
 
-$disabled: '[disabled], :disabled, [disabled-interactive], [data-disabled], [data-group-disabled]';
+$disabled: '[disabled], :disabled, [disabled-interactive], [data-disabled], :state(disabled)';
```
jeripeierSBB: do you know in which case the mini button receives :state(disabled)? maybe it's obsoloete?

src/elements/core/controllers/ancestor-watcher-controller.ts
```diff
@@ -0,0 +1,120 @@
...
+            if (!this._watchers!.size && this._ancestor) {
+              propertyWatchers.delete(this._ancestor);
+            }
+          }
+        }
+      }
+    }
+  }
+
+  private _isTargetAncestor(): boolean {
```
jeripeierSBB: I don't get what is handled in this method and why the removal only happens if its not "isTargetAncestor"?

PR #4202, opened 2025-11-17
refactor(sbb-error): rename `sbb-form-error` to `sbb-error`

kyubisation:
BREAKING CHANGE: `sbb-form-error` has been renamed to `sbb-error` and is now part of the `form-field` module.

### DIFF
--- a/src/elements/checkbox/checkbox-group/checkbox-group.stories.ts
+++ b/src/elements/checkbox/checkbox-group/checkbox-group.stories.ts
@@ -405,13 +405,13 @@ export const verticalIconEndLongLabel: StoryObj = {
   args: { ...checkboxArgsVertical, ...iconEnd, label: longLabelText },
 };
 
-export const horizontalWithSbbFormError: StoryObj = {
+export const horizontalWitherror: StoryObj = {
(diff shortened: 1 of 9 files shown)


src/elements/checkbox/checkbox-group/checkbox-group.stories.ts
```diff
@@ -112,9 +112,7 @@ const ErrorMessageTemplate = ({
 }: Args): TemplateResult => html`
   <sbb-checkbox-group ${sbbSpread(args)} id="sbb-checkbox-group">
     ${checkboxes(checked, disabledSingle, iconName, iconPlacement, label)}
-    ${args.required
-      ? html`<sbb-form-error slot="error">This is a required field.</sbb-form-error>`
-      : nothing}
+    ${args.required ? html`<sbb-error slot="error">This is a required field.</sbb-error>` : nothing}
```
jeripeierSBB: We seem to have cases where no form-field is present. Does to module change still make sense?

jeripeierSBB: I think yes, and probably you as well :P

PR #4203, opened 2025-11-17
fix(lists): display icon with forced colors

DavideMininni-Fincons:
Two different cases:
1. for the default icon case, it would be enough to set the `--sbb-icon-list-marker-icon-color` to `CanvasText` in high contrast mode; however,
2. this would not cover the case when consumers override the variable to set a custom color via inline style, due to its higher specificity; so, instead than overriding the variable, it is necessary to override the background rule.


jeripeierSBB (approved):
LGTM, nice solution

PR #4216, opened 2025-11-20
refactor: remove obsolete CSS variables

jeripeierSBB:


scripts/find-obsolete-css-variables.ts
```diff
@@ -0,0 +1,34 @@
+import { readFileSync } from 'node:fs';
+
+import * as glob from 'glob';
```
kyubisation: Nitpick; Preferably use the native glob
https://nodejs.org/api/fs.html#fsglobsyncpattern-options

PR #4231, opened 2025-11-25
fix(sbb-popover): remove duplicate folder

DavideMininni-Fincons:
After the removal of the sbb-popover-trigger, the sbb-popover has a module structure with a duplicate 'popover' folder. 
This is unwanted.

The lyne-angular library should be updated after this is merged to adapt to the new structure.

BEGIN_COMMIT_OVERRIDE
fix(sbb-popover): remove duplicate folder

BREAKING CHANGE: the component import should be checked and possibly adapted with the new path, e.g.: 
```ts
// old 
import '@sbb-esta/lyne-elements/popover/popover.js';`
// new 
import '@sbb-esta/lyne-elements/popover.js';
```

END_COMMIT_OVERRIDE

### DIFF
--- a/src/elements/popover/popover.stories.ts
+++ b/src/elements/popover/popover.stories.ts
@@ -228,7 +228,7 @@ const meta: Meta = {
       extractComponentDescription: () => readme,
     },
   },
-  title: 'elements/sbb-popover/sbb-popover',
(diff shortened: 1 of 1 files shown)


jeripeierSBB (approved):
LGTM, thank you!

Issue #3969, opened 2025-08-14
CSS Refactoring


jeripeierSBB:
To optimize our styles we would like to refactor various topics:
- As pre condition per component we have to ensure that visual tests are as complete as possible and also include forced colors.
- Default values of component properties should not be used as selector in the styles as in future lit won't reflect them anymore. One solution is to use :not() selector if not possible otherwise. 
- Where possible, minimize payload. Especially the font styles are duplicated a lot. Titles are already optimized.
- CSS should be optimized to use with SSR, like move more styles in global stylesheets instead.
- Lean and off brand theming should be made as separat files and without using javascript.
- Dark mode values should be implemented for every component
- Where possible using context specific colors to make it easier to override, e.g. for off brand theme.
- Minimize amount of data-attributes by either using custom states or by using :has() selectors in global styles.
- For simple components, consider adding the styles to the host to minimize DOM elements and increasing consumer flexibility.
- We have to adapt the styles to the reduced breakpoints.

PR #4351, opened 2026-01-08
refactor(sbb-expansion-panel): css refactoring

TomMenga:
Part of #3969


src/elements/expansion-panel/expansion-panel-content/expansion-panel-content.global.scss
```diff
@@ -0,0 +1,10 @@
+@use '../../core/styles' as sbb;
+
+$theme: 'standard' !default;
+
+@mixin base {
+  --sbb-expansion-panel-content-padding-inline: var(--sbb-spacing-fixed-6x);
+  --sbb-expansion-panel-content-padding-inline-start: var(
```
jeripeierSBB: I think I'd keep the inline start in the shadow CSS. because if you override "--sbb-expansion-panel-content-padding-inline" ond expansion panel level (what is expected consumer behavior), the inline start gets the wrong value.

src/elements/expansion-panel/expansion-panel/expansion-panel.scss
```diff
@@ -11,21 +11,6 @@ $open-anim-opacity-to: 1;
     --sbb-disable-animation-duration,
```
jeripeierSBB: in the expansion panel there is only one div contained. this could be removed and styles could be applied to the host (also expansion-panel-content)

src/elements/expansion-panel/expansion-panel/expansion-panel.scss
jeripeierSBB: Could you check whether nested expansion panels are fully supported? It seems that we don't have any tests or stories so far

src/elements/expansion-panel/expansion-panel-header/expansion-panel-header.scss
```diff
@@ -53,6 +56,9 @@
   text-align: start;
   cursor: var(--sbb-expansion-panel-header-cursor);
   color: var(--sbb-expansion-panel-header-text-color);
+  font-size: var(--sbb-expansion-panel-header-font-size);
+  font-weight: normal;
```
jeripeierSBB: is font-weight needed?

TomMenga: Yep, to override the `h` tag weight

src/elements/expansion-panel/expansion-panel/expansion-panel.stories.ts
```diff
@@ -158,6 +158,26 @@ const TemplateSlottedIcon = ({
...
 `;
 
+const NestedTemplate = ({
+  headerText,
+  iconName,
+  contentText,
+  'disabled-interactive': disabledInteractive,
+  ...args
+}: Args): TemplateResult => html`
+  <sbb-expansion-panel ${sbbSpread(args)}>
```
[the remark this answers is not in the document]
jeripeierSBB: does maybe contain: layout or contain: paint work instead of overflow:hidden?

PR #4365, opened 2026-01-13
style(sbb-card): handle nested use case

TomMenga:
Part of #3969


src/elements/card/card/card.stories.ts
```diff
@@ -118,7 +118,7 @@ const spacing: InputType = {
   control: {
     type: 'inline-radio',
   },
-  options: ['3x-xxs', 'xxxs-xxs', 'xxxs-s', '4x-xxs', 'xxs-xxs', 's-s', 'l-l'],
```
jeripeierSBB: nice catch

PR #4384, opened 2026-01-19
refactor(sbb-divider): css refactoring

jeripeierSBB:
Part of #3969

### DIFF
--- a/src/elements/divider/divider.scss
+++ b/src/elements/divider/divider.scss
@@ -4,6 +4,10 @@
   display: block;
 }
 
+:host([orientation='vertical']) {
+  height: 100%;
+}
+
 :host([negative]) {
   --sbb-divider-color: var(--sbb-divider-color-negative);
 }


src/elements/divider/divider.scss
```diff
@@ -1,27 +1,24 @@
...
 
 :host {
-  --sbb-divider-color: var(--sbb-background-color-4);
-  --sbb-divider-border-width: var(--sbb-border-width-1x);
-
   display: block;
-}
-
-:host([orientation='vertical']) {
   height: 100%;
```
TomMenga: In the previous implementation, the height was 100% only when `vertical`.
I'm not sure, but this might create a problem if you use a horizontal divider in a column flex container

TomMenga (changes requested):

src/elements/divider/divider.scss
jeripeierSBB: yes absolutely, I was just about to revert the changes :) now I re-introduced the additional div, as it seems necessary for our use cases we have...

PR #4441, opened 2026-02-02
refactor(sbb-autocomplete, sbb-select): css refactoring

TomMenga:
Part of #3969

### DIFF
--- a/src/elements/core/styles/core.scss
+++ b/src/elements/core/styles/core.scss
@@ -69,6 +69,18 @@ $theme: 'standard' !default;
 @use '../../container/sticky-bar/sticky-bar.global' as sticky-bar with (
   $theme: $theme
 );
+@use '../../dialog/dialog/dialog.global' as dialog with (
+  $theme: $theme
+);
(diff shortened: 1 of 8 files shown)


src/elements/option/option/option.scss
```diff
@@ -82,8 +66,8 @@
 }
 
 .sbb-option {
-  @include sbb.text-s--regular;
-
+  font-size: var(--sbb-option-font-size);
```
jeripeierSBB: Would it be possible to set the font-size on the host?

.storybook/preview.ts
```diff
@@ -9,6 +9,7 @@ import type { Parameters, StoryContext } from 'storybook/internal/types';
 import { makeDecorator } from 'storybook/preview-api';
 
 import '../src/elements/core/styles/standard-theme.scss';
+import '../src/elements-experimental/core/styles/standard-theme.scss';
 
 import leanTheme from '../src/elements/core/styles/lean-theme.scss?inline';
```
jeripeierSBB: All the following stuff should also be configured for the experimental styles.

PR #4445, opened 2026-02-02
fix(sbb-toggle): allow Space to toggle state

kyubisation:
According to the [W3C Switch Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/switch/), Space should toggle the state.

### DIFF
--- a/src/elements/toggle/toggle/toggle.component.ts
+++ b/src/elements/toggle/toggle/toggle.component.ts
@@ -8,7 +8,7 @@ import {
 } from 'lit';
 import { customElement, property } from 'lit/decorators.js';
 
-import { getNextElementIndex, interactivityChecker, isArrowKeyPressed } from '../../core/a11y.ts';
+import { interactivityChecker } from '../../core/a11y.ts';
 import { forceType } from '../../core/decorators.ts';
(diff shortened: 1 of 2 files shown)


src/elements/toggle/toggle/toggle.component.ts
```diff
@@ -228,20 +228,29 @@ class SbbToggleElement<T = string> extends SbbDisabledMixin(
...
-        );
-      }
+    const options = this.options;
+    const isRtl = this.matches(':dir(rtl)');
+    const currentIndex = options.findIndex((option) => option.checked) ?? options[0];
+    const availableOptions = options
+      .slice(currentIndex + 1)
+      .concat(options.slice(0, currentIndex))
+      .filter((o) => !o.disabled && interactivityChecker.isVisible(o));
+    if (!isRtl ? evt.key === ' ' || evt.key === 'ArrowRight' : evt.key === 'ArrowLeft') {
```
jeripeierSBB: I don't understand why in this context the direction matters? Is this kind of future proof if more than two options are supported?

jeripeierSBB (approved):
LGTM

PR #4450, opened 2026-02-03
refactor(sbb-carousel): css refactoring

DavideMininni-Fincons:
Subtask of https://github.com/sbb-design-systems/lyne-components/issues/3969

### DIFF
--- a/src/elements/carousel/carousel/carousel.visual.spec.ts
+++ b/src/elements/carousel/carousel/carousel.visual.spec.ts
@@ -98,9 +98,9 @@ describe('sbb-carousel', () => {
 
             setup.withPostSetupAction(async () => {
               await Promise.all(
-                Array.from(setup.snapshotElement.querySelectorAll<HTMLImageElement>('img')).map(
-                  (el) => waitForImageReady(el),
-                ),
+                Array.from(
+                  setup.snapshotElement.querySelectorAll<HTMLImageElement>('sbb-image'),
+                ).map((el) => waitForImageReady(el)),
               );
               setup.snapshotElement
                 .querySelectorAll('sbb-carousel-item')[0]!


src/elements/carousel/carousel/carousel.scss
TomMenga: We could create a var also for the box-shadow

TomMenga (approved):
LGTM 👍

src/elements/carousel/carousel-item/carousel-item.scss
```diff
@@ -5,3 +5,7 @@
   scroll-snap-align: start;
   flex-shrink: 0;
 }
+
+::slotted(sbb-image) {
```
jeripeierSBB: Sorry again :), I think we should just adapt the example with the sbb-image-border-radius-none class. But is maybe also worth a discussion with @kyubisation. what do you think? general solution or example adaption?

jeripeierSBB: maybe add a visual test, if there is not already one?

src/elements/carousel/carousel/carousel.visual.spec.ts
```diff
@@ -8,44 +8,109 @@ import './carousel.component.ts';
...
+              {
+                backgroundColor: shadow
+                  ? 'var(--sbb-background-color-3)'
+                  : 'var(--sbb-background-color-1)',
+              },
+            );
+
+            setup.withPostSetupAction(async () => {
+              await Promise.all(
+                Array.from(setup.snapshotElement.querySelectorAll<HTMLImageElement>('img')).map(
```
jeripeierSBB:
```suggestion
                Array.from(setup.snapshotElement.querySelectorAll<HTMLImageElement>('sbb-image')).map(
```

PR #4452, opened 2026-02-04
build: add CSP for image CDN

kyubisation:
Currently images from the SBB image CDN are blocked due to CSP. This PR adds the necessary config to allow these.


jeripeierSBB (approved):
LGTM

PR #4456, opened 2026-02-04
refactor(sbb-button): css refactoring

DavideMininni-Fincons:
Subtask of #3969


jeripeierSBB (changes requested):
Looks good for the current state, but I wonder whether also primary-button.scss, accent-button.scss, ... should be migrated?

PR #4462, opened 2026-02-05
refactor(sbb-radio-button): expose gap variable

jeripeierSBB:
In customer projects it could happen that that using a calculation with the gap is needed. Having a CSS variable helps here (aligned to how it's done in the sbb-checkbox)


kyubisation (approved):
LGTM 👍

PR #4465, opened 2026-02-06
refactor(sbb-chip-label): css refactoring

DavideMininni-Fincons:
Subtask of #3969


src/elements/core/styles/mixins/image.scss
```diff
@@ -26,7 +26,6 @@
 @mixin figure-overlap-base {
   display: flex;
   gap: var(--sbb-spacing-fixed-1x);
-  flex-wrap: wrap;
```
jeripeierSBB: This was a recent fix of me. where does it make problems?

jeripeierSBB (changes requested):

PR #4493, opened 2026-02-12
refactor(sbb-container, sbb-sticky-bar): css refactoring

DavideMininni-Fincons:
Subtask of #3969.

### DIFF
--- a/src/elements/container/sticky-bar/sticky-bar.global.scss
+++ b/src/elements/container/sticky-bar/sticky-bar.global.scss
@@ -14,5 +14,4 @@ $theme: 'standard' !default;
     )};
   --sbb-sticky-bar-position: sticky;
   --sbb-sticky-bar-slide-vertically-animation-easing: ease-out;
-  --sbb-sticky-bar-slide-vertically-animation-delay: 0s;
 }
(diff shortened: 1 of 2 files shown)


src/elements/container/sticky-bar/sticky-bar.global.scss
```diff
@@ -0,0 +1,19 @@
...
+      (
+        'standard': var(--sbb-spacing-responsive-xs),
+        'lean': var(--sbb-spacing-responsive-xxxs),
+      ),
+      $theme
+    )};
+  --sbb-sticky-bar-position: sticky;
+  --sbb-sticky-bar-slide-vertically-animation-easing: ease-out;
+  --sbb-sticky-bar-slide-vertically-animation-delay: 0s;
+  --sbb-sticky-bar-slide-vertically-animation-name: unset;
```
TomMenga:
```suggestion
```
You can probably remove these and use the default where they're referenced

TomMenga (approved):
LGMT 👍  just a little improvement

src/elements/container/sticky-bar/sticky-bar.global.scss
DavideMininni-Fincons: could you check 6bfee4d15e56159285b6c19b4e30915e2da94c8f ? i set the fallback for the animation name, but the delay is not reassigned, so i keep it as it is

PR #4510, opened 2026-02-18
fix(sbb-radio-button-group): handle init with falsy values

TomMenga:
This PR Closes [#289](https://github.com/sbb-design-systems/lyne-angular/issues/289)


jeripeierSBB (approved):
LGTM, thank you for the quick fix!

PR #4587, opened 2026-03-05
refactor(header): css refactoring

DavideMininni-Fincons:
Subtask of #3969


src/elements/header/common/header-action.global.scss
```diff
@@ -0,0 +1,28 @@
...
+
+@mixin base {
+  --sbb-header-action-color: var(--sbb-color-1);
+  --sbb-header-action-background-color: transparent;
+  --sbb-header-action-min-height: var(--sbb-size-element-s);
+  --sbb-header-action-padding-inline: var(--sbb-spacing-fixed-5x);
+  --sbb-header-action-transition-easing: var(--sbb-animation-easing);
+  --sbb-header-action-border-color: transparent;
+  --sbb-header-action-background-inset: 0;
+  --sbb-header-action-translate-y: 0;
```
jeripeierSBB: could it be used just as a fallback for the variable? or write the value here which is then applied on hover

src/elements/header/common/header-action.global.scss
```diff
@@ -0,0 +1,28 @@
...
+  --sbb-header-action-gap: var(--sbb-spacing-fixed-2x);
+  --sbb-header-action-icon-dimension: var(--sbb-size-icon-ui-small);
+  --sbb-header-action-active-border-width: var(--sbb-border-width-2x);
+  --sbb-header-action-active-border-color: var(--sbb-header-action-color);
+  --sbb-header-action-active-border-scale: 0;
```
jeripeierSBB: could it be used just as a fallback for the variable? or write the value here which is then applied with .sbb-active class

jeripeierSBB (approved):
LGTM, thank you

PR #4589, opened 2026-03-06
refactor(sbb-image): css refactoring

DavideMininni-Fincons:
Subtask of #3969


TomMenga (changes requested):
We have `styles/image.scss`, which contains the global rules for images. We could move those in the `image.global.scss` in the `@mixin rules`. What do you think? @DavideMininni-Fincons @jeripeierSBB

jeripeierSBB (changes requested):

PR #4644, opened 2026-03-16
refactor(sbb-notification): css refactoring

DavideMininni-Fincons:
Subtask of #3969


src/elements/notification/notification.component.ts
```diff
@@ -212,8 +212,11 @@ export class SbbNotificationElement extends SbbIconNameMixin(SbbReadonlyMixin(Sb
   private _configureTitle(): void {
     const title = this.querySelector?.<SbbTitleElement>('sbb-title');
     if (title) {
+      this.internals.states.add('slotted-title');
```
jeripeierSBB: this shouldn't be needed. the new SbbElement always creates states whether there is something slotted or not

jeripeierSBB (approved):
LGTM, the manual control of slotted state should not be needed

PR #4645, opened 2026-03-16
fix(sbb-teaser): avoid gap if no image slotted

jeripeierSBB:


jeripeierSBB:
> Is this a valid case? If the use of the teaser without a slotted image is intended does it need a story? If not does it need a fix or is to be considered broken anyway?

It's a very rare case, brought by @mcilurzo. It is intentional to have no story because normally it doesn't make sense. But there is a visual test for it.

PR #4734, opened 2026-04-01
fix: avoid bleeding of font-weight

jeripeierSBB:
A couple of components are expected to be used inside a bold context. By setting font-weight: normal, we avoid displaying it in bold.


kyubisation (approved):
LGTM

PR #4777, opened 2026-04-10
fix(dialog): tests broken on Safari

DavideMininni-Fincons:
On dialog opening, the internal `sbb-screenreader-only` content is set based on the content of the `sbb-dialog-title`
for some unknown reason, in WebKit only the `innerText` is an empty string, so it has been changed to `textContent`.


kyubisation (approved):
LGTM 👍 
Thanks!

PR #4779, opened 2026-04-10
feat(teaser-panel): component implementation

TomMenga:
This PR Closes #4768

The panel size in the `sbb-teaser-hero` wasn't following the Figma specifications for the large viewport. Now it's correct


src/elements/teaser-panel/teaser-panel.visual.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+import '../teaser-panel.ts';
+
+describe(`sbb-teaser-panel`, () => {
+  describeViewports(() => {
+    for (const state of [visualDiffDefault, visualDiffHover]) {
+      it(
+        state.name,
+        visualDiffDefault.with(async (setup) => {
+          await setup.withFixture(html`
+            <div style="height: 550px; position: relative; background-color: cyan">
```
jeripeierSBB: I suggest to use the png with the red and blue color (sorry name missing in my head) as the hover to non hover difference would better be visible.

src/elements/teaser-panel/teaser-panel.stories.ts
```diff
@@ -0,0 +1,27 @@
+import type { Meta, StoryObj } from '@storybook/web-components-vite';
+import type { TemplateResult } from 'lit';
+import { html } from 'lit';
+
+import readme from './readme.md?raw';
+import '../teaser-panel.ts';
+
+const Template = (): TemplateResult => html`
+  <div style="height: 550px; position: relative; background-color: cyan">
```
jeripeierSBB: instead of cyan I recommend to use an sbb color or the one we define in developer style mixin (it's kind of a purple)

jeripeierSBB (commented):
Looks good!

focus visible styles and text-decoration: none should be re-added

jeripeierSBB (approved):
LGTM, thank you!

PR #4780, opened 2026-04-10
fix: missing theme in visual-regression-app

DavideMininni-Fincons:
Commit #4707 broke the visual-regression app.
Directly importing the core style seems to fix it (not sure it's the right approach)

### DIFF
--- a/src/visual-regression-app/src/main.ts
+++ b/src/visual-regression-app/src/main.ts
@@ -3,7 +3,7 @@ import { html, LitElement, type TemplateResult } from 'lit';
 import { customElement } from 'lit/decorators.js';
 
 import '@sbb-esta/lyne-elements/core.js';
-import '@sbb-esta/lyne-elements/core/styles/core.scss';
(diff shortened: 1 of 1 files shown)


src/visual-regression-app/src/main.ts
```diff
@@ -3,6 +3,7 @@ import { html, LitElement, type TemplateResult } from 'lit';
 import { customElement } from 'lit/decorators.js';
 
 import '@sbb-esta/lyne-elements/core.js';
+import '@sbb-esta/lyne-elements/core/styles/core.scss';
```
jeripeierSBB: Does standard-theme not work?
```suggestion
import '@sbb-esta/lyne-elements/core/styles/standard-theme.scss';
```

DavideMininni-Fincons: it does on my local, but on the deployed version it does not work; i'm trying to relaunch the vr action

jeripeierSBB (approved):
LGTM, thank you

PR #4785, opened 2026-04-13
feat(form-field): add optional type to custom control API

kyubisation:
This PR adds the optional `type?: 'select' | string` property to the `SbbFormFieldElementControl` interface.
This enables setting `type` to `select`, which will display the drop down icon in the form field.

Closes #4773


jeripeierSBB (approved):
LGTM, thank you! Do you think we need to adapt also the readme with a dedicated section?

PR #4788, opened 2026-04-14
docs: add missing documentation for input and change events

### LINKED ISSUE #4786: Document input and change events for our form elements
We should document the input and change events for our form elements.
For each form element that we have, add class body annotations for these two events.
The description can be adapted from the existing documentation (where we have it) or from MDN.

DavideMininni-Fincons:


jeripeierSBB (approved):
LGTM, thanks

TomMenga (approved):
LGTM 👍

PR #4793, opened 2026-04-15
fix(autocomplete): reposition correctly on dynamically adding options

### LINKED ISSUE #350: [Bug]: Autocomplete with lazy options hides input
### Preflight Checklist

### Bug type

Functionality

Components

    "@sbb-esta/lyne-elements": "4.10.0",     "@sbb-esta/lyne-angular": "21.6.0",

### Operating system

win11

### Browser / Browser version

chrome 

### Input mode

None

### Additional settings

_No response_

### What happened?

When the options of an autocomplete change while it is open the panel is not correctly repositioned.
When initially the panel with the options fit below and then additional options are added, so that the panel does not fit any more below, it is not positioned above, but somewhere inbetween. This hides the input where the user might be typing, which is bad.

1. go to https://stackblitz.com/edit/1qhcbuyp?file=src%2Fapp%2Fautocomplete-basic-example.html
2. focus the input
3. wait 1 second
4. observe that the panel is positioned over the input, effectively hiding the input

Image

### Relevant log output

```shell

```

kyubisation:
Closes sbb-design-systems/lyne-angular#350

### DIFF
--- a/src/elements/autocomplete/autocomplete-base-element.ts
+++ b/src/elements/autocomplete/autocomplete-base-element.ts
@@ -143,14 +143,18 @@ export abstract class SbbAutocompleteBaseElement<T = string> extends SbbNegative
   protected abstract panelRole: string;
   protected activeOption: SbbOptionBaseElement<T> | null = null;
   protected pendingAutoSelectedOption: SbbOptionBaseElement<T> | null = null;
-  private _originResizeObserver = new ResizeController(this, {
+  private _resizeObserver = new ResizeController(this, {
     target: null,
     skipInitial: true,
-    callback: () => {
-      if (this.state === 'opened') {
-        this._setOverlayPosition();
-      }
-    },
+    // This is an IIFE, because we need to keep track of the timeout state
+    // for debouncing the resize callbacks.
+    callback: (() => {
+      let timeoutId: ReturnType<typeof setTimeout>;
+      return () => {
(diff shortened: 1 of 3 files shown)


src/elements/autocomplete/autocomplete-base-element.ts
```diff
@@ -395,6 +400,13 @@ export abstract class SbbAutocompleteBaseElement<T = string> extends SbbNegative
   private _handleSlotchange(): void {
     this._highlightOptions(this.triggerElement?.value);
 
+    // It is possible that an option instance is added that has not been rendered
+    // yet and therefore has height 0. Due to this we also observe the size of the options.
+    const currentElements = Array.from(this.querySelectorAll('*'));
+    this._previousElements?.forEach((option) => this._resizeObserver.unobserve(option));
```
jeripeierSBB: "option" maybe not the most accurate name anymore

src/elements/autocomplete/autocomplete.stories.ts
```diff
@@ -557,6 +557,19 @@ export const WithEllipsis: StoryObj = {
   args: { ...defaultArgs, ellipsis: true },
 };
 
+export const Repositiontest: StoryObj = {
```
jeripeierSBB: I think we shouldn't add this story?

jeripeierSBB (approved):
LGTM, two comments

PR #4795, opened 2026-04-15
refactor(tabs): replace throttle with debounce mechanism

kyubisation:
The current throttle implementation might fail if called again within the throttle timer. Due to this, we replace throttle with a local debounce mechanism and deprecate the throttle function.


jeripeierSBB (approved):
LGTM, thanks :)

PR #4799, opened 2026-04-16
chore: optimize imports

jeripeierSBB:
Since the refactoring of the new modules, some imports could be grouped. This PR does group imports where possible and also sorts alphabetically. Example for grouping: https://github.com/sbb-design-systems/lyne-components/pull/4799/changes#diff-2953218cd10ed973f30192362b51e1959b81fd2760a3cb23f3b948bcc15444b3


kyubisation (approved):
LGTM 👍

PR #4801, opened 2026-04-16
fix: ensure slot connection for children

jeripeierSBB:
If you assign a slot attribute without a corresponding slot element, it is not fully part of the DOM and e.g. media queries will fail. This is currently the case with the named slot list mixin and we changed the lifecycle to ensure that it is always correctly connected.


kyubisation (approved):
LGTM 👍

PR #4802, opened 2026-04-16
fix(form-field): layout for Angular custom components in size=s

### LINKED ISSUE #353: [Bug]: custom SbbFormFieldControl layout broken
### Preflight Checklist

### Bug type

Other

Design Tokens

21.6.1 / 4.11.0

### Operating system

win11 

### Browser / Browser version

chrome

### Input mode

None

### Additional settings

_No response_

### What happened?

When using custom input components (following docs: https://lyne-angular.app.sbb.ch/angular/components/form-field/overview#angular-4 ), then their height inside sbb-form-field is not correct:

Image

1. https://stackblitz.com/edit/yuueayq8?file=src%2Fapp%2Fform-field-native-select-example.html

### Relevant log output

```shell

```

DavideMininni-Fincons:
In the sbb-form-field, the input's containers in the shadowDom are all in display flex, so even if the slotted form (input, select,..) is naturally is inline-block, it become block, as specified in https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Flexible_box_layout/Basic_concepts#the_flex_container
So, setting display: block would suffice to solve the height issue; however, the date-input and the time-input are in display flex, so block has been changed to flex (**NOTE TO REVIEWER**: split the behavior?). 
This fixes the issue because in an angular custom component as in the provided example, the rule is applied to any slotted input, tested with, e.g.
```
@Component({
  selector: 'sbb-custom-control-textarea',
  template: `
    <div>
      <div>
        <textarea id="textarea" rows="3"></textarea>
      </div>
    </div>`,
    ...
})
export class CustomControlComponentTextarea { }
```

The same idea applies with the negative margin, which adapts a similar rule present in the form-field.scss (the shown rule is applied on the input container)
```
.sbb-form-field__input {
  ...
  :host(
      [size='s']:is(
        :state(input-type-input),
        :state(input-type-select),
        :state(input-type-sbb-select),
        :state(input-type-sbb-date-input),
        :state(input-type-sbb-time-input)
      )
    )
    & {
    // In size s, the natural height of the text input is too small.
    // To not reserve too much space, we decrease the height.
    margin-block-end: #{sbb.px-to-rem-build(-2)};
  }
}
```

### DIFF
--- a/src/elements/form-field/form-field/form-field.visual.spec.ts
+++ b/src/elements/form-field/form-field/form-field.visual.spec.ts
@@ -15,7 +15,7 @@ import '../../button.ts';
 
 import '../../popover.ts';
 
-@customElement('custom-control-element-input')
+@customElement('custom-input-element')
 class SbbCustomControlElementInput extends SbbElement {
   protected override render(): TemplateResult {
     return html`
@@ -23,12 +23,6 @@ class SbbCustomControlElementInput extends SbbElement {
         <slot></slot>
       </div>
     `;
-  }
-}
-@customElement('custom-control-element-select')
-class SbbCustomControlElementSelect extends SbbElement {
-  protected override render(): TemplateResult {
-    return html` <slot></slot> `;
   }
 }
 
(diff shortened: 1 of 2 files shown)


jeripeierSBB (approved):
LGTM, thank you! Could you add a visual test with the situation in the Repro?

src/elements/form-field/form-field/form-field.visual.spec.ts
```diff
@@ -7,13 +8,30 @@ import {
...
 } from '../../core/testing/private.ts';
-import { isChromium } from '../../core.ts';
+import { isChromium, SbbElement } from '../../core.ts';
 
 import '../../form-field.ts';
 import '../../button.ts';
 
 import '../../popover.ts';
 
+@customElement('custom-control-element-input')
```
jeripeierSBB: I think you don't need the extra step with a custom element. some nested divs should be enough?

jeripeierSBB (approved):

src/elements/form-field/form-field/form-field.visual.spec.ts
DavideMininni-Fincons: I'm keeping a single one just to show that works also with it (until forms are in the light dom the style is applied)

jeripeierSBB (approved):
LGTM, thank you :)

PR #4811, opened 2026-04-21
refactor: improve support for scrollbar styles in shadow DOM

jeripeierSBB:
Currently, when using some of the components where the `sbb-scrollbar` CSS class is added to the host, the scrollbar styles do not work at all when using such a component inside a shadow DOM. With this PR, the scrollbar styles are also added to the host. However, unfortunately, the hover effect of the scrollbar thumb is not working when applying the scrollbar styles on the host. Due to this fact we keep both solutions: styles on shadow DOM but also additionally the CSS class on the host. With this solutions we only miss the hover effect if a consumer uses a component in the shadow DOM.

### DIFF
--- a/src/elements/dialog/dialog-content/dialog-content.global.scss
+++ b/src/elements/dialog/dialog-content/dialog-content.global.scss
@@ -4,13 +4,14 @@ $theme: 'standard' !default;
 
 @mixin rules {
   sbb-dialog-content {
-    // If there is a title, remove top spacing for the content
+    // If there is a title, reduce the top spacing for the content
     sbb-dialog:has(> sbb-dialog-title) > &,
     sbb-dialog:state(has-intermediate-element):has(> * > sbb-dialog-title) > * > & {
       padding-block-start: var(--sbb-spacing-fixed-1x);
     }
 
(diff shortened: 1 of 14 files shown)


src/elements/dialog/dialog-content/dialog-content.global.scss
```diff
@@ -4,13 +4,14 @@ $theme: 'standard' !default;
...
   sbb-dialog-content {
-    // If there is a title, remove top spacing for the content
+    // If there is a title, reduce the top spacing for the content
     sbb-dialog:has(> sbb-dialog-title) > &,
     sbb-dialog:state(has-intermediate-element):has(> * > sbb-dialog-title) > * > & {
       padding-block-start: var(--sbb-spacing-fixed-1x);
     }
 
     // Set the negative scrollbar colors for dialog content when dialog is negative
+    // This is additionally ended in Light DOM to be more specific in overriding the variables.
```
jeripeierSBB:
```suggestion
    // This is additionally added to the Light DOM to be more specific in overriding the variables.
```

kyubisation (approved):
LGTM 👍

PR #4814, opened 2026-04-21
fix(select): force update value

### LINKED ISSUE #347: [Bug]: sbb-option does not work with [value] in dialog
### Preflight Checklist

### Bug type

Functionality

Components

4.10.0 / 21.6.0

### Operating system

_No response_

### Browser / Browser version

_No response_

### Input mode

None

### Additional settings

_No response_

### What happened?

When using `[value]` in `sbb-option` the initially set value in the form control is not (visually) selected in the input. This happens when the component is opened in a dialog.

Note: when using `value` this works fine!

1. open https://stackblitz.com/edit/majcvwag?file=src%2Fapp%2Fselect-basic-example.html
2. click "open dialog"
3. observe that the select has no value selected - even though the form value matches the `[value]`

### Relevant log output

```shell

```

DavideMininni-Fincons:
Proposal for resolution of bug [#347](https://github.com/sbb-design-systems/lyne-angular/issues/347)

In Angular, when the option value is set via binding, the 'value' attribute is undefined when the select does the checks on its value and tries to set it based on the provided options.

The idea here is to provide a listener for an internal event, as done for `ɵoptgroupslotchange`; this PR should be completed with another one in the Angular repo which basically triggers the event on the sbb-option after the value is set, something like: 
```
 /**
   * Value of the option.
   */
  @Input()
  public set value(value: T) {
    this.#ngZone.runOutsideAngular(() => {
      this.#element.nativeElement.value = value;
      this.#element.nativeElement.dispatchEvent(new Event('ɵoptionvaluechange'));
    });
  }
  public get value(): T {
    return this.#element.nativeElement.value;
  }
```

### DIFF
--- a/src/elements/select/select.component.ts
+++ b/src/elements/select/select.component.ts
@@ -221,6 +221,20 @@ export class SbbSelectElement<T = string> extends SbbUpdateSchedulerMixin(
       }),
     );
 
+    // If any option value changes, recheck it against the select's value.
+    this.addController(
+      new MutationController(this, {
+        config: { attributeFilter: ['value'], subtree: true },
+        callback: (record) => {
+          record.forEach((e) => {
+            if ((e.target as HTMLElement).localName === 'sbb-option') {
+              this._updateValueOptionState();
+            }
+          });
+        },
+      }),
+    );
+
     this.addController(
       new SbbPropertyWatcherController(this, () => this.closest('sbb-form-field'), {
         negative: (e) => {
--- a/src/elements/select/select.component.ts
+++ b/src/elements/select/select.component.ts
@@ -194,6 +194,10 @@ export class SbbSelectElement<T = string> extends SbbUpdateSchedulerMixin(
     super();
     this.addEventListener?.('optionselectionchange', (e: Event) => this._onOptionChanged(e));
     this.addEventListener?.('optionLabelChanged', (e: Event) => this._onOptionLabelChanged(e));
+    /** Forces the sbb-select to update his value. */
+    this.addEventListener?.('ɵoptionvaluechange', () => this._updateValueOptionState(), {
+      capture: true,
+    });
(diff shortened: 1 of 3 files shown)


jeripeierSBB (commented):
is it possible to somehow create a test with the specific timing? e.g. by manually creating and adding to DOM, or whatever is needed.

DavideMininni-Fincons:
After some tests, I found out that the issue is not strictly an Angular one, since the logic of 'late update' of the option's value is not present at all.
So I removed the listener to the private event I created and I added a MutationObserver.

src/elements/select/select.component.ts
```diff
@@ -217,6 +217,20 @@ export class SbbSelectElement<T = string> extends SbbUpdateSchedulerMixin(
       }),
     );
 
+    // If any option value changes, recheck it against the select's value.
+    this.addController(
```
jeripeierSBB: shouldn't we better introduce an event and synchronously handle it? with a mutation observer there is some time gap. currently, the option has the following code. maybe an event could be triggered when the setter has set it's value? Or are there other problems with that idea?

```ts
  /**
   * Value of the option.
   *
   * @description Developer note: In this case updating the attribute must be synchronous.
   * Due to this, it is implemented as a getter/setter and the attributeChangedCallback() handles the diff check.
   */
  @property()
  public set value(value: T) {
    if (typeof value === 'string') {
      this.setAttribute('value', `${value}`);
      this._value = null;
    } else {
      this._value = value;
    }
  }
  public get value(): T {
    return (this._value ?? this.getAttribute('value')) as T;
  }
  private _value: T | null = null;
```

DavideMininni-Fincons: i was thinking about dispatching it only in angular, but yes, we can trigger it also in the base component; i updated the PR

jeripeierSBB (approved):
LGTM, thank you!

PR #4817, opened 2026-04-22
fix(autocomplete): prevent opening on untrusted input events

### LINKED ISSUE #4199: [Bug]: sbb-autocomplete sbb-option overlay does not close & selected option is not taken over
### Preflight Checklist

### Bug type

Accessibility

Components

3.12.1

### Operating system

macOS - Sequoia 15.6 (24G84)

### Browser / Browser version

Google Chrome Version 142.0.7444.162 (Official Build) (arm64)

### Input mode

Mouse/Keyboard

### Additional settings

_No response_

#### sbb-options overlay not closing
While using the sbb-autocomplete within the sbb-timetable-form the sbb-option overlay does not close when I switch from the "From" input field to the "To" input field by clicking into the "To" input field.
This issue only happens when I switch into the "To" input field by using the mouse. It does not happen if I tab into it. While analysing this issue with a developer from ESTA Web Design Systems, we noticed that the HTML element sbb-autocomplete__panel covers both the input fields. This sets `_isPointerDownEventOnMenu`to true and therefore does not recognise the switch to the "To" input field as a focusout event. That is why the sbb-option overlay does not close by itself.

Image

The relevant code for this particular problem is [index-1.html](https://github.com/user-attachments/files/23551870/index-1.html) and [main-1.js](https://github.com/user-attachments/files/23551723/main-1.js).

### selected sbb-option not being taken over in input field
To resolve this issue, I added an eventListener to input fields as such:
````
field.addEventListener("focusout", () => {
            autocomplete.close();
});
````

This fixed the abovementioned issue of the overlay not closing. However, it created a new issue. Now selecting an option via mouse click no longer set the text content of the input field to the value of the selected option. This only happened with the mouse click and the selection still worked if it was done via the keyboard. The developer that supported me was under the impression that there's a timing issue leading us to adapt the code to have a timeout as such:
````
field.addEventListener("focusout", () => {
            setTimeout(() => {
                    autocomplete.close();
             },
             200);
});
````

With these two workarounds we were able to fix the issues. As it is not desirable to have to do such workarounds to be able to use the timetable, I am creating this ticket so that the issues can be fixed at their root.

This is what the code looks like after all the workarounds:

[index.html](https://github.com/user-attachments/files/23551938/index.html)
[main.js](https://github.com/user-attachments/files/23551937/main.js)

In case you'd like to reproduce the issue I'll also provide you the csv file used for the stations:
[swiss-service-point-11062025.csv](https://github.com/user-attachments/files/23551957/swiss-service-point-11062025.csv)

**First issue - sbb-options overlay not closing** 
1. Enter "Bern" in "From" input field
2. Using the mouse click into the "To" input field
3. sbb-option overlay does not close

**Second issue - selected sbb-option not being taken over in input field** 
1. Add eventListener to input fields as described above (without timeout)
2. Type "Bern" into "From" input field
3. Using the mouse click the "Bern Bümpliz Süd" sbb-option
4. sbb-option overlay closes
5. "From" input field content is still just "Bern" instead of "Bern Bümpliz Süd"

### Relevant log output

```shell

```

DavideMininni-Fincons:
In the `sbb-timetable-form` the swap-button component swaps the from/to inputs and triggers change and input events.

However, if the input is connected to an autocomplete, the input event triggers the autocomplete's open method.

To prevent this, the listener on the autocomplete has been changed to react only to trusted events.

### DIFF
--- a/src/elements/autocomplete/autocomplete-base-element.ts
+++ b/src/elements/autocomplete/autocomplete-base-element.ts
@@ -495,7 +495,8 @@ export abstract class SbbAutocompleteBaseElement<T = string> extends SbbNegative
       (event) => {
         const value: string = (event.target as HTMLInputElement).value;
 
-        if (value) {
+        // Do not open if the event is triggered via dispatchEvent (e.g. click on timetable-swap-button)
+        if (value && event.isTrusted) {
           this.open();
         }
         this._highlightOptions(value);
--- a/src/elements/timetable-form/readme.md
+++ b/src/elements/timetable-form/readme.md
@@ -163,10 +163,9 @@ and dispatches the `input` and `change` events. The default `aria-label` value i
 
 #### Events
 
-| Name     | Type         | Description                                                                                                                                                                                                 | Inherited From |
-| -------- | ------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------- |
-| `change` | `Event`      | The change event is fired on the associated inputs when the user modifies the element's value. Unlike the input event, the change event is not necessarily fired for each alteration to an element's value. |                |
-| `input`  | `InputEvent` | The input event fires on the associated inputs when the value has been changed as a direct result of a user action.                                                                                         |                |
+| Name     | Type    | Description                                                                                                                                                                                                 | Inherited From |
+| -------- | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | -------------- |
+| `change` | `Event` | The change event is fired on the associated inputs when the user modifies the element's value. Unlike the input event, the change event is not necessarily fired for each alteration to an element's value. |                |
 
 #### CSS Properties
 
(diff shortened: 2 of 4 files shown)


kyubisation (commented):
I think this breaks Angular change detection: https://github.com/angular/angular/blob/main/packages/forms/src/directives/default_value_accessor.ts#L93C17-L96
Another option could be to add an event.isTrusted check in the autocomplete.
e.g.
```ts
        if (value && event.isTrusted) {
          this.open();
        }
```
@DavideMininni-Fincons What do you think?

DavideMininni-Fincons:
> I think this breaks Angular change detection: https://github.com/angular/angular/blob/main/packages/forms/src/directives/default_value_accessor.ts#L93C17-L96 Another option could be to add an event.isTrusted check in the autocomplete. e.g.
> 
> ```ts
>         if (value && event.isTrusted) {
>           this.open();
>         }
> ```
> 
> @DavideMininni-Fincons What do you think?

I didn't know about it, thanks! :) i'll update the PR

jeripeierSBB (approved):
LGTM, thank you!

PR #4818, opened 2026-04-22
style: set line-height to avoid overrides

jeripeierSBB:
In some circumstandes, consumer-set line-heights can win over component line-heights. Wherever it makes sense, line-height definition was added to stabilize the design.


kyubisation (approved):
LGTM 👍

PR #4825, opened 2026-04-23
refactor(button, action-group): change default size value from `l` to `m`

jeripeierSBB:
BREAKING CHANGE: The default size of the buttons has changed from `l` to `m` in standard theme. The `sbb-action-group` doesn't have a default `buttonSize` and `linkSize` applied anymore but delegates the default size choice to slotted button and links.


kyubisation (approved):
LGTM 👍 
Thanks!

PR #4826, opened 2026-04-23
chore(option): fix event bubbling

DavideMininni-Fincons:
Missing 'bubbles: true' on the option event, moved the setting of '_isValueManuallyAssigned' due to timing issues in tests.


jeripeierSBB (approved):
LGTM, thanks for fixing it

PR #4847, opened 2026-04-28
refactor(title): CSS refactoring

jeripeierSBB:
BREAKING CHANGE: the CSS variables `--sbb-title-text-color-normal` and `--sbb-title-text-color-normal-override` have been renamed to `--sbb-title-color`.

Note: With this change lean support is dropped, but will be re-introduce with lean-theme in a follow up PR.


TomMenga (approved):
Ok with the changes but there are some visual specs to be checked =)
lgtm 👍

PR #4849, opened 2026-04-28
refactor: minimize payload of sbb-screen-reader-only styles

jeripeierSBB:
The mixin of sbb-screen-reader-only was duplicated a lot. With this change, we provide a styles for the shadow DOM which also makes the sbb-screen-reader-only component obsolete.
Furthermore, styles are now applied in the mixins they belong to.

BREAKING CHANGE: `<sbb-screen-reader-only>` component has been removed. Use the CSS class `sbb-screen-reader-only` as a replacement.


TomMenga (approved):
LGTM 👍

PR #4852, opened 2026-04-29
fix: avoid transform `:state()` inside `CSS.supports()`

### LINKED ISSUE #4686: [Bug]: SbbSlotStateController throws with element-internals-polyfill when adding slotted custom states
### Preflight Checklist

### Bug type

Functionality

Components

4.7.0

### Operating system

Observed in production monitoring across end-user desktop systems

### Browser / Browser version

Predominantly Chrome Desktop according to NewRelic

### Input mode

Mouse/Keyboard

### Additional settings

Next.js SSR/static export setup using @lit-labs/nextjs together with element-internals-polyfill 3.0.2

### What happened?

We frequently see the following runtime error in production monitoring:

Error: Failed to execute 'add' on 'CustomStateSet': The specified value 'slotted' must start with '--'.

The stack resolves to SbbSlotStateController._updateSlottedState, where Lyne computes state names like slotted or slotted-name and passes them into internals.states.add(...).

In our runtime, element-internals-polyfill 3.0.2 is active. Its CustomStateSet implementation rejects bare state names and only accepts values starting with --. As soon as a component with slot state tracking processes a slot change, the add() call throws.

From a consumer perspective this makes Lyne components incompatible with this runtime setup and creates frequent production errors. Depending on the affected component, slot-dependent styling or behavior may also not update correctly after the exception.

### Expected behavior

Updating slot-related custom states should not throw.

A few possible acceptable outcomes from our perspective:
Lyne handles polyfilled CustomStateSet environments compatibly
Lyne gracefully skips or falls back when the active CustomStateSet implementation is incompatible
Lyne explicitly documents that this polyfill/runtime combination is unsupported
We also checked the current main branch source and SbbSlotStateController still appears to use bare state names for slot state updates.

Create a Next.js setup with Lit SSR integration and element-internals-polyfill 3.0.2
Render any Lyne component that relies on SbbSlotStateController and uses a default or named slot
Trigger the initial render or any slotchange
Observe the exception when internals.states.add(slotted) or internals.states.add(slotted-...) is called

### Relevant log output
[https://onenr.io/0oQD1NJ4Zwy](https://onenr.io/0oQD1NJ4Zwy)

```shell
Error: Failed to execute 'add' on 'CustomStateSet': The specified value 'slotted' must start with '--'
```

jeripeierSBB:
Closes #4686

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -380,9 +380,22 @@ function stateTransform(): PluginOption {
     transform(code: string, id: string) {
       if (/.(js|ts)$/.test(id)) {
         const ms = new MagicString(code);
-        ms.replaceAll(/:state\(([^)]+)\)/g, (_match, p1) => {
-          return `:is(:state(${p1}),[state--${p1}])`;
-        });
+        // Replace :state(...) everywhere except inside CSS.supports(...) call arguments,
+        // where :state() is used intentionally to feature-detect native browser support.
+        // All other occurrences (including in selector strings like matches(':state(x)'))
+        // are replaced so that the polyfill attribute selector is also emitted.
+
+        // Collect ranges of CSS.supports(...) calls to skip
+        const skipRanges = Array.from(code.matchAll(/CSS\.supports\([^)]*\)/g), (m) => [
+          m.index,
+          m.index + m[0].length,
+        ]);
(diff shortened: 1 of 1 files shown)


TomMenga (approved):
Nice find!
LGTM 👍

PR #4856, opened 2026-04-29
refactor: merge styles in SbbElement

### LINKED ISSUE #4647: Check if possible to merge styles from base class/mixin automatically
We currently aggregate the static styles property with all styles from the inheritance chain.
We want to check whether it is possible to automatically merge these styles.
Maybe Lit already supports this in a way.

jeripeierSBB:
With this change, every class, base class or mixin can just declare it's own styles and it is then merged together in the SbbElement. So no need of referencing super.styles anymore.

Closes #4647


TomMenga (approved):
LGTM 👍

PR #4858, opened 2026-04-29
fix(select): fix multiple checkbox styling

### LINKED ISSUE #4857: [Bug]:  sbb-select - multiple checkbox is wrongly styled on Safari
### Preflight Checklist

### Bug type

Functionality

Components

_No response_

### Operating system

_No response_

### Browser / Browser version

Safari

### Input mode

None

### Additional settings

_No response_

### What happened?

https://github.com/user-attachments/assets/8c521b1d-869e-4655-80fd-ac7bbb891df0

.

### Relevant log output

```shell

```

TomMenga:
This PR Closes #4857

Apparently, `internals.states.has('...')` works unreliably on the Safari browser

### DIFF
--- a/src/elements/container/sticky-bar/sticky-bar.component.ts
+++ b/src/elements/container/sticky-bar/sticky-bar.component.ts
@@ -158,7 +158,7 @@ export class SbbStickyBarElement extends SbbUpdateSchedulerMixin(SbbElement) {
     }
 
     this._state = 'sticking';
-    if (!this.internals.states.has('sticking') || this._isZeroAnimationDuration()) {
+    if (!this.matches?.(':state(sticking)') || this._isZeroAnimationDuration()) {
       this._stickyCallback();
     }
   }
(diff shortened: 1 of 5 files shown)


jeripeierSBB (approved):
Thank you for changing this. It's directly related to the bug I reported at webkit: https://bugs.webkit.org/show_bug.cgi?id=303467 
I don't fully remember whether  also `matches` is always reliable, but let's do it for now. An alternative would be to introduce local state copies (we did this at a some of places already), but I think this would be a bigger refactoring, so let's hope for this solution :) As a little improvement, we maybe should introduce a simple eslint rule (ask copilot :P), that marks usages of states.has(), so nobody is accidentally using it.

TomMenga:
> Thank you for changing this. It's directly related to the bug I reported at webkit: https://bugs.webkit.org/show_bug.cgi?id=303467 I don't fully remember whether also `matches` is always reliable, but let's do it for now. An alternative would be to introduce local state copies (we did this at a some of places already), but I think this would be a bigger refactoring, so let's hope for this solution :) As a little improvement, we maybe should introduce a simple eslint rule (ask copilot :P), that marks usages of states.has(), so nobody is accidentally using it.

From "empirical" testing, it seems that `matches` works 🤞

PR #4864, opened 2026-04-30
refactor: inline types from core

### LINKED ISSUE #4709: Inline types from core/interfaces where possible

TomMenga:
This PR Closes #4709. Full list of changes:
| Old value | New value |
|---|---|
| CalendarView | `'day' \| 'month' \| 'year'` |
| LinkTargetType | `'_blank' \| '_self' \| '_parent' \| '_top'` |
| PlaceState | `'FREE' \| 'ALLOCATED' \| 'RESTRICTED' \| 'SELECTED'`|
| PlaceType | `'SEAT' \| 'BICYCLE'` |
| SbbButtonSize | `'s' \| 'm' \| 'l'` |
| SbbButtonType | `'button' \| 'reset' \| 'submit'` |
| SbbCheckboxSize | `'xs' \| 's' \| 'm'` |
| SbbFlipCardImageAlignment | `'after' \| 'below' ` |
| SbbHorizontalFrom | `'zero' \| 'small' \| 'large' \| 'ultra'` |
| SbbIconPlacement  | `'start' \| 'end'` |
| SbbLanguage | `Record<string, string>` |
| SbbLinkSize  | `'xs' \| 's' \| 'm'`|
| SbbMiniButtonGroupSize | `'s' \| 'm' \| 'l' \| 'xl'` |
| SbbNavigationActionSize  | `'s' \| 'm' \| 'l'` |
| SbbOccupancy | `'high' \| 'medium' \| 'low' \| 'none'`| 
| SbbOrientation  | `'horizontal' \| 'vertical'` |
| SbbProtectiveRoom | `'none' \| 'minimal' \| 'ideal'` |
| SbbRadioButtonSize  | `'xs' \| 's' \| 'm'` |
| SbbSignetProtectiveRoom   | `'none' \| 'minimal' \| 'ideal' \| 'panel'` |
| SbbStatusType  | `'info' \| 'success' \| 'warning' \| 'error' \| 'pending' \| 'incomplete' \| 'not-started' \| 'in-progress'`                                                    |
| SbbTagSize  | `'s' \| 'm'` |
| SbbTime  | ``${number}:${number}:${number}`` |
| SbbTitleLevel  | `SbbHeadingLevel` |
| SbbToastPosition  | `'top-left' \| 'top-start' \| 'top-center' \| 'top-right' \| 'top-end' \| 'bottom-left' \| 'bottom-start' \| 'bottom-center' \| 'bottom-right' \| 'bottom-end'` |
| TravelDirection  | `'LEFT' \| 'RIGHT' \| 'NONE'` |

BREAKING CHANGE: Various TypeScript types have been removed. `SbbTitleLevel` was renamed to `SbbHeadingLevel` and moved to core module.


DavideMininni-Fincons (commented):
Generally looks good
I think types are not documented in storybook.. is this the only reason for the change, or is there something else?

TomMenga:
> Generally looks good I think types are not documented in storybook.. is this the only reason for the change, or is there something else?

That's the main reason, yes. Essentially, inlining them makes them more readable (both in the IDE and the readmes)

kyubisation (approved):
LGTM 👍 
Thanks!

kyubisation:
I do not have a strong opinion on this change; I think it slightly improves readability. The duplication does not feel like a problem to me, as most usages are very self contained and mostly don't change.
@TomMenga You mentioned you are not fully convinced. Can you elaborate?

TomMenga:
> I do not have a strong opinion on this change; I think it slightly improves readability. The duplication does not feel like a problem to me, as most usages are very self contained and mostly don't change. @TomMenga You mentioned you are not fully convinced. Can you elaborate?

Mainly, the doubt is on where we draw the line of when it's worth inlining them. As you mentioned, the cons are duplication and decoupling of linked properties.
Personally, I stopped when the type was used in multiple components. But maybe @jeripeierSBB had a bit of a stronger opinion on this

src/elements/core/i18n/i18n.ts
```diff
@@ -299,7 +299,7 @@ export const i18nSupersaver: Record<string, string> = {
   it: 'Biglietti risparmio',
 };
 
-export const i18nOccupancy: Record<SbbOccupancy, Record<SbbLanguage, string>> = {
+export const i18nOccupancy: Record<SbbOccupancy, Record<string, string>> = {
```
jeripeierSBB, line 302: I feel we should not just add string as we should point to the available languages, not?

src/elements/core/interfaces/types.ts
```diff
@@ -1,15 +1,5 @@
...
-
-export type SbbTime = `${number}:${number}:${number}`;
-
 export type SbbHorizontalFrom = 'zero' | 'small' | 'large' | 'ultra';
 
-export type SbbIconPlacement = 'start' | 'end';
-
 export type SbbOccupancy = 'high' | 'medium' | 'low' | 'none';
 
-export type SbbOrientation = 'horizontal' | 'vertical';
```
jeripeierSBB, line 11: Hmm, for the orientation I somehow feel that a global type would be worth. As an alternative, it's maybe worth to create a mixin? like we have for disabled or required. it's small, but it would align that there is a common term for having an orientation?

src/elements/datepicker/readme.md
```diff
@@ -99,7 +99,7 @@ todayStub.restore();
 | `input`   | `input`   | public  | `SbbDateInputElement<T> \| null` | `null`  | Reference to the sbb-date-input instance or the native input connected to the datepicker. For attribute usage, provide an id reference. |
 | `isOpen`  | -         | public  | `boolean`                        |         | Whether the element is open.                                                                                                            |
 | `trigger` | `trigger` | public  | `HTMLElement \| null`            | `null`  | The element that will trigger the popover overlay. For attribute usage, provide an id reference.                                        |
-| `view`    | `view`    | public  | `CalendarView`                   | `'day'` | The initial view of calendar which should be displayed on opening.                                                                      |
+| `view`    | `view`    | public  | `SbbCalendarElement['view']`     | `'day'` | The initial view of calendar which should be displayed on opening.                                                                      |
```
jeripeierSBB: maybe we could fix docs or manifest generation that it looks up types, especially if they are coming with an Sbb prefix

src/elements/flip-card/flip-card/flip-card.visual.spec.ts
```diff
@@ -19,6 +19,8 @@ import '../../image.ts';
 import '../../link.ts';
 import '../../title.ts';
 
+type SbbFlipCardImageAlignment = SbbFlipCardSummaryElement['imageAlignment'];
```
jeripeierSBB, line 22: Do you think this extra type has a value?

src/elements/tag/tag-group/tag-group.component.ts
```diff
@@ -56,7 +56,7 @@ export class SbbTagGroupElement<T = string> extends SbbDisabledMixin(
    * Tag group size, either s or m.
    * @default 'm' / 's' (lean)
    */
-  @property({ reflect: true }) public accessor size: SbbTagSize = isLean() ? 's' : 'm';
+  @property({ reflect: true }) public accessor size: 's' | 'm' = isLean() ? 's' : 'm';
```
jeripeierSBB: shouldn't we link to the tag size?
```suggestion
  @property({ reflect: true }) public accessor size: SbbTagElement['size'] = isLean() ? 's' : 'm';
```

jeripeierSBB (commented):
I like the refactoring. The only one I am in doubt is the SbbOrientation. maybe we could also create a mixin for that?

src/elements/core/i18n/i18n.ts
TomMenga: You're not wrong, but we never use the `SbbLanguage` type except for this case. I18n strings are all typed as `Record<string, string>` and the `SbbLanguageController` uses an array of supported languages

src/elements/core/interfaces/types.ts
TomMenga: I was a bit conflicted too. The leaning factor was that the orientation will probably never change. So duplicating `horizontal | vertical` won't be a problem

src/elements/flip-card/flip-card/flip-card.visual.spec.ts
TomMenga: Just for readability inside the spec file (it's not exported)

src/elements/tag/tag-group/tag-group.component.ts
TomMenga: Yes, nice catch!

src/elements/datepicker/readme.md
TomMenga: That would help, I'll see what we can do =)

src/elements/core/i18n/i18n.ts
jeripeierSBB: ah you are right, I did misunderstood something. yeah so if this was the only occurrence, we can absolutely remove it :)

src/elements/mini-calendar/mini-calendar-month/mini-calendar-month.component.ts
```diff
@@ -36,7 +36,7 @@ export class SbbMiniCalendarMonthElement<T = Date> extends SbbElement {
   private _monthNames = this._dateAdapter.getMonthNames('short');
   private _monthLabel: string | null = null;
   private _yearLabel: string | null = null;
-  private _previousOrientation: SbbOrientation | null = null;
+  private _previousOrientation: 'horizontal' | 'vertical' | null = null;
```
jeripeierSBB:
```suggestion
  private _previousOrientation: SbbMiniCalenderElement['orientation'] | null = null;
```

src/elements/train/train-formation-orientation-mixin.ts
```diff
@@ -22,7 +22,7 @@ export const SbbTrainFormationOrientationMixin = <T extends AbstractConstructor<
     extends superClass
     implements Partial<SbbTrainFormationOrientationMixinType>
   {
-    @state() protected accessor orientation: SbbOrientation | null = null;
+    @state() protected accessor orientation: 'horizontal' | 'vertical' | null = null;
```
jeripeierSBB:
```suggestion
    @state() protected accessor orientation: SbbTrainFormationOrientationMixinType['orientation'] = null;
```

src/elements/core/interfaces/types.ts
jeripeierSBB: we keep your refactoring :)

jeripeierSBB (approved):
LGTM. thank you!

PR #4888, opened 2026-05-06
feat(stepper): add `active` event to step element

kyubisation:
This PR adds a `active` event to `<sbb-step>`, which is dispatched on step activation.
This enables us to implement lazy loading in lyne-angular for steps.


TomMenga (approved):
LGTM 👍

jeripeierSBB (approved):
LGTM

PR #4907, opened 2026-05-12
style(table): table default is unstriped

### LINKED ISSUE #4898: story(table): table default should be non-striped
Currently, striped is the default. We should invert the default and adapt documentation.

TomMenga:
BREAKING CHANGE: the `sbb-table`default row styling has been changed from striped to unstriped. To achieve the striped look, use the CSS class `sbb-table--striped` on your table.

This PR Closes #4898


src/elements/table/table.global.scss
jeripeierSBB: Is the ampersand (&) needed here? in the striped variant it isn't. what's the reason we keep the unstriped here? nesting?

TomMenga: You're right, the & it's not needed.
About having the 'unstriped' class, it made sense when we had mixins because a consumer could choose to have the striped/unstriped as the default. They could still do it without the mixin but it's less plausible, I think. I don't have a strong opinion, we could leave it as a remote option

jeripeierSBB: for me it's ok to keep

jeripeierSBB (approved):
LGTM, thank you!

PR #4908, opened 2026-05-12
refactor(navigation): remove `size` property of buttons and links

jeripeierSBB:
The size of the navigation actions is always inherited from its group component.

BREAKING CHANGE: The `size` property of the `sbb-navigation-button`  and `sbb-navigation-link` has been removed.


kyubisation (approved):
LGTM 👍 
Thanks!

PR #4918, opened 2026-05-13
refactor: replace CustomEvents with specific events

### LINKED ISSUE #4899: Replace CustomEvent with our own Event classes
We currently use `CustomEvent` in some places. We would like to replace these usages with our own Event classes, extended from `Event`.
Additionally, in some places we already did that but still inherited from `CustomEvent`. This should be changed to inherit from `Event`.

This is a breaking change, as the resulting Event classes won't have a details property anymore.

DavideMininni-Fincons:
The type of the following events have been changed from `CustomEvents` to specific events which extend the native `Event`. The `detail` property has been removed and the properties are now available on the `event` object itself:
- `inputAutocomplete` on `sbb-autocomplete`;
- `dateselected` on the `sbb-calendar` / `sbb-datepicker`;
- `beforeshow` and `show` on the `sbb-carousel-item`;
- `filechanged` on the `sbb-file-selector` / `sbb-file-selector-dropzone`;
- `page` on the `sbb-paginator` / `sbb-compact-paginator`;
- `beforeclose` and `close` on the `sbb-popover` / `sbb-datepicker`;
- `validate` on the `sbb-step`;
- `tabchange` on the `sbb-tab-group`;
- `selectcoach` on the `sbb-seat-reservation-navigation-coach`;

Other changes:
- the `overlayOutsidePointer` is now typed as `Event`;
- the `SbbChipInputTokenEndEvent`, `SbbOverlayCloseEvent` and `SbbDialogCloseEvent` now extend `Event`;
- the `SbbChipInputTokenEndEventDetails` and `SbbOverlayCloseEventDetails` interfaces have been removed.

Closes #4899 

BREAKING CHANGE: The type of several events have been changed from `CustomEvents` to specific events which extend the native `Event` (see details in #4918).


src/elements/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -4,16 +4,37 @@ import { SbbElement } from '../../core.ts';
 
 import style from './carousel-item.scss?inline';
 
-export interface SbbCarouselItemEventDetail {
-  index: number;
+class SbbCarouselItemShowingEvent extends Event {
```
kyubisation: I would prefer to only have only one class (with name `SbbCarouselItemShowEvent`) for both show and beforeshow events.
Or what do you think?

src/elements/carousel/carousel-item/carousel-item.component.ts
```diff
@@ -4,16 +4,37 @@ import { SbbElement } from '../../core.ts';
...
+  private readonly _index: number;
+
+  public get index(): number {
+    return this._index;
+  }
+
+  protected constructor(type: string, index: number) {
```
kyubisation: Can we change the second parameter to an object? (Currently just `{ index }`).
In that case we can more easily add more properties, if required.

src/elements/file-selector/common/file-selector-common.ts
```diff
@@ -30,6 +30,19 @@ import {
 
 import style from './file-selector-common.scss?inline';
 
+export class SbbFileChangedEvent extends Event {
```
kyubisation: Nitpick; I would prefer `SbbFileChangeEvent`
I am aware that this slighly conflicts with the event name. Maybe let's discuss this.

src/elements/paginator/common/paginator-common.ts
```diff
@@ -36,6 +36,19 @@ export interface SbbPaginatorPageEventDetails {
   previousPageIndex: number;
 }
 
+export class SbbPaginatorPageEvent extends Event {
```
kyubisation, line 32: Can we inline the properties from the `SbbPaginatorPageEventDetails`?

src/elements/popover/popover.component.ts
```diff
@@ -45,6 +45,31 @@ const HORIZONTAL_OFFSET = 32;
 const popoversRef = new Set<SbbPopoverBaseElement>();
 const pointerCoarse = isServer ? false : matchMedia(SbbMediaQueryPointerCoarse).matches;
 
+class SbbPopoverClosingEvent extends Event {
```
kyubisation: Similar as before; Only one event `SbbPopoverCloseEvent`

src/elements/stepper/step/step.component.ts
```diff
@@ -114,22 +127,19 @@ export class SbbStepElement extends SbbElement {
   /**
    * Emits a validate event whenever step switch is triggered.
    * @internal
+   *
    * TODO: @breaking-change: make protected
+   *
+   * FIXME: the name of this variable appears as event name in the readme
```
kyubisation: I think we should move this comment into the method body, as it is not relevant for the consumer.

src/elements/tabs/tab-group/tab-group.component.ts
```diff
@@ -31,11 +31,24 @@ export interface SbbTabChangedEventDetails {
   previousTab: SbbTabElement | undefined;
 }
 
+export class SbbTabChangedEvent extends Event {
```
kyubisation: Nitpick; I would prefer the name `SbbTabChangeEvent`

kyubisation: Can we inline the `SbbTabChangedEventDetails` properties

kyubisation (dismissed):
Nice work!

src/elements/stepper/step/step.component.ts
```diff
@@ -25,6 +25,19 @@ export interface SbbStepValidateEventDetails {
   nextStep: SbbStepElement | null;
 }
 
+export class SbbStepValidateEvent extends Event {
+  private readonly _detail: Readonly<SbbStepValidateEventDetails>;
+
+  public get detail(): Readonly<SbbStepValidateEventDetails> {
```
jeripeierSBB: should we inline the properties?

jeripeierSBB (commented):
I think having the "detail()" getter / property is an anti-pattern for the new solution, as the new events can have the props inlined? some are already commented by @kyubisation, but there are also other (e.g. seat reservation). It can also be, that I oversee something ;)

jeripeierSBB (approved):
Great, thank you! LGTM

PR #4931, opened 2026-05-20
refactor(form-field): remove deprecated optional property

jeripeierSBB:
BREAKING CHANGE: The optional property of the `sbb-form-field` has been removed. The `(optional)` string should be managed by consumer inside the `<label>`.


src/elements/form-field/form-field.stories.ts
```diff
@@ -14,6 +14,13 @@ import '../link.ts';
 import '../popover.ts';
 import '../title.ts';
 
+const i18nOptional: Record<string, string> = {
+  de: '(optional)',
+  en: '(optional)',
+  fr: '(facultatif)',
+  it: '(facoltativo)',
+};
+
```
TomMenga: I think we never handled i18n in stories 🤔 
Maybe we can do without it?

TomMenga (approved):
LGTM 👍

src/elements/form-field/form-field.stories.ts
jeripeierSBB: Basically yes, but as we here have the translations also in the documentation I thought we maybe can also show it here, but on the other side, everything is English only. I will probably remove it but also ask @kyubisation  :D

DavideMininni-Fincons (approved):

PR #4936, opened 2026-05-20
refactor(option): remove automatic id assignment

TomMenga:
BREAKING CHANGE: `sbb-option` does not automatically assign an ID anymore


jeripeierSBB (approved):
LGTM, thank you

PR #4937, opened 2026-05-20
refactor(popover): extract close button from shadow DOM

jeripeierSBB:
BREAKING CHANGE: The properties `hide-close-button` and `accessibilityCloseLabel` of the `<sbb-popover>` have been removed. Slot the `<sbb-popover-close-button>` element whenever a close button is needed.


DavideMininni-Fincons (approved):
LGTM, thanks

src/elements/popover/popover-close-button/popover-close-button.component.ts
```diff
@@ -0,0 +1,44 @@
...
+export class SbbPopoverCloseButtonElement extends SbbSecondaryButtonElement {
+  public static override readonly elementName: string = 'sbb-popover-close-button';
+
+  private _languageController = new SbbLanguageController(this).withHandler(
+    () => (this.internals.ariaLabel = i18nClosePopover[this._languageController.current]),
+  );
+
+  public constructor() {
+    super();
+    this.size = 's' as this['size'];
```
TomMenga, line 21:
```suggestion
    this.size = 's';
```
This shouldn't be needed (I saw that the IDE interprets 's' as any) :/

TomMenga (approved):
LGTM 👍

src/elements/popover/popover-close-button/popover-close-button.component.ts
jeripeierSBB: it's for the custom element manifest. with that the correct default value is written. we have this pattern at various places, but maybe we could comment it :)

src/elements/core/eventing/composed-path-contains.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Check if there is an element in the composed path that has 'attribute'
+ *
+ * @param event The event object
+ * @param condition The condition callback to check for each element in the composed path. If it returns true, the element will be returned.
+ * @param host The boundary where the path will be cut, endElement excluded. If null or not present, the whole path will be used.
+ * @returns The element that has 'attribute', undefined otherwise.
+ */
+export function composedPathContains(
```
kyubisation: Not overly fond of this function, but it is fine
Inline alternative:
```ts
      event
        .composedPath()
        .filter((el, i, a) => el instanceof HTMLElement && i < a.indexOf(this))
        .some((el) => this._isCloseElement(el))
```

src/elements/popover/popover/popover.component.ts
```diff
@@ -628,18 +618,8 @@ export class SbbPopoverElement extends SbbPopoverBaseElement {
...
-        aria-label=${this.accessibilityCloseLabel || i18nClosePopover[this._language.current]}
-        size="s"
-        icon-name="cross-small"
-        sbb-popover-close
-      ></sbb-secondary-button>
-    `;
-
     return html`
-      ${!this.hideCloseButton && !this._hoverTrigger ? closeButton : nothing}
+      <slot name="close-button"></slot>
```
kyubisation: Does it need a slot or could it just be placed absolutely?

kyubisation (approved):
LGTM 👍 
Thank you!

src/elements/popover/popover/popover.component.ts
jeripeierSBB: I tried with a grid, but this wasn't possible. The requirement is that the focus lands on on the close button. this would require, that the consumer slots it first (maybe ok).

src/elements/core/eventing/composed-path-contains.ts
jeripeierSBB: Don't we lose performance again here? filter has to walk through all entries event if the current was reached and then a second iteration to find the element. Other than that I like one liners :)

kyubisation: Yes, from a performance perspective this is worse.
This is however not a hot path. A user click only happens sporadically.

src/elements/popover/popover/popover.component.ts
kyubisation: I would consider this user responsibility and document it accordingly.
But I also don't feel strongly about this.

kyubisation (approved):
LGTM 👍 
Thank you!

PR #4939, opened 2026-05-21
feat: expose module Sass mixins globally

### LINKED ISSUE #3969: CSS Refactoring

jeripeierSBB:
This PR exposes all global module files to our Sass api. With that, consumers could generate their custom them files with the possibility of fine granular customization. Additionally, there is now only one `_*.global.scss` file per module, including all global variables of contained components.

Closes #3969


src/elements-experimental/_index.scss
```diff
@@ -0,0 +1,3 @@
+// This file is auto-generated by scripts/generate-scss-api.ts
+@forward './autocomplete-grid/autocomplete-grid-row/autocomplete-grid-row.global' as
+  autocomplete-grid-row-*;
```
jeripeierSBB: module file

src/elements/breadcrumb/breadcrumb/breadcrumb.scss
```diff
@@ -18,7 +18,7 @@
 .sbb-breadcrumb {
   --sbb-link-text-decoration-line: none;
 
-  @include link.base;
+  @include link.base-rules;
```
jeripeierSBB, line 21: @kyubisation this is an exception, as we have element-rules already here :/

kyubisation (approved):
LGTM 👍

PR #4951, opened 2026-05-29
fix: ensure stable state for inputs while editing

kyubisation:
Currently the date and time input completely reset the value, if an assignment to valueAsDate is performed. This interrupts the typing if done during editing. This is currently an issue with the upcoming Angular Forms Signals, which continuously call writeValue and causes this problem.

Relates to sbb-design-systems/lyne-angular#316


jeripeierSBB (approved):
LGTM, thank you

PR #4958, opened 2026-06-02
refactor(header): rename action property from `expandFrom` to `hideLabelBelow` and unset default

jeripeierSBB:
BEGIN_COMMIT_OVERRIDE
refactor(header): rename action property from expandFrom to hideLabelBelow and unset default

BREAKING CHANGE: The `expandFrom` property of `<sbb-header-button>` and `<sbb-header-link>` has been renamed to `hideLabelBelow` and no longer has a default value. Previously, `expandFrom` defaulted to `large`.
To preserve the existing behavior, all instances of `<sbb-header-button>` and `<sbb-header-link>` that do not explicitly define `expandFrom` must now be configured with `hide-label-below="large"`.
END_COMMIT_OVERRIDE

BREAKING CHANGE: The `expandFrom` property of `<sbb-header-button>` and `<sbb-header-link>` has been renamed to `hideLabelBelow` and no longer has a default value. Previously, `expandFrom` defaulted to `large`.
To preserve the existing behavior, all instances of `<sbb-header-button>` and `<sbb-header-link>` that do not explicitly define `expandFrom` must now be configured with `hide-label-below="large"`


kyubisation (approved):
LGTM 👍

PR #4961, opened 2026-06-02
feat(calendar): add property `fixedMonth` to disable navigation

kyubisation:
This enables a use case where you want to allow users to edit a fixed range of days.

### DIFF
--- a/src/elements/calendar/calendar.stories.ts
+++ b/src/elements/calendar/calendar.stories.ts
@@ -439,6 +439,10 @@ export const CalendarEnhancedWideWeekNumbersMultiple: StoryObj = {
   },
 };
 
+export const CalendarFixedMonth: StoryObj = {
+  render: () => html`<sbb-calendar fixed-month="2023-08"></sbb-calendar>`,
+};
+
 const meta: Meta = {
   decorators: [withActions as Decorator],
   parameters: {
--- a/src/elements/calendar/calendar.stories.ts
+++ b/src/elements/calendar/calendar.stories.ts
@@ -440,7 +440,15 @@ export const CalendarEnhancedWideWeekNumbersMultiple: StoryObj = {
 };
 
 export const CalendarFixedMonth: StoryObj = {
-  render: () => html`<sbb-calendar fixed-month="2023-08"></sbb-calendar>`,
+  render: ({ amount }: Args) =>
+    html`<sbb-calendar
+      amount=${amount}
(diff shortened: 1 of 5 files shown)


src/elements/calendar/calendar/calendar.spec.ts
```diff
@@ -2161,6 +2161,33 @@ describe(`sbb-calendar`, () => {
...
+                    `}
+            </sbb-calendar>`,
+          );
+        });
+
+        it('should not render control buttons', () => {
+          expect(element.shadowRoot!.querySelector('.sbb-calendar__control')).to.be.null;
+        });
+
+        it('should render the given month', () => {
```
jeripeierSBB, line 2186: I think we should add a test, that tests also mutation of the fixedMonth, on order that the view gets correctly updated?

src/elements/calendar/calendar.stories.ts
```diff
@@ -439,6 +439,10 @@ export const CalendarEnhancedWideWeekNumbersMultiple: StoryObj = {
   },
 };
 
+export const CalendarFixedMonth: StoryObj = {
+  render: () => html`<sbb-calendar fixed-month="2023-08"></sbb-calendar>`,
```
jeripeierSBB: maybe a control? also the amount property would be nice. So it gets in the direction of a real example

src/elements/calendar/readme.md
```diff
@@ -158,6 +158,17 @@ all the days in the week.
 <sbb-calendar multiple week-numbers></sbb-calendar>
 ```
 
+### Fixed month
+
+In the case where a fixed month should be displayed, the `fixedMonth` property should be used.
+With this configuration, the currently displayed month cannot be changed.
```
jeripeierSBB: maybe a sentence that if there is an amount, the top left month is set here?

src/elements/calendar/calendar/calendar.spec.ts
```diff
@@ -2161,6 +2161,33 @@ describe(`sbb-calendar`, () => {
           });
         });
       });
+
+      describe('fixed-month', () => {
```
jeripeierSBB, line 2165: I recommend to also add a visual test, even when this basically covers the logic.

jeripeierSBB (commented):
Nice and simple solution!

jeripeierSBB (approved):
LGTM, thank you!

PR #4982, opened 2026-06-10
fix(calendar): fix 'dateFilter' type

TomMenga:
Should we mark it as a breaking change? In theory, it's transparent to consumers


jeripeierSBB (approved):
LGTM

PR #5004, opened 2026-06-18
feat(calendar): add 'visibleDays' method

TomMenga:


src/elements/calendar/calendar/calendar.component.ts
```diff
@@ -375,6 +375,13 @@ export class SbbCalendarElement<T = Date> extends SbbFormAssociatedMixin(SbbElem
     this.shadowRoot?.removeEventListener('slotchange', this._onSlotChange, { capture: true });
   }
 
+  /**
+   * Return the list of days that are visible in the calendar.
```
jeripeierSBB: Doesn't matter, but most of the time (except for two occurrences), we use the `s`
```suggestion
   * Returns the list of days that are visible in the calendar.
```

jeripeierSBB (approved):
LGTM, thank you!

PR #5005, opened 2026-06-18
fix(signet, logo): fix height adaption

jeripeierSBB:
Setting a dimension on the logo should scale it properly.


kyubisation (approved):
LGTM 👍

PR #5006, opened 2026-06-18
feat(download): initial implementation

### LINKED ISSUE #4949: story(download): initial implementation
### Feature Description

We want to provide a download component, which works as an anchor tag for downloading files.

Figma: https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?node-id=12009-749819&p=f&m=dev

Tech Spec:

- `<sbb-download>`
  - Inherits the `SbbLinkBaseElement` class and the `SbbIconNameMixin`
  - Can use the Sass mixins from the card module
  - Property `color: 'white' | 'milk' = 'white'`
  - Property `label: string = ''` (in rendered content defaults to basename of href)
  - Override the `renderIconName` method to add default icons based on href extension (Suggestion `switch (this.href.split('.').at(-1)) { case 'pdf: ...`)
  - Has unnamed slot for any content
  - Has internal static secondary button for the trash icon
- `<sbb-download-info>`
  - Host has display block
  - Property `type: string` (can optionally try to get the extension of the parent download href as a fallback)
  - Property `size: string` (if value contains non-number characters, render as is, if only numbers, shorten to closest B(yte) value (see SbbFileSelectorCommonElementMixin._formatFileSize))
  - Property `changed: string` which should accept an ISO8601 string
  - Property `nonAccessible: boolean` which indicates that a file is not accessible
  - Content is rendered in order: type, size, non-accessible, changed (while skipping info that has not been provided)

Examples:

```html
<sbb-download label="Name of download"></sbb-download>
```

```html
<sbb-download label="Name of download">
  Some description
</sbb-download>
```

```html
<sbb-download label="Name of download">
  <sbb-download-info ...></sbb-download-info>
</sbb-download>
```

```html
<sbb-download label="Name of download">
  Some description
  <sbb-download-info ...></sbb-download-info>
</sbb-download>
```

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

feerglas:
## Preflight Checklist

## Issue

This PR Closes #4949 

## Pull request checklist

See [Review Guidelines](../docs/REVIEW.md) for more information on what is checked during the review process.

## Changes

Changes in this pull request:

- Adding initial implementation for download component

## Browsers

- [x] Firefox Desktop
- [x] Chrome Desktop
- [x] Edge Desktop
- [x] Safari Desktop
- [x] Chrome Mobile
- [x] Safari Mobile

## Screen readers

- [x] JAWS Firefox Desktop
- [x] JAWS Chrome Desktop
- [x] NVDA Firefox Desktop
- [x] NVDA Chrome Desktop
- [x] VoiceOver Safari Mobile

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/elements/download/download/download.visual.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+  describeViewports,
+  visualDiffDefault,
+  visualRegressionFixture,
+} from '../../core/testing/private.ts';
+
+import type { SbbDownloadElement } from './download.component.ts';
+
+import '../../download.ts';
+
+describe(`sbb-download`, () => {
```
jeripeierSBB, line 16: Please also add test for
 - emulation of media for forced colors and dark mode
 - ellipsis case
 - hover, active and focus state.

src/elements/download/download-info/download-info.visual.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { html } from 'lit';
+
+import { describeViewports, visualDiffDefault } from '../../core/testing/private.ts';
+
+import '../../download.ts';
+
+describe(`sbb-download-info`, () => {
+  describeViewports({ viewports: ['zero', 'large'] }, () => {
```
jeripeierSBB, line 17: it would be nice to create cases for all sensible properties, so concatenation is visually tested. On the other hand I see tests in download-info.spec.ts, so maybe it's ok :)

src/elements/download/download/download.visual.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+      it(
+        visualDiffDefault.name,
+        visualDiffDefault.with((setup) => {
+          setup.withSnapshotElement(root);
+        }),
+      );
+    });
+
+    describe('with custom content and info', () => {
+      beforeEach(async () => {
```
jeripeierSBB: instead of a beforeEach block for only one test, you could also directly use `visualDiffDefault.withFixture()...` (check occurrences in code base).

src/elements/download/download.stories.ts
```diff
@@ -0,0 +1,318 @@
...
+  control: {
+    type: 'text',
+  },
+};
+
+const color: InputType = {
+  control: {
+    type: 'inline-radio',
+  },
+  options: ['white', 'milk'],
```
jeripeierSBB: nitpicky, but helps for potential refactorings :)
```suggestion
  options: ['white', 'milk'] satisfies SbbDownloadElement['color'][],
```

src/elements/download/download/download.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    return (extension && fileExtensionIcons.get(extension)) || 'document-standard-small';
+  }
+
+  /**
+   * Auto-assigns a `sbb-download-info` placed in the unnamed slot to the
+   * dedicated `info` slot, so consumers don't have to set the slot manually.
+   */
+  private _handleSlotchange(): void {
+    for (const child of Array.from(this.children)) {
+      if (child.localName === 'sbb-download-info' && child.slot !== 'info') {
```
jeripeierSBB: Basically fine, but normally we do this (for this kind of situations) in the child components connectedCallback(). maybe the section in the readme could also be adapted?
```ts
    super.connectedCallback();
    this.slot ||= 'info';
  }
```

src/elements/download/readme.md
```diff
@@ -0,0 +1,120 @@
+The `<sbb-download>` is a component that displays a downloadable document, styled as a
```
jeripeierSBB, line 1: Nice documentation!

src/elements/download/download-info/download-info.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect(element.shadowRoot!.textContent!.trim()).to.be.equal('FOO, not accessible');
+  });
+
+  it('renders the size as is when it contains non-numeric characters', async () => {
+    element.size = '123 KB';
+    await waitForLitRender(element);
+    expect(element.shadowRoot!.textContent!.trim()).to.be.equal('123 KB');
+  });
+
+  it('formats a numeric size to the closest unit', async () => {
```
jeripeierSBB, line 71: I'd appreciate some more combinations here

src/elements/download/download-info/download-info.component.ts
```diff
@@ -0,0 +1,153 @@
...
+export class SbbDownloadInfoElement extends SbbElement {
+  public static override readonly elementName: string = 'sbb-download-info';
+  public static override styles: CSSResultGroup = [unsafeCSS(style)];
+
+  /**
+   * The type of the document (e.g. `PDF`). Falls back to the extension of the
+   * parent `sbb-download`.
+   */
+  @forceType()
+  @property({ reflect: true, converter: omitEmptyConverter })
```
jeripeierSBB: I think all properties in this file don't need to be reflected as they are not considered for styles.

src/elements/download/download-info/download-info.component.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  /** Resolves the type, falling back to the extension of the parent download's
+   * href.
+   */
+  private _resolveType(): string {
+    if (this.type) {
+      return this.type;
+    }
+
+    const parentHref = this.closest?.('sbb-download')?.href ?? '';
```
jeripeierSBB: wondering whether it makes sense to have a method / getter in the sbb-download? So it would be the responsibility of the sbb-download to parse the href.

src/elements/download/download-info/download-info.component.ts
```diff
@@ -0,0 +1,153 @@
...
+   * case the type is redundant for assistive technology.
+   */
+  private _isTypeRedundant(): boolean {
+    const parent = this.closest?.('sbb-download');
+
+    if (!parent || parent.label) {
+      return false;
+    }
+
+    const fileName = parent.href.split(/[?#]/)[0].split('/').filter(Boolean).at(-1) ?? '';
```
jeripeierSBB: looks like a duplication of line 61?

src/elements/download/download-info/download-info.component.ts
```diff
@@ -0,0 +1,153 @@
...
+      return '';
+    }
+
+    const date = new Date(this.changed);
+
+    if (Number.isNaN(date.getTime())) {
+      return this.changed;
+    }
+
+    return new Intl.DateTimeFormat(this._language.current, {
```
jeripeierSBB: We should use the dateAdapter to be consistent with other date related components

src/elements/download/download-info/download-info.component.ts
```diff
@@ -0,0 +1,153 @@
...
+      month: '2-digit',
+      day: '2-digit',
+    }).format(date);
+  }
+
+  protected override render(): TemplateResult {
+    const separator = ', ';
+    const type = this._resolveType();
+    const items = [
+      this._resolveSize(),
```
jeripeierSBB: Normally we pre-render the values in the dedicated setter / requestUpdate() or willChange(). With the current solution, for every re-render all the values need to be calculated.

src/elements/download/download-info/download-info.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect(element.shadowRoot!.textContent!.trim()).to.be.equal('not-a-date');
+  });
+
+  it('falls back to the parent download extension as type', async () => {
+    const download = await fixture(html`
+      <sbb-download href="files/report.pdf">
+        <sbb-download-info></sbb-download-info>
+      </sbb-download>
+    `);
+    const info = download.querySelector<SbbDownloadInfoElement>('sbb-download-info')!;
```
jeripeierSBB, line 142: Should not be needed for correct typing
```suggestion
    const info = download.querySelector('sbb-download-info')!;
```

src/elements/download/download-info/download-info.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    await waitForLitRender(info);
+    expect(info.shadowRoot!.textContent!.trim()).to.be.equal('PDF');
+  });
+});
```
jeripeierSBB, line 187: I recommend to also test moving an info between different downloads with different extensions. in this case, the type should be updated accordingly. I feel (not tested) that with the current implementation, this doesn't work.

src/elements/download/download/download.component.ts
```diff
@@ -0,0 +1,138 @@
...
+  @property({ reflect: true }) public accessor color: 'white' | 'milk' = 'white';
+
+  /** The label of the download. Defaults to the file name of the `href`. */
+  @forceType()
+  @property({ converter: omitEmptyConverter })
+  public accessor label: string = '';
+
+  /**
+   * The download behavior is always enabled and handled internally, therefore
+   * this property cannot be disabled and is not reflected to the host element.
```
jeripeierSBB, line 69: with `@internal` we could also completely hide it from documentation

src/elements/download/download/download.component.ts
```diff
@@ -0,0 +1,138 @@
...
+   */
+  @forceType()
+  @property({ type: Boolean, reflect: false })
+  public override accessor download: boolean = true;
+
+  protected override willUpdate(changedProperties: PropertyValues<this>): void {
+    super.willUpdate(changedProperties);
+
+    // The download behavior is always enabled; consumers must not be able to disable it.
+    this.download = true;
```
jeripeierSBB: should we really always set it back to true? We can consider this a consumer fault if someone sets it to false. I think we should use a setter / getter and ignore values in setter and always return true in the getter.

src/elements/download/download/download.scss
```diff
@@ -0,0 +1,69 @@
+@use '../../core/styles' as sbb;
```
jeripeierSBB, line 1: Is the hover transition applied? By manual testing it feels hard (I can be wrong!) I think a 1px shift on hover of the content (like in the card) is missing?

src/elements/download/download/download.scss
```diff
@@ -0,0 +1,69 @@
...
+@use '../../card/card.global' as card;
+
+:host {
+  @include card.variables--global;
```
jeripeierSBB, line 7: What is the case that we need to re-include global variables? nesting?

src/elements/download/download/download.scss
```diff
@@ -0,0 +1,69 @@
...
+  @include sbb.ellipsis;
+
+  font-weight: bold;
+  color: var(--sbb-download-label-color);
+}
+
+// Custom slotted content and the `sbb-download-info` share the same style
+.sbb-download__custom-content,
+.sbb-download__info {
+  @include sbb.text-xxs;
```
jeripeierSBB: We should only include the minimal needed styles.
```suggestion
  line-height: var(--sbb-typo-line-height-text);
  letter-spacing: var(--sbb-typo-letter-spacing-text);
  font-size: var(--sbb-text-font-size-xxs);
```

src/elements/download/download/download.spec.ts
```diff
@@ -0,0 +1,125 @@
+import { assert, expect } from '@open-wc/testing';
```
jeripeierSBB, line 1: nice set of tests!

jeripeierSBB (commented):
Nice work! I left some some ideas to improve or fix.

src/elements/download/download-info/download-info.component.ts
feerglas: obsolete after refactor

src/elements/download/download-info/download-info.spec.ts
feerglas: oh wow, good catch @jeripeierSBB, indeed this was an issue ! fixed it by calling requestUpdate in the connectedCallback

src/elements/download/download/download.component.ts
feerglas: this doesn't seem to work. any suggestions or should i leave it as is @jeripeierSBB ?

src/elements/download/download/download.scss
feerglas: good catch @jeripeierSBB . I thought about the case when a download-component is included in another card-like element: in such cases, it would inherit the parent card's overriden values. or am i wrong? does it make sense?

src/elements/download/download/download.component.ts
```diff
@@ -0,0 +1,124 @@
+import { type CSSResultGroup, html, type TemplateResult, unsafeCSS } from 'lit';
```
jeripeierSBB: I forgot one case: When the href on the download component changes, the download info should update. for that I recommend to use the SbbPropertyWatcherController in the download-info component. Please also add a test for that case.

src/elements/download/download/download.scss
jeripeierSBB: yeah maybe it's that case that should be covered. Could you add a visual test, so the line makes sense somewhere ;)

src/elements/download/download/download.visual.spec.ts
```diff
@@ -0,0 +1,96 @@
+import { html, type TemplateResult } from 'lit';
```
jeripeierSBB, line 1: The icons used in this tests need to be preloaded (tools/web-test-runner/preload-icons.ts). See https://lyne-components-visual-regression-diff-pr5006.app.sbb.ch/compare/sbb-download/color=milk-emulateMedia=(forcedColors=false-darkMode=false)_active as example

src/elements/download/download/download.component.ts
jeripeierSBB: You are right, it's not possible to make a derived property internal. We keep your solution :)

jeripeierSBB (changes requested):
Excellent code quality, almost done.

src/elements/download/download/download.component.ts
feerglas: good catch, forgot about that

jeripeierSBB (approved):
LGTM. Thank you!

PR #5017, opened 2026-06-22
build: fix Temporal types

kyubisation:
We now use the types from TypeScript, which provide Temporal types since version 6.0.


jeripeierSBB (approved):
LGTM

PR #5031, opened 2026-06-24
fix(clock): made `start` and `stop` method synchronous

TomMenga:
Improve the robustness of the start and stop methods, making them synchronous.

The current implementation lands in an inconsistent state if `start()` and `stop()` is called sequentially


jeripeierSBB (approved):
LGTM, thank you

PR #5032, opened 2026-06-24
docs(carousel): improve 'link' use case a11y

TomMenga:
Open topic:
- The focus outline of the image is not visible because it's contained in an overflow element. Solving it is a bit complicated :/
  We could either leave it as it is or, as a compromise, add an `outline-offset: -1px` on the <a> element

### DIFF
--- a/src/elements/carousel/_carousel.global.scss
+++ b/src/elements/carousel/_carousel.global.scss
@@ -9,3 +9,9 @@ $theme: 'standard' !default;
   --sbb-carousel-paginator-padding: var(--sbb-spacing-responsive-xs);
   --sbb-carousel-list-border-radius: var(--sbb-border-radius-4x);
 }
+
+@mixin rules {
+  sbb-carousel-item sbb-image {
+    border-radius: unset;
+  }
+}
--- a/src/elements/carousel/carousel.stories.ts
+++ b/src/elements/carousel/carousel.stories.ts
@@ -109,6 +109,7 @@ const Template = ({ imgType, ...args }: Args): TemplateResult => html`
                     href="https://github.com/sbb-design-systems/lyne-components"
                     target="_blank"
                     aria-label="Navigate to lyne-angular repo"
+                    style="border-radius: var(--sbb-carousel-border-radius)"
                   >
                     <sbb-image
                       image-src=${img}
--- a/src/elements/carousel/carousel.stories.ts
+++ b/src/elements/carousel/carousel.stories.ts
@@ -108,7 +108,7 @@ const Template = ({ imgType, ...args }: Args): TemplateResult => html`
                   <a
                     href="https://github.com/sbb-design-systems/lyne-components"
                     target="_blank"
-                    tabindex="-1"
+                    aria-label="Navigate to lyne-angular repo"
                   >
                     <sbb-image
                       image-src=${img}
--- a/src/elements/carousel/carousel.stories.ts
+++ b/src/elements/carousel/carousel.stories.ts
@@ -109,7 +109,6 @@ const Template = ({ imgType, ...args }: Args): TemplateResult => html`
                     href="https://github.com/sbb-design-systems/lyne-components"
                     target="_blank"
                     aria-label="Navigate to lyne-angular repo"
-                    style="border-radius: var(--sbb-carousel-border-radius)"
                   >
                     <sbb-image
                       image-src=${img}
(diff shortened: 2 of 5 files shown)


src/elements/carousel/carousel.stories.ts
jeripeierSBB, line 87: I also saw, that the border-radius in the link example is wrong. and if you look very closely to the corner, you see the blue outline of the a element.

TomMenga: I don't see a satisfying solution :/ The negative offset doesn't work either

The only solution that comes to mind is to have an outline when something is focused inside the carousel. However, this is not ideal in cases where you have a small element inside (like a button)
image

jeripeierSBB: I think we should only solve it on example side, not generally. we can't do it precise enough :)

src/elements/carousel/carousel.stories.ts
```diff
@@ -108,7 +108,8 @@ const Template = ({ imgType, ...args }: Args): TemplateResult => html`
                   <a
                     href="https://github.com/sbb-design-systems/lyne-components"
                     target="_blank"
-                    tabindex="-1"
+                    aria-label="Navigate to lyne-angular repo"
+                    style="border-radius: var(--sbb-carousel-border-radius)"
```
jeripeierSBB: Doesn't look that it's fixed?

TomMenga: Maybe I misunderstood your request 🤔 Let's talk about it on monday 👍

jeripeierSBB: My thought was that the marked corners shouldn't have a border radius at all (if you compare to other examples).
Bildschirmfoto 2026-06-29 um 08 28 52

TomMenga: I confirm that I misunderstood it 😐 I'll fix that 👍

src/elements/carousel/_carousel.global.scss
```diff
@@ -9,3 +9,9 @@ $theme: 'standard' !default;
   --sbb-carousel-paginator-padding: var(--sbb-spacing-responsive-xs);
   --sbb-carousel-list-border-radius: var(--sbb-border-radius-4x);
 }
+
+@mixin rules {
+  sbb-carousel-item sbb-image {
+    border-radius: unset;
+  }
```
TomMenga, line 16: We also need to cover the `figure > sbb-image` case

jeripeierSBB: I think this is ok. Just wondering if ever sbb-images are going to be used that are not full width? Like some composed stuff where you maybe want to have the border-radius. On the other hand, you could override it there.

jeripeierSBB (commented):
The radius looks wrong here: https://lyne-components-visual-regression-diff-pr5032.app.sbb.ch/compare/sbb-carousel/shadow=false_native ?

src/elements/carousel/_carousel.global.scss
TomMenga: I thought of the possibility, but it's implausible.
We could make it more specific like this, but it feels excessive
```
sbb-carousel-item {
  &>sbb-image,
  &>figure>sbb-image,
  &>link>sbb-image {
    border-radius: unset;
  }
}
```

jeripeierSBB: Let's go the less code way :)

TomMenga:
> The radius looks wrong here: https://lyne-components-visual-regression-diff-pr5032.app.sbb.ch/compare/sbb-carousel/shadow=false_native ?

No, the problem is that the size of the image is static and exceeds the zero viewport (the screenshot is cut)

jeripeierSBB:
> > The radius looks wrong here: https://lyne-components-visual-regression-diff-pr5032.app.sbb.ch/compare/sbb-carousel/shadow=false_native ?
> 
> No, the problem is that the size of the image is static and exceeds the zero viewport (the screenshot is cut)

It seems on all viewport sizes?

TomMenga:
> > > The radius looks wrong here: https://lyne-components-visual-regression-diff-pr5032.app.sbb.ch/compare/sbb-carousel/shadow=false_native ?
> > 
> > 
> > No, the problem is that the size of the image is static and exceeds the zero viewport (the screenshot is cut)
> 
> It seems on all viewport sizes?

My bad again :/
We don't have the bottom border-radius only when "shadow" is set on the carousel

jeripeierSBB (approved):
Maximum confusion in such a simple PR :D. Sorry and thanks for the patience. LGTM

PR #5036, opened 2026-06-26
style(calendar): fix style for month/year view with multiple months

DavideMininni-Fincons:
There's a bug on calendar when not in day view and amount > 1:

image

### DIFF
--- a/src/elements/calendar/calendar/calendar.scss
+++ b/src/elements/calendar/calendar/calendar.scss
@@ -31,7 +31,7 @@
 
 .sbb-calendar__day-view {
   :host(:not(:state(view-day))) & {
-    visibility: hidden;
+    display: none;
   }
 }
 


src/elements/calendar/calendar/calendar.scss
```diff
@@ -31,7 +31,7 @@
 
 .sbb-calendar__day-view {
   :host(:not(:state(view-day))) & {
-    visibility: hidden;
+    display: none;
```
jeripeierSBB, line 34: I think this is intentionally to avoid layout shifts when changing to year / month selections.

DavideMininni-Fincons: I got it; however the result on datepicker on wide + month view is 
image

what's the best approach in your opinion?

jeripeierSBB: it's intended, even when looking not optimized in this case. @kyubisation FYI

DavideMininni-Fincons: It's fine, so i'll close the PR

PR #5061, opened 2026-07-07
docs(header): fix typo

TomMenga:


jeripeierSBB (approved):
Good catch, thank you!

PR #5077, opened 2026-07-08
fix(chip): ensure `sbb-chip` is defined when created

jeripeierSBB:
Avoids that the `sbb-chip` can get tree shaked although it's needed.


kyubisation (approved):
LGTM 👍

PR #5084, opened 2026-07-09
feat(easter-egg): add easter egg as experimental component

stefan-schubert-sbb:
easter egg


scripts/build.ts
```diff
@@ -387,13 +387,29 @@ function stateTransform(): PluginOption {
 }
 
 function rawLoader(): PluginOption {
+  const mimeByExtension: Record<string, string> = {
```
kyubisation, line 390: Nitpick; If the values are the same as `image/${mime-type}` then we don't need this record and can just inline it (or keep a list if you want to be explicit).

scripts/build.ts
```diff
@@ -387,13 +387,29 @@ function stateTransform(): PluginOption {
...
+    webp: 'image/webp',
+  };
   return {
     name: 'raw-loader',
     load(id) {
       if (id.endsWith('?raw')) {
         const code = `export default ${JSON.stringify(readFileSync(id.replace('?raw', ''), 'utf8'))};`;
         return { code };
       }
+      if (id.endsWith('?inline')) {
```
kyubisation: Nitpick (as it cannot overlap)
```suggestion
} else if (id.endsWith('?inline')) {
```

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,759 @@
...
+ */
+export class SbbEasterEggElement extends SbbDialogElement {
+  public static override readonly elementName: string = 'sbb-easter-egg';
+  public static override elementDependencies: SbbElementType[] = [
+    SbbDialogCloseButtonElement,
+    SbbButtonElement,
+  ];
+  public static override styles: CSSResultGroup = [
+    SbbDialogElement.styles,
+    css`
```
kyubisation: Please extract this to a .scss file

kyubisation (commented):
Nice work!
Generally looks good to me, some minor issues

src/elements-experimental/easter-egg/easter-egg.component.ts
stefan-schubert-sbb: background-image definition can not be defined in scss, because it must be loaded inline

scripts/build.ts
stefan-schubert-sbb: the list is necessary, because for jpg the mime type is image/jpeg

src/elements-experimental/easter-egg/easter-egg.component.ts
kyubisation: You can manually convert the image to base64, which is what happens anyway with ?inline

src/elements-experimental/easter-egg/easter-egg.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+  it('exposes score getter that starts at 0', () => {
+    expect(element.score).to.be.equal(0);
+  });
+
+  it('keeps the high score across restarts and resets it on close', async () => {
+    await openDialog(element);
+    expect(element.highScore).to.be.equal(0);
+
+    // Simulate a run of 5 points ending with game over — high score commits on finish.
+    const internal = element as unknown as Record<string, () => void> & Record<string, unknown>;
```
kyubisation: Do not depend on private properties/methods. Our plan is to migrate to native private fields in 2027, which would then cause this test to break.
If you need to call private API, make it protected and call it as element['propertyName'].
Ideally there could be another way to test this functionality, without depending on private or protected API.

kyubisation (changes requested):

src/elements-experimental/easter-egg/easter-egg.spec.ts
stefan-schubert-sbb: I have rewritten the test case.

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+ * the readonly `score` getter.
+ */
+export class SbbEasterEggElement extends SbbDialogElement {
+  public static override readonly elementName: string = 'sbb-easter-egg';
+  public static override elementDependencies: SbbElementType[] = [
+    SbbDialogCloseButtonElement,
+    SbbButtonElement,
+    SbbTitleElement,
+  ];
+  public static override styles: CSSResultGroup = [SbbDialogElement.styles, unsafeCSS(style)];
```
jeripeierSBB: This is not needed, as it is inherited
```suggestion
  public static override styles: CSSResultGroup = [unsafeCSS(style)];
```

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+    this._highScore = 0;
+    this._resetGame(true);
+    this._openAbortController?.abort();
+    this._openAbortController = null;
+  }
+
+  private async _initGame(): Promise<void> {
+    if (isServer) {
+      return;
+    }
```
jeripeierSBB: I think there is no path, where initGame is called on server side?
```suggestion
```

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+    return img
+      .decode()
+      .then(() => img)
+      .catch(() => img);
+  }
+
+  private _attachInputListeners(): void {
+    if (isServer) {
+      return;
+    }
```
jeripeierSBB: Also here, It's guarded before
```suggestion
```

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+    this._openAbortController?.abort();
+    this._openAbortController = new AbortController();
+    const signal = this._openAbortController.signal;
+
+    document.addEventListener('keyup', (e) => this._onKeyup(e), { signal });
```
jeripeierSBB: Only from testing the game, it feels for me, that it should be `keydown`. Otherwise if pressing kind of slowly it's already too late.
```suggestion
    document.addEventListener('keydown', (e) => this._onKeyup(e), { signal });
```

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+      default:
+        return;
+    }
+    this._started = true;
+  }
+
+  private _resetGame(withoutStart?: boolean): void {
+    if (isServer) {
+      return;
+    }
```
jeripeierSBB:
```suggestion
```

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+      // Vertical.
+      return tail.y - prev.y > 0 ? 13 : 11;
+    }
+    // Horizontal.
+    return tail.x - prev.x > 0 ? 10 : 12;
+  }
+
+  protected override render(): TemplateResult {
+    const lang = this.language.current;
+    const buttonSize = this._compact ? 's' : 'm';
```
jeripeierSBB: This is not basically a problem, but an anti-pattern. I recommend to not switch button sizes depending on breakpoints as the buttons itself already look different depending on breakpoint. With that we could also get rid of the mediamatcher, which per design doesn't work with well with SSR.

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,685 @@
...
+    const buttonLabel = this._gameOver ? i18nSnakeRestart[lang] : i18nSnakeStart[lang];
+
+    return html`
+      <div class="sbb-dialog__container">
+        <div @animationend=${this.onOverlayAnimationEnd} class="sbb-dialog">
+          <div
+            @click=${(event: Event) => this.closeOnSbbOverlayCloseClick(event)}
+            class="sbb-dialog__wrapper sbb-easter-egg__wrapper"
+          >
+            <header class="sbb-easter-egg__header">
```
jeripeierSBB: Why are you not using the sbb-dialog-title component?

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
+.sbb-easter-egg__wrapper {
```
jeripeierSBB: on zero breakpoint the game seems cut? http://localhost:6006/?path=/story/experimental-easter-egg--default&globals=viewport.value:zero

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
...
+  flex-direction: column;
+  align-items: center;
+  gap: var(--sbb-spacing-responsive-xxxs);
+  padding: var(--sbb-spacing-responsive-s);
+  background-color: var(--sbb-dialog-background-color);
+  border-radius: var(--sbb-border-radius-4x);
+  margin-inline: auto;
+  width: 50vw;
+  min-width: fit-content;
+  max-width: 828px;
```
jeripeierSBB:
```suggestion
  max-width: #{sbb.px-to-rem-build(828)};
```

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
...
+}
+
+.sbb-easter-egg__score {
+  display: flex;
+  justify-content: space-between;
+  width: 100%;
+  gap: var(--sbb-spacing-fixed-4x);
+  font-family: var(--sbb-typo-type-face-sbb-roman);
+  font-size: var(--sbb-font-size-text-m);
+  font-weight: 600;
```
jeripeierSBB: We have no direct support of weight 600. 
```suggestion
  font-weight: bold;
```

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
...
+  color: var(--_sbb-dialog-color);
+}
+
+.sbb-easter-egg__score {
+  display: flex;
+  justify-content: space-between;
+  width: 100%;
+  gap: var(--sbb-spacing-fixed-4x);
+  font-family: var(--sbb-typo-type-face-sbb-roman);
+  font-size: var(--sbb-font-size-text-m);
```
jeripeierSBB: In case you want to have it working :P
```suggestion
  font-size: var(--sbb-text-font-size-m);
```

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
...
+  margin: 0;
+  color: var(--_sbb-dialog-color);
+}
+
+.sbb-easter-egg__score {
+  display: flex;
+  justify-content: space-between;
+  width: 100%;
+  gap: var(--sbb-spacing-fixed-4x);
+  font-family: var(--sbb-typo-type-face-sbb-roman);
```
jeripeierSBB: font-families never need to be redeclared.
```suggestion
```

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
...
+  font-weight: 600;
+}
+
+.sbb-easter-egg__canvas-wrapper {
+  display: flex;
+  justify-content: center;
+  position: relative;
+  line-height: 0;
+  padding: var(--sbb-spacing-fixed-3x);
+  padding-block-end: 150px;
```
jeripeierSBB:
```suggestion
  padding-block-end: #{sbb.px-to-rem-build(150)};
```

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,72 @@
...
...elZQzHE96EgoGwZsH39kk1HiMAf+RRpXEOt/lLMMAV3O89JuHyEyJEXwK+qliCwuBxmEN9yrNwj1D2E6S1YsoWh6Oc3rXr4JccDXxJfuHVyxkpWrxmTEdYM6T2wdl/wvGZsYm+Mp9bV3hRZ1SdIhfg06I2w9ui7ft5dH13bTyhte+tofEDHE+qXF2HO+2IjNxxQnLskmsKlAWjJE81q7WFmynXcG6vFGcX0mw8tgY0c1DEMpfxOhjxOZowcNZHtjG61hyBPTotcdM8FPWSLMEiqzu40kOCpTTb4BkvNy1YGD76w/pZFDyvGj36Nr71fOt5ASTEun8JiujJJPobH9JZiNn/Jw85S/Ha9hG0Tpub8Zu1COZO1jl9198OW9lhMD2L3ozMWhFiAu+vEYxYH3YWLDPSKuScMHAIROkjsJCnH4MTFBxM8mA+FTTjcvq1qRmTY0Qvw4oAUSZ3uM2ScjXpKLwiY3hxihgf3zX0ZJFeM6+9ZU3VH/RBzienChuS/ca97BV5qFUyPemdLvBKpgSanN5LBYIedZrko/C35Bu89A+t4N/6HRqAP+xlwIxOlKKbbPinBzl4eMALaMkGZ0RAhacdaSxgE3DAwfgz3FgjunOct9+glbWcrf91B0Ws/YVuEiAlCDL5rwv07ZfBZ6cifHiDTiG6+F97rCSUNq2b4JZJ56XYk8KJggT8Th2QT4739i90z7M/dihItFq7ZHOi8pz/6LigOh5nhxx7rU3L3k82no2Eq10LTvoZzEcXxG/pGrjYWOEmVTPdEjDjWXoYX6Mv+X7e/kLJT21/2srZOUGm7LRsDR5DS7T5a2tQS1O5Z88wA/mUKGEbBOuSUGjHJp5WYBBbe1/3vcM8yNXa+OLFtUZGn4fZMG3kgj++xalSZHuCz3NkfDow4DcPh7sxie8SNf2NBMN7a9s7zYLH/h8vSrYG7zRjN9gu1TtBNl5Ze9VoxGZF6673/+GpAftMNiSUtc7yDdgKvId/6VY4JXvofDmi+EM8QZbN6/sldVsLnbekSM9MpxXwoxOcCFB9Z12qXPSG/8LoAi26dkk13ze/oHqJzHgA389+d4+TW9fCGN4CYX4HkdLB00yt5fC/JKX31fBTxo/9zUYJ8vgZTqMFNz0GB/baNnzfNUx0GJugx2wHPginwSWwgqMe3U1XvPg+8yMJEB6Ff18KKsBtvAVJIP
```
jeripeierSBB:
```suggestion
  height: #{sbb.px-to-rem-build(320)};
  width: #{sbb.px-to-rem-build(320)};
```

jeripeierSBB: Should there be a CSS variable for 320 as it is used three times?
```suggestion
  height: #{sbb.px-to-rem-build(320)};
```

src/elements-experimental/easter-egg/easter-egg.snapshot.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+import type { SbbEasterEggElement } from './easter-egg.component.ts';
+
+import '../easter-egg.ts';
+
+describe(`sbb-easter-egg`, () => {
+  let element: SbbEasterEggElement;
+
+  describe('closed', () => {
+    beforeEach(async () => {
+      document.documentElement.setAttribute('lang', 'en');
```
jeripeierSBB: en is the default
```suggestion
```

src/elements-experimental/easter-egg/readme.md
```diff
@@ -0,0 +1,56 @@
...
+- Colliding with a wall or with itself ends the run — the final score is shown above a **Restart** button in the middle of the game field.
+
+## Internationalization
+
+The visible labels — the dialog title (**Swiss Boa Express.**), the score label (**Score**), and the start / restart button — are localized in English, German, French and Italian. They follow the current UI language, read from the `lang` attribute on `<html>`.
+The screen-reader announcement (`accessibilityLabel`) is kept in sync automatically.
+
+## Responsive behavior
+
+The start / restart button uses the `sbb-button` component. It renders in the default `m` size on larger viewports and switches to the compact `s` size below the small breakpoint.
```
jeripeierSBB: In case you agree with my previous comment, this section could be removed.
```suggestion
```

jeripeierSBB (commented):
This review made fun :-). Nice game!

It seems that if the viewport height is insufficient, the game is cropped and you don't see items in the hidden area, so it's a little but confusing in this situation. Should there maybe an error? shrinking is probably not an option. Happens if you e.g. use a phone in landscape mode. 

Image

jeripeierSBB:
I removed the "target: 4.x" label. There is much more effort needed than first thought, so we decided to only land this in major version 5.

src/elements-experimental/easter-egg/easter-egg.component.ts
```diff
@@ -0,0 +1,662 @@
...
+        ? dx > 0
+          ? 'ArrowRight'
+          : 'ArrowLeft'
+        : dy > 0
+          ? 'ArrowDown'
+          : 'ArrowUp';
+    this._applyDirectionKey(key);
+  }
+
+  private _onKeyup(e: KeyboardEvent): void {
```
jeripeierSBB:
```suggestion
  private _onKeydown(e: KeyboardEvent): void {
```

src/elements-experimental/easter-egg/easter-egg.scss
```diff
@@ -0,0 +1,65 @@
...
+.sbb-easter-egg__content {
+  overflow: hidden;
+  width: 50vw;
+  min-width: fit-content;
+  max-width: #{sbb.px-to-rem-build(828)};
+  padding: 0;
+  margin-block: var(--sbb-dialog-padding-block);
+  margin-inline: var(--sbb-dialog-padding-inline);
+  @include sbb.mq($to: small) {
+    width: calc(100% - (var(--sbb-spacing-fixed-4x) * 2));
```
jeripeierSBB: not sure, but it seems that you are removing the padding here? So it could probably be the padding variable?
```suggestion
    width: calc(100% - (var(--sbb-dialog-padding-inline) * 2));
```

src/elements-experimental/easter-egg/readme.md
jeripeierSBB:
```suggestion
```

stefan-schubert-sbb:
> This review made fun :-). Nice game!
> 
> It seems that if the viewport height is insufficient, the game is cropped and you don't see items in the hidden area, so it's a little but confusing in this situation. Should there maybe an error? shrinking is probably not an option. Happens if you e.g. use a phone in landscape mode.
> 
> Image

Should be handled by the consumer not to open the game in landscape mode.

jeripeierSBB (approved):
LGTM. The only issue I see is the landscape mode on mobile phones where the bottom of the field is cut. Maybe acceptable?

jeripeierSBB:
> LGTM. The only issue I see is the landscape mode on mobile phones where the bottom of the field is cut. Maybe acceptable?

Ah I missed your previous comment. Then we should add it to documentation?

jeripeierSBB (approved):
LGTM

kyubisation (approved):
LGTM 👍 
Thanks!

PR #5113, opened 2026-07-21
refactor(form-field): extend from control to include interacted and invalid

kyubisation:
This will allow us to sync state with Angular Signal Forms.


TomMenga (approved):
LGTM 👍

PR #5123, opened 2026-07-27
fix(calendar): prevent null day crash with min/max updates

### LINKED ISSUE #5122: [Bug]: sbb-calendar-day crashes when min/max updates before slotted day value is initialized
### Preflight Checklist

### Bug type
Functionality

Components

5.0.0 (also reproducible on current `main`)

### Operating system
macOS (Darwin)

### Browser / Browser version
Chromium (Playwright)

### Input mode
Mouse/Keyboard

### Additional settings
N/A

### What happened?
When `sbb-calendar` has `min`/`max` set and a slotted `sbb-calendar-day` is temporarily uninitialized (no valid `slot` date yet), the calendar can crash with a null date error.

**Root cause:** `SbbCalendarDayElement` can run state updates before `value` is initialized from `slot`. In that state, date operations are still executed with `null`:
1. range checks in `_isDayInRange` call `clampDate`/`compareDate` with `null`;
2. rendering calls `dateAdapter.getDate(this.value)` while `value` is `null`.

**Expected:** Calendar day should not crash while slotted day value is temporarily null/uninitialized; normal min/max behavior should remain unchanged once a valid date exists.

**Actual:** Runtime exception (`TypeError`) is thrown (e.g. `NativeDateAdapter.getYear`/`getDate` called with `null`).

1. Render a calendar with a slotted day that has no initialized date yet:
   - `<sbb-calendar><sbb-calendar-day></sbb-calendar-day></sbb-calendar>`
2. Set `calendar.min` and `calendar.max`.
3. Trigger update cycle.
4. Observe crash from date adapter receiving `null`.

A regression test was added in `src/elements/calendar/calendar-day/calendar-day.spec.ts`:
`does not crash when parent min/max updates before slotted day value is initialized`

### Relevant log output
TypeError: Cannot read properties of null (reading 'getDate')
  at NativeDateAdapter.getDate (.../native-date-adapter.ts)
  at SbbCalendarDayElement.renderTemplate (.../calendar-day.component.ts)

(Equivalent crash path also reported with `getYear` from `_isDayInRange -> clampDate -> compareDate` when min/max checks run with null day value.)

rolego:
## Summary
- guard `sbb-calendar-day` range checks when the day value is temporarily null/uninitialized
- guard day rendering so `dateAdapter.getDate` is not called with `null`
- add regression test for slotted `sbb-calendar-day` without initialized slot while `min`/`max` are set

## Root cause
`SbbCalendarDayElement` may run update logic before `slot` has initialized `value`. In that transient state, date adapter methods were called with `null`, which could crash in either rendering (`getDate`) or range checks (`_isDayInRange` -> `clampDate` -> `compareDate`).

## Verification
- `yarn -s wtr --group default src/elements/calendar/calendar-day/calendar-day.spec.ts`

Closes #5122

### DIFF
--- a/src/elements/calendar/calendar-day/calendar-day.component.ts
+++ b/src/elements/calendar/calendar-day/calendar-day.component.ts
@@ -85,6 +85,9 @@ export class SbbCalendarDayElement<T = Date> extends SbbCalendarCellBaseElement<
   }
 
   private _isDayInRange(min: T | null, max: T | null): boolean {
+    if (!this.value) {
+      return true;
+    }
     if (!min && !max) {
       return true;
     }
@@ -97,7 +100,7 @@ export class SbbCalendarDayElement<T = Date> extends SbbCalendarCellBaseElement<
 
   protected override renderTemplate(): TemplateResult {
     return html` <span class="sbb-calendar-day__value" aria-hidden="true">
-        ${this.dateAdapter.getDate(this.value)}
+        ${this.value ? this.dateAdapter.getDate(this.value) : ''}
       </span>
       <span class="sbb-calendar-day__extra">
         <slot @slotchange=${(event: Event) => this._handleSlotchange(event)}></slot>
--- a/src/elements/calendar/calendar-day/calendar-day.component.ts
+++ b/src/elements/calendar/calendar-day/calendar-day.component.ts
@@ -85,10 +85,7 @@ export class SbbCalendarDayElement<T = Date> extends SbbCalendarCellBaseElement<
   }
 
   private _isDayInRange(min: T | null, max: T | null): boolean {
-    if (!this.value) {
-      return true;
-    }
-    if (!min && !max) {
+    if (!this.value || (!min && !max)) {
       return true;
     }
     return this.dateAdapter.sameDate(this.value, this.dateAdapter.clampDate(this.value, min, max));
--- a/src/elements/calendar/calendar-day/calendar-day.spec.ts
+++ b/src/elements/calendar/calendar-day/calendar-day.spec.ts
@@ -29,6 +29,23 @@ describe('sbb-calendar-day', () => {
     assert.instanceOf(todayElement, SbbCalendarDayElement);
   });
 
+  it('does not crash when parent min/max updates before slotted day value is initialized', async () => {
+    root = await fixture(html`
+      <sbb-calendar>
+        <sbb-calendar-day></sbb-calendar-day>
+      </sbb-calendar>
+    `);
+    const min = defaultDateAdapter.createDate(year, month, 10);
+    const max = defaultDateAdapter.createDate(year, month, 20);
+
+    root.min = min;
+    root.max = max;
+    await waitForLitRender(root);
+    const day = root.querySelector<SbbCalendarDayElement>('sbb-calendar-day');
+    expect(day).to.exist;
+    expect(day?.disabled).to.be.false;
+  });
+
   it('should have the correct properties on todayElement', async () => {
     expect(todayElement).to.match(':state(current)');
     expect(todayElement).not.to.match(':state(crossed-out)');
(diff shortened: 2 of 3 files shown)


src/elements/calendar/calendar-day/calendar-day.component.ts
```diff
@@ -85,6 +85,9 @@ export class SbbCalendarDayElement<T = Date> extends SbbCalendarCellBaseElement<
   }
 
   private _isDayInRange(min: T | null, max: T | null): boolean {
+    if (!this.value) {
+      return true;
+    }
     if (!min && !max) {
```
TomMenga:
```suggestion
    if (!this.value || (!min && !max)) {
```

src/elements/calendar/calendar-day/calendar-day.spec.ts
```diff
@@ -29,6 +29,23 @@ describe('sbb-calendar-day', () => {
     assert.instanceOf(todayElement, SbbCalendarDayElement);
   });
 
+  it('does not crash when parent min/max updates before slotted day value is initialized', async () => {
```
TomMenga: I will refactor this a bit to comply with our best practices

TomMenga (changes requested):
Hello Roland, thank you a lot for your contribution :)

Out of curiosity, how did you find yourself in that misconfigured case? 
Is it an initialization timing issue? Are you in an Angular environment, perhaps?

rolego:
> Hello Roland, thank you a lot for your contribution :)
> 
> Out of curiosity, how did you find yourself in that misconfigured case? Is it an initialization timing issue? Are you in an Angular environment, perhaps?

Dear @TomMenga,

Thanks for the follow-up! Yes, we're using lyne-angular. The issue I'm trying to solve is that disabled days aren't clearly distinguishable from enabled ones, so I tried rendering the days myself to have more control over the styling.

In hindsight, opening an issue first to discuss the approach would probably have been the better path — happy to do that if it's more useful going forward, and I'm open to a different solution if you have one in mind.

TomMenga:
> Thanks for the follow-up! Yes, we're using lyne-angular. The issue I'm trying to solve is that disabled days aren't clearly distinguishable from enabled ones, so I tried rendering the days myself to have more control over the styling.

It's a bit of a workaround, but did you have a look at the `dateFilter` function? ([storybook](https://lyne-storybook.app.sbb.ch/?path=/story/elements-calendar--calendar-filter-function)). Filtered days are still disabled but with a clearer visual state, maybe that could do for you.

In any case, your PR improves robustness, so I'd like to merge it anyway

> In hindsight, opening an issue first to discuss the approach would probably have been the better path — happy to do that if it's more useful going forward, and I'm open to a different solution if you have one in mind.

That's absolutely fine, don't worry :) 
If it's okay for you, I can commit the test refactoring directly on your branch (if I'm allowed to do so)

rolego:
> > Hello Roland, thank you a lot for your contribution :)
> > Out of curiosity, how did you find yourself in that misconfigured case? Is it an initialization timing issue? Are you in an Angular environment, perhaps?
> 
> Dear @TomMenga,
> 
> Thanks for the follow-up! Yes, we're using lyne-angular. The issue I'm trying to solve is that disabled days aren't clearly distinguishable from enabled ones, so I tried rendering the days myself to have more control over the styling.
> 
> In hindsight, opening an issue first to discuss the approach would probably have been the better path — happy to do that if it's more useful going forward, and I'm open to a different solution if you have one in mind.

Thank you for the suggested workaround.

I already tried this. The problem with this is that the combination with disabled (min/max) does not work which is important to us, since we want to limit the date range the user can navigate at all. 
I'll open another issue for the styling of disabled days for further discussions...

rolego:
See also: https://github.com/sbb-design-systems/lyne-components/issues/5124

TomMenga:
> #5124

For this, we will need to wait for the opinion of the UX team, which is currently unavailable 😔

TomMenga (approved):
Legit to me 👍 
Feel free to merge it once all the actions are green 😄

PR #5128, opened 2026-07-29
fix(date-input, time-input): avoid setCursorAt exception

### LINKED ISSUE #5044: [Bug]: Uncaught TypeError: Failed to execute 'setStart' on 'Range': parameter 1 is not of type 'Node'.
### Preflight Checklist

### Bug type

Other

Components

4.12.0

### Operating system

win 10

### Browser / Browser version

Chrome 149

### Input mode

_No response_

### Additional settings

_No response_

### What happened?

In our production monitoring we have a handful of instances of the below error:

```
Uncaught TypeError: Failed to execute 'setStart' on 'Range': parameter 1 is not of type 'Node'.
```
With stacktrace:
```
node_modules/@sbb-esta/lyne-elements/core/mixins/form-associated-input-mixin.js at 188:13
node_modules/@sbb-esta/lyne-elements/core/mixins/form-associated-input-mixin.js at 116:211
```

---
Preliminary AI Analysis points to a bug in this lib:

Root cause: library bug in SbbFormAssociatedInputMixin cursor handling for empty contenteditable inputs.

The failing call is effectively:
range.setStart(this.firstChild!, position);
in src/elements/core/mixins/form-associated-input-mixin.ts:449.

At runtime this.firstChild can be null, so the browser receives:
range.setStart(null, 0)
and throws:
Failed to execute 'setStart' on 'Range': parameter 1 is not of type 'Node'

Why it happens:
- _setCursorAt(0) assumes the component has a text child.
- But _assignValue('') sets innerHTML = '', removing all children.
- The empty-input mobile/coarse-pointer workaround can then call _setCursorAt(0) while the element is still empty.
- The stack’s second frame likely corresponds to the keyup path:
} else if (this._requiresEmptyPatch()) {
  this._setCursorAt(0);
}

Affected components:
- sbb-date-input
- sbb-time-input

Likely production trigger:
- Empty date/time input
- Mobile/tablet/coarse pointer environment, especially Blink/Chrome Android
- User interaction such as keyup/backspace/delete/touch-related focus handling while the contenteditable host has no child node

Not the primary root cause: consumer code. This is an unsafe internal non-null assertion in Lyne’s mixin. The fix should ensure _setCursorAt() creates or reuses a text node before calling Range.setStart, or bail out when no node exists.
 

1. Step one
2. Step two
3. ...

### Relevant log output

```shell

```

DavideMininni-Fincons:
Bug #5044 has been replicated on Chrome with mobile emulation and it's caused by the keyup listener;
a possible tested solution was add a nbsp as done in the touchend listener, but it causes the placholder to flicker.

Simply checking that the firstChild is available solves the issue; however we do not have specific tests for the _requiresEmptyPatch function, which is Blink-dependant

### DIFF
--- a/src/elements/core/mixins/form-associated-input-mixin.ts
+++ b/src/elements/core/mixins/form-associated-input-mixin.ts
@@ -425,12 +425,12 @@ export const SbbFormAssociatedInputMixin = <
 
     private _setCursorAt(position: number): void {
       const selection = window.getSelection();
-      if (!selection) {
+      if (!this.firstChild || !selection) {
         return;
       }
 
       const range = document.createRange();
-      range.setStart(this.firstChild!, position);
+      range.setStart(this.firstChild, position);
       range.collapse(true);
       selection.removeAllRanges();
       selection.addRange(range);


TomMenga (approved):
LGTM 👍

PR #5135, opened 2026-08-03
fix(popover): improve positioning for triggers near viewport boundaries

### LINKED ISSUE #5116: [Bug]: SbbPopover overflows viewport near the right edge
### Preflight Checklist

### Bug type

Functionality

Components

5.0.2

### Operating system

macOS

### Browser / Browser version

Arc (Chromium)

### Input mode

None

### Additional settings

_No response_

### What happened?

SbbPopover currently provides limited control over its horizontal positioning. In layouts where the trigger is close to a viewport edge, consumers cannot configure the offset, alignment, or viewport inset.

The popover should be constrained to the viewport from the first rendered frame, while its arrow remains aligned with the trigger.

Move the trigger closer to the edge of the screen. In our case, that would be 16px; in the example below, we used 20px. In my opinion, either spacing would be reasonable on mobile screens, and the popover should be able to handle both correctly.
https://lyne-storybook.app.sbb.ch/?path=/story/elements-popover--end-below&globals=mode:light
Image

### Relevant log output

```shell

```

jeripeierSBB:
Closes #5116

### DIFF
--- a/src/elements/popover/popover-base/popover-base.ts
+++ b/src/elements/popover/popover-base/popover-base.ts
@@ -29,6 +29,8 @@ import style from './popover-base.scss?inline';
 
 const VERTICAL_OFFSET = 16;
 const HORIZONTAL_OFFSET = 32;
+const VIEWPORT_MARGIN = 2;
+const ARROW_SIZE = 16; // full width of the popover arrow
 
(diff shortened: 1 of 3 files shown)


TomMenga (approved):
LGMT 👍

PR #5148, opened 2026-08-05
build: improve manifests generation

DavideMininni-Fincons:
On the manifest generation script, a parameter has been added which allows the loading on an external manifest.


jeripeierSBB (approved):
Very nice! LGTM

PR #5149, opened 2026-08-05
docs: fix broken escaping of html elements

jeripeierSBB:


DavideMininni-Fincons (approved):
LGTM, thanks!
I remember I fixed a single similar case before the v5 release, but i didn't check for other occurencies

PR #5150, opened 2026-08-05
fix(stepper): enable steps switching from linear to non-linear

### LINKED ISSUE #5133: bug(stepper): switch to non-linear keeps steps disabled
### Feature Description

https://lyne-elements.app.sbb.ch/?path=/story/elements-stepper--default
Manually adding a 'linear' attribute to the 'sbb-stepper' disables all the steps except the first; this is done calling the '_configureLinearMode' method on stepper which calls the internal 'disable' method on the label.

In theory, this should avoid calling directly the disable setter, which adds a 'user-disabled' state; however, the setter on the disabledMixin superclass toggles the attribute, which results in the label's disabled setter to be called anyway.

As a result, all the steps are marked as 'user-disabled', so when the 'linear' is removed, they stay off.

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

DavideMininni-Fincons:
Closes bug #5133.
Since the issue has not an immediate fix, Copilot was used to gather ideas.

### DIFF
--- a/src/elements/stepper/step-label/step-label.component.ts
+++ b/src/elements/stepper/step-label/step-label.component.ts
@@ -39,7 +39,9 @@ export class SbbStepLabelElement extends SbbIconNameMixin(SbbDisabledMixin(SbbBu
 
     // We additionally keep track of the `disabled` state to preserve the user configured disabled state
     // of step labels in case of switching between linear and non-linear mode.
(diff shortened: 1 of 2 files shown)


jeripeierSBB (approved):
LGTM, thank you!

PR #5152, opened 2026-08-06
style(icon): reserve space when changing name

jeripeierSBB:
Currently, when changing the icon name, e.g. if the type of the `sbb-status` changes, the width collapses. With this solution the empty state (which was only set in the constructor) is now set before every reload again.


DavideMininni-Fincons (approved):
LGTM, thanks!

PR #5158, opened 2026-08-10
fix(dialog): ignore consumer set dimension on mobile view

DavideMininni-Fincons:
Part of https://github.com/sbb-design-systems/lyne-angular/pull/551


jeripeierSBB (approved):
LGTM, but please also try out my new solution :P

jeripeierSBB (approved):
LGTM, thank you!

PR #5193, opened 2026-08-24
chore: add agents.md instructions

TomMenga:
I've tested it for a while, and it adds some value when using AI. 
I'm also testing a few skills that improve code quality and token consumption (but that's for another time). Let's start with this for now 😄


kyubisation (approved):
Looks like a good starting point
Thanks! 😄

PR #5205, opened 2026-08-28
feat(calendar): add `activeMonth` property

### LINKED ISSUE #5168: [Contributing request]: Set the initially displayed month of `sbb-calendar` without locking navigation
### Preflight Checklist

### Contribution type

Contribute an enhancement of a component

Components

### Feature Description

`<sbb-calendar>` can be told *which month to display*, but only in ways that have an unwanted side effect: either a day gets selected, or month navigation is switched off. We would like a public input that sets the initially displayed month and nothing else.

### Current behaviour

There are two ways to influence which month is shown:
Property | Sets the month | Side effect
-- | -- | --
value | yes | selects that day
fixedMonth | yes | removes the navigation controls; the month can never change

Without either, the calendar always opens on the current month. Since the calendar renders `amount` months **forward** from that month, the past is never visible.

### Problem
Applications often need to open the calendar somewhere other than the current month while leaving the user free to browse. Two common situations:

1. **Opening on a past month.** A calendar that should show a window around today, for example a few days before and after, cannot do so at the start of a month, because everything before the 1st lies in the previous month and the calendar only renders forward.
2. **Data that does not include today.** When only a limited period contains selectable dates (a booking season, a published schedule, an archived period) and today is outside it, the calendar opens on months where nothing is selectable. The user has to page back and forth to find the data, and with `amount="2"` that can be several clicks.

In both cases the month is known to the application, and in both cases the user must still be able to navigate afterwards. That combination is not expressible today.

### Why the existing properties don't solve it
**`fixedMonth`** positions correctly, but by design it disables navigation, so it only fits read-only, single-month displays.
**`value`** cannot be used as a positioning trick, for three reasons:
- It selects a day. Preselecting something the user did not choose is often wrong (and can trigger application logic tied to the selection).

- The setter silently drops any date that `dateFilter` rejects:
```JavaScript
if (!!selectedDate && (!this._isDayInRange(iso) || this._dateFilter(selectedDate)))
  this._value = selectedDate;
else this._value = null;

```
The month you want to open on frequently contains **no** selectable day, which is often the very reason for wanting to control the position. In that case `value` is discarded and the calendar falls back to the current month.

- It only positions on connect. `willUpdate` re-runs `resetPosition()` for `amount`, `orientation`, `fixedMonth` and `view`, but never for `value`, so setting `value` later does not move the view. Applications that load their data asynchronously only know the target month after the first render.

### Proposal
A public input for the initially displayed month, independent of selection and navigation:

```html
<sbb-calendar amount="2" startMonth="2026-06"></sbb-calendar>
```
- Format as in `fixedMonth` (`YYYY-MM`), or as in `value` (date object / ISO string / timestamp).
- The calendar renders `amount` months forward from it, exactly like the existing anchor.
- Navigation controls stay visible; the user can browse to any month.
- No day is selected, and the value is not subject to `dateFilter`.
- Changing it later repositions the view, so it also works with asynchronously loaded data.

Conceptually this is `fixedMonth` without the lock: the positioning half of it, made reusable.

### Current workaround
We call the private `_init(anchor)` on the element from a wrapper component. This is the same method the built-in previous/next buttons use, so the behaviour is correct, but it depends on internals that are not covered by semver and may break on any release.

### Open questions
- Should `startMonth` be clamped by `min`/`max`?
- What takes precedence if `fixedMonth` is also set — should the two be mutually exclusive?
- What takes precedence if `value` is set: `startMonth` for positioning, `value` for selection?
- Should changing `startMonth` at runtime emit `monthchange`?
- Is `startMonth` the preferred name, or would `initialMonth` / `activeMonth` fit the API better?

_No response_

### Please describe the required properties (with type and value) for the component

startMonth (string): "YYYY-MM" — e.g. "2026-06"; default: null (current behaviour, calendar opens on `value` or today)

### Please send a proposal for a html markup

```html
<sbb-calendar amount="2" start-month="2026-06"></sbb-calendar>
```

### LINKED ISSUE #5169: [Feature request]: Calendar – control the initially displayed month independently of the selected value
### Preflight Checklist

### Request type

Components

### Functionality

**Component:** `sbb-calendar` (used via `@sbb-esta/lyne-angular` / `@sbb-esta/lyne-elements`)

#### Current behaviour

With `view="month"` and no value set, the calendar initially renders the current month. As soon as a value (single date or range) is set, the calendar takes that value as the starting point and renders the month of the selected date instead.

Internally the displayed period is held in the private state `_activeDate`, which is initialised from `today()` but then derived from `value`. There is no public way to read or set it. The only related public API today is `fixedMonth`, which pins the calendar to one month but also **disables navigation** entirely — too restrictive for our use case.

#### Our requirement

In our application (ZLD Wartefristen Tool) the calendar should always open on the *current* month on first rendering, independently of the values that are already selected. Selected values may lie far in the past or future, but the user's working context is "today". Currently the only way to get the current month is to not set a value at all, which we cannot do because the selection has to stay visible.

Beyond the initial rendering we would also like to **read and control the calendar navigation at runtime** — e.g. a "Jump to today" button next to the calendar, or syncing two calendars / an external month picker with the calendar's displayed period.

---

### Proposal: a public API to control calendar navigation

The goal is: make the *displayed period* a first-class, controllable concept, independent of the *selected value*.

**1. `activeDate` — public, two-way property (main request)**

Expose the existing private `_activeDate` state as a public property/attribute:

```ts
/**
 * The date whose period (month/year) is currently displayed.
 * Defaults to `value` (first/last selected date) or `today()` when unset.
 * Unlike `fixedMonth`, navigation stays fully enabled.
 */
@plainDate()
@property({ attribute: 'active-date' })
public accessor activeDate: T | null = null;
```

Semantics:
- If set by the consumer, it wins over the value-derived starting point.
- If `null` / not set, current behaviour is preserved (fully backwards compatible).
- Clamped to `min` / `max` (and ignored/overridden when `fixedMonth` is set).
- Setting it at runtime navigates the calendar to that period.

Usage:

```html
<!-- always open on the current month, even though a value is set -->
<sbb-calendar active-date="2026-08-14" .value=${selectedDates}></sbb-calendar>
```

```ts
// Angular: "jump to today"
this.calendar.nativeElement.activeDate = new Date();
```

**2. Read the current state**

- `activeDate` is readable at any time (getter reflects the internally navigated date, not only what was set).
- Extend the existing `monthchange` event (or add an `activedatechange` event) so that the emitted detail also carries the new active date / period, not only `visibleDays()`. Today consumers have to derive the displayed month from the days array.

```ts
calendar.addEventListener('monthchange', (e) => {
  console.log(calendar.activeDate); // Date of the newly displayed period
});
```

Combined, `activeDate` + the change event give a proper controlled/uncontrolled pattern (like `value` + `dateselected`), and it maps cleanly onto an Angular two-way binding `[(activeDate)]`.

**3. Optional: imperative navigation helpers**

Nice-to-have sugar on top of `activeDate`, mirroring what the internal navigation buttons already do:

```ts
calendar.navigateTo(date);        // set the displayed period
calendar.navigateToToday();       // convenience for the most common case
calendar.next();  calendar.previous();  // same as the header arrows
```

**4. Alternative**

If a full `activeDate` API is considered too much, a `now`-style input (as already available on `sbb-clock`) that overrides the reference date used for the initial rendering would already solve our primary requirement — but it would not cover the runtime read/update part described above.

#### Context

This was discussed with the Lyne team in the Teams channel; the suggestion was to open a feature request so a solution (e.g. a `now` parameter similar to the clock component) can be discussed. Happy to contribute a PR if you agree on the direction.

_No response_

kyubisation:
The `activeMonth` property allows configuring the  displayed month.

Closes #5168, closes #5169


src/elements/calendar/calendar/calendar.component.ts
```diff
@@ -553,6 +585,13 @@ export class SbbCalendarElement<T = Date> extends SbbFormAssociatedMixin(SbbElem
     this._initialized = true;
```
jeripeierSBB, line 585: Would there be a more suitable place to set initialized? now it's done everytime, but it's also no problem :)

src/elements/calendar/readme.md
```diff
@@ -289,9 +289,10 @@ For accessibility purposes, the component is rendered as a native table element
```
jeripeierSBB, line 300: As there is already a section "fixed month" in the readme, I'd also add a section "active month", so both is explained with the same weight.

jeripeierSBB (approved):
Good enhancement, LGTM!

src/elements/calendar/calendar/calendar.component.ts
kyubisation: Not sure. As it depends on whether the data has been created, I would leave it here.

jeripeierSBB: Let's keep it

jeripeierSBB (approved):
A second LGTM :)

PR #5206, opened 2026-08-31
feat(autocomplete): implement 'actions' in options

### LINKED ISSUE #4322: story(sbb-autocomplete-grid): improve accessibility
### Feature Description

Basically we have to improve the accessibility of the sbb-autocomplete-grid by introducing a different approach.
Once more we can take a look at https://music.youtube.com/ and try to adapt it.

#### Implementation suggestion (to be validated):

If this suggestion works, it can be implemented directly in the existing autocomplete module.

```html
<sbb-autocomplete>
  <sbb-autocomplete-row>
    <sbb-option>Example 1</sbb-option>
    <sbb-autocomplete-button icon-name="pen-small"></sbb-autocomplete-button>
  </sbb-autocomplete-row>
  <sbb-option>Example 2</sbb-option>
  ...
</sbb-autocomplete>
```

Things to consider:

- Keyboard navigation
- Deprecate the autocomplete grid components

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

TomMenga:
This PR Closes #4322

Follow-up: Update Angular wrappers and add example


src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
```diff
@@ -0,0 +1,102 @@
...
+        }),
+      );
+    }
+  }
+
+  /**
+   * Whether the button is currently active.
+   * @internal
+   */
+  public setActive(value: boolean): void {
```
jeripeierSBB, line 59: not sure whether we usually stick to protected methods in this case?

src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.toggleState('focus-visible', value);
+  }
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    this.id ||= `sbb-autocomplete-button-${++autocompleteButtonNextId}`;
```
jeripeierSBB: Do we really need an id (maybe later in the review I'll find out :P)

src/elements/autocomplete/autocomplete-row/autocomplete-row.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    );
+  }
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    this.id ||= `sbb-autocomplete-row-${++autocompleteRowNextId}`;
+  }
+
+  protected override render(): TemplateResult {
+    return html` <slot></slot> `;
```
jeripeierSBB:
```suggestion
    return html`<slot></slot>`;
```

src/elements/autocomplete/autocomplete-row/autocomplete-row.component.ts
```diff
@@ -0,0 +1,49 @@
...
+    this.addController(
+      new SbbPropertyWatcherController(this, () => this.querySelector('sbb-option'), {
+        disabled: (e) => this.toggleState('disabled', e.disabled),
+      }),
+    );
+  }
+
+  public override connectedCallback(): void {
+    super.connectedCallback();
+    this.id ||= `sbb-autocomplete-row-${++autocompleteRowNextId}`;
```
jeripeierSBB: Also here, ID needed?

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
```diff
@@ -0,0 +1,70 @@
+@use '../../core/styles' as sbb;
+
+:host {
+  --sbb-autocomplete-row-padding-inline-end: var(--sbb-spacing-responsive-xxxs);
+  --sbb-autocomplete-row-gap: var(--sbb-spacing-fixed-6x);
```
jeripeierSBB: Variables can probably be moved to global?

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  @media (forced-colors: active) {
+    --sbb-option-color: GrayText;
+  }
+
+  &::before {
+    content: '';
+    display: block;
+    position: absolute;
+    inset: #{sbb.px-to-rem-build(6)};
```
jeripeierSBB: I'd say we should create (or use existing? ) token here.

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
```diff
@@ -0,0 +1,70 @@
...
+    border: var(--sbb-border-width-1x) dashed var(--sbb-option-disabled-border-color);
+    border-radius: var(--sbb-border-radius-2x);
```
jeripeierSBB: I think we should either create a token or use calc (e.g. subtract inset from border radius var (if there is any))

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
```diff
@@ -0,0 +1,70 @@
...
+    background-color: var(--sbb-option-disabled-background-color);
+    z-index: -1;
+
+    @media (forced-colors: active) {
+      border-color: GrayText;
```
jeripeierSBB: maybe better create a variable in the block above?

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
```diff
@@ -0,0 +1,70 @@
+@use '../../core/styles' as sbb;
```
jeripeierSBB, line 1: is there any option that it also works without z-index? e.g. with grid

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
```diff
@@ -0,0 +1,70 @@
...
+      border-color: GrayText;
+    }
+  }
+}
+
+::slotted(sbb-option) {
+  --sbb-option-background-color-active: var(--sbb-option-background-color-hover);
+
+  flex: 1 1 auto;
+  margin-right: calc(-1 * var(--sbb-spacing-fixed-2x));
```
jeripeierSBB, line 59: does that mean after the option there is an additional gap, so in total 8x spacing between option and first button? I'd add a comment and maybe also a token

src/elements/autocomplete/autocomplete/autocomplete-base-element.scss
```diff
@@ -1,4 +1,4 @@
-@use '../core/styles' as sbb;
+@use '../../core/styles/index' as sbb;
```
jeripeierSBB:
```suggestion
@use '../../core/styles' as sbb;
```

src/elements/autocomplete/autocomplete/autocomplete-base-element.ts
```diff
@@ -26,12 +26,14 @@ import {
...
-} from '../core.ts';
-import type { SbbFormFieldElement } from '../form-field/form-field/form-field.component.ts';
-import { optionPanelStyles, type SbbOptionBaseElement } from '../option.pure.ts';
+} from '../../core.ts';
+import type { SbbFormFieldElement } from '../../form-field/form-field/form-field.component.ts';
+import { optionPanelStyles, type SbbOptionBaseElement } from '../../option.pure.ts';
 
 import style from './autocomplete-base-element.scss?inline';
 
+// TODO: The base class will no longer be needed once the autocomplete-grid is deleted in the next major release. Merge this in the main class.
```
jeripeierSBB:
```suggestion
// TODO(breaking-change): The base class will no longer be needed once the autocomplete-grid is deleted in the next major release. Merge this in the main class.
```

src/elements/autocomplete/autocomplete/autocomplete.component.ts
```diff
@@ -97,7 +115,60 @@ export class SbbAutocompleteElement<T = string> extends SbbAutocompleteBaseEleme
...
+    if (!this.activeOption) {
+      return;
+    }
+
+    const elementsInRow = this._elementsInRow();
+
+    if (elementsInRow.length < 2) {
+      return;
+    }
+    event.preventDefault();
```
jeripeierSBB, line 135: should we document why the preventDefault() is needed?

src/elements/autocomplete/autocomplete/autocomplete.component.ts
```diff
@@ -97,7 +115,60 @@ export class SbbAutocompleteElement<T = string> extends SbbAutocompleteBaseEleme
...
+
+    const nextIndex = getNextElementIndex(event, this._activeColumnIndex, elementsInRow.length);
+    const current = elementsInRow[this._activeColumnIndex];
+    const next = elementsInRow[nextIndex];
+
+    current.setActive(false);
+    next.setActive(true);
+
+    this.triggerElement!.ariaActiveDescendantElement = next;
```
jeripeierSBB: I know it shouldn't be needed, but be more safe for any of special situation we never could imagine :P
```suggestion
    if(this.triggerElement) {
      this.triggerElement.ariaActiveDescendantElement = next;
    }
```

src/elements/autocomplete/autocomplete/autocomplete.visual.spec.ts
```diff
@@ -320,6 +356,63 @@ describe('sbb-autocomplete', () => {
...
+            { ...style, backgroundColor: 'var(--sbb-background-color-1-negative)' },
+          );
+          setup.withPostSetupAction(async () => {
+            await openAutocomplete(setup);
+            await sendKeys({ press: 'ArrowDown' });
+          });
+        }),
+      );
+
+      it(
```
jeripeierSBB, line 424: I'd also add a test the triggers an arrowright to show the ring of the button

jeripeierSBB (commented):
Really nice and clean refactoring. I hope it will withstand the A11Y review :) I was wondering whether we need some hints that there is an action on the sbb-option, but apparently there is also no on youtube music too (did only a quick search). So just let's see what the review brings :)

src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
```diff
@@ -0,0 +1,102 @@
...
+    this.removeAttribute('tabindex');
```
kyubisation: What sets the tabindex? I think the currently used mixins do not set the tabindex.

src/elements/chip/chip-group/chip-group.component.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from 'lit';
 import { property } from 'lit/decorators.js';
 
-import type { SbbInputAutocompleteEvent } from '../../autocomplete/autocomplete-base-element.ts';
+import type { SbbInputAutocompleteEvent } from '../../autocomplete/autocomplete/autocomplete-base-element.ts';
```
kyubisation: Nitpick; This import is odd and, if possible, should be changed to the autocomplete pure import.

src/elements/chip/chip-group/chip-group.spec.ts
```diff
@@ -2,7 +2,7 @@ import { assert, aTimeout, expect } from '@open-wc/testing';
 import { sendKeys } from '@web/test-runner-commands';
 import { html } from 'lit/static-html.js';
 
-import type { SbbAutocompleteElement } from '../../autocomplete/autocomplete.component.ts';
+import type { SbbAutocompleteElement } from '../../autocomplete/autocomplete/autocomplete.component.ts';
```
kyubisation: Nitpick; This import is odd and, if possible, should be changed to the autocomplete pure import.

kyubisation (commented):
Excellent work!

src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
TomMenga: Inherited from the autocomplete-grid-button :/ Apparently, it wasn't needed anymore

src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
TomMenga: In case I would also change `optionBase.setActive` to protected. I'm fine with both solutions btw, slightly in favour of `@internal`

src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
TomMenga: I still forget that we are using internals 😅

src/elements/autocomplete/autocomplete-row/autocomplete-row.scss
TomMenga: It's a negative margin that partially compensates for the option padding. 
It's something coming from the autocomplete-grid and I assumed it was requested at some point in a visual review

src/elements/autocomplete/autocomplete-button/autocomplete-button.component.ts
jeripeierSBB: Let's keep it synced with options

jeripeierSBB (approved):
LGTM, great!

TomMenga:
### a11y review

1. Operating instructions are missing
When the input field is reached, the screen reader does not issue a message that these elements can be deleted and possibly edited. Thus, blind users cannot necessarily perceive and reach these elements. An operating message should be issued that these functions can be accessed with the arrow keys left and right.
(I haven't checked yet whether it should happen via aria-label, or rather other way..)
 
2. Delete and Edit buttons are not labeled
 
3. with NVDA, the component does not always work correctly
Without a screen reader, the component can be operated flawlessly with the keyboard. But with screen reader (NVDA) together, it sometimes doesn't navigate any further, but focuses on the whole pop-up. Even the last element is not always accessible (but sometimes it is). I didn't find any problems with Jaws.
 
4. Clock icon for "Last Visited" is not displayed
 
5. In mobile view, the longer place names are truncated
 
6. Question:
In this example, no suggestions are displayed. But I think otherwise the search would work the same way as it does now, e.g. 10 suggestions are listed, right?

PR #5207, opened 2026-08-31
feat(teaser): add new size, alignment and static options

### LINKED ISSUE #4914: story(teaser): new property "size", new text-orientation variants and other improvements
### Feature Description

- a new prorperty "size" is needed (default: M, new: L)
- It should be possible to place a button (static variant) 
- new variants are needed: "before" and "before-centred"
- a static variant is needed (without href)

design-spec: https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?node-id=15302-313111&t=ZyxdFX1ybUTZhNjH-11

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

jeripeierSBB:
Closes #4914


src/elements/teaser/common/teaser-common.ts
```diff
@@ -0,0 +1,102 @@
...
+            foundTitle = true;
+          }
+        } else if (
+          /^sbb-(secondary-|accent-|transparent-)?button(-static|-link)?$/.test(child.localName)
+        ) {
+          if (!child.slot) {
+            child.slot = 'action';
+          }
+        }
+      }
```
kyubisation, line 54: More a question that a direct suggestion. Why did you opt toward iteration?

```suggestion
      const chipLabel = this.querySelector(':scope > sbb-chip-label:not([slot])');
      if (chipLabel) {
        chipLabel.slot = 'chip';
      }
      
      const title = this.querySelector(':scope > sbb-title:not([slot])');
      if (title) {
        title.slot = 'title';
      }
      
      const action = this.querySelector(':scope > ...:not([slot])');
      if (action) {
        action.slot = 'action';
      }
```

kyubisation (approved):
LGTM 👍

src/elements/teaser/common/teaser-common.ts
jeripeierSBB: I think the proposed solution is not idempotent, like for a follow up slot change another label would also be assigned with the slot and then the next and so on. I tried to preserve the existing logic by only apply the slot the the very first  label and title, regardless wether the slot was already assigned or not. with that you could also have other chips in the content. although, the overall solution is for sure discussable :)
the reason for the `for` was that before there were two `find` operations and I thought instead of adding one more, I just iterate once over every element (performance, making @kyubisation happy :-) ).

kyubisation: Took me a moment to understand, but you are right 😄

src/elements/teaser/teaser-static/teaser-static.component.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Non-interactive variant of the `<sbb-teaser>`.
+ *
+ * @slot action - Slot for an interactive action, e.g. a `<sbb-secondary-button-link>` element. The action is displayed below the description.
+ * @slot chip - Slot for the `sbb-chip-label` element. The slot on the `sbb-chip-label` element is automatically assigned when slotted in the unnamed slot.
+ * @slot image - Slot used to render the image.
+ * @slot title - Slot for the title. For the standard `sbb-title` element, the slot is automatically assigned when slotted in the unnamed slot.
+ * @slot - Use the unnamed slot to render the description, the sbb-title and the sbb-chip-label.
+ */
+export class SbbTeaserStaticElement extends SbbTeaserCommonElementMixin(SbbActionBaseElement) {
```
TomMenga, line 16: I'm wondering whether it makes sense for a static component to extend `SbbActionBaseElement` 🤔 
Probably you were forced to extend it to have a common base for the link variant, but it still seems a bit flawed

TomMenga (approved):
LGTM 👍

src/elements/teaser/teaser-static/teaser-static.component.ts
jeripeierSBB: Yeah it's kind of an alignment with other static components that also extend the actionbaseelement and also that it fits with types in the common mixin. Probably not a hard requirement, but as written, at least an aligned state and an easy way to have the link teaser and static teaser share almost all of the same logic.

PR #5208, opened 2026-09-01
fix(icon): avoid caching failed requests

### LINKED ISSUE #5187: [Feature request]: Do not cache empty or failed icon requests
### Preflight Checklist

### Request type

Components

### Functionality

The sbb-icon component caches icon-loading promises by URL. Currently, an empty or failed request can remain cached for the lifetime of the page. This means that a temporary CDN or network failure may prevent subsequent instances of the same icon from retrying the request, leaving the icon empty.

_No response_

DavideMininni-Fincons:
Closes #5187 

A try...catch has been used instead than the proposed IIFE, since it doesn't introduces an extra async step, which can lead to a timing regression.

### DIFF
--- a/src/elements/icon/icon-request.ts
+++ b/src/elements/icon/icon-request.ts
@@ -10,7 +10,7 @@ const iconCdn = 'https://icons.app.sbb.ch/';
 const iconNamespaces = new Map<string, string>()
   .set('default', `${iconCdn}icons/`)
   .set('picto', `${iconCdn}picto/`);
-const requests = new Map<string, Promise<any>>();
+const requests = new Map<string, Promise<string>>();
 
 /** Fetches icon svg content from providers and asserts only one request per icon is made. */
 export const getSvgContent = (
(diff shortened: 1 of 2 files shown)


src/elements/icon/icon-request.ts
```diff
@@ -37,23 +37,41 @@ export const getSvgContent = (
...
+          return content;
+        },
+        (error: unknown) => {
+          if (requests.get(url) === req) {
+            requests.delete(url);
+          }
+          console.warn(error);
+          return '';
+        },
+      );
```
jeripeierSBB, line 74: I think (not tested) a Promise chain could be used here as slightly nicer solution than the extra promise.

```suggestion
   req = Promise.resolve()
  .then(() =>
    interceptor({
      namespace,
      name,
      url,
      request: async () => {
        const response = await fetch(url);

        if (!response.ok) {
          throw new Error(`Failed to load icon ${namespace}:${name}`);
        }

        return validateContent(await response.text(), sanitize);
      },
    }),
  )
  .then(
    (content) => {
      if (!content && requests.get(url) === req) {
        requests.delete(url);
      }

      return content;
    },
    (error: unknown) => {
      if (requests.get(url) === req) {
        requests.delete(url);
      }

      console.warn(error);
      return '';
    },
  );
```

jeripeierSBB (approved):
LGTM, please also await a review of @kyubisation.

src/elements/icon/icon-request.ts
```diff
@@ -37,23 +37,41 @@ export const getSvgContent = (
...
+        );
+      } catch (error) {
+        interceptorRequest = Promise.reject(error);
+      }
+
+      req = interceptorRequest.then(
+        (content) => {
+          if (!content && requests.get(url) === req) {
+            requests.delete(url);
+          }
```
kyubisation, line 64: If the request returned `ok`, but the content is empty, then it is questionable that a future request will return something else.
I think we can reduce this to just a `catch`.

kyubisation (approved):
Generally LGTM
I think @jeripeierSBB suggestion is a bit of a cleaner API, but it doesn't matter all that much
Thanks!

src/elements/icon/icon-request.ts
DavideMininni-Fincons: The extra Promise at top-level has the same issue of the original proposed solution: it's an extra async step, which can cause a timing issue (I had a failing test on `icon.snapshot.spec.ts` due to a not-yet-loaded icon when the name prop is changed, even if `waitForLitRender` was called)

jeripeierSBB: I understand, let's just keep your solution :)

src/elements/icon/icon-request.ts
DavideMininni-Fincons: I would prefer to keep it to cover edge cases: 

1. the `interceptor` object can be overridden by consumers via config, and can have a different error management, eg.
    ```
    interceptor: async ({ request }) => {
      try {
        return await request();
      } catch {
        return '';
      }
    }
    ```
    the possible failures can be recovered later (cache should be emptied);

2. since the method is called also for custom namespaces, a malformed SVG can be loaded by consumers; in that case there could be a 200 code with an empty string returned by the sanitizer. Also here, the svg can be fixed later and in order to work cache should be empty.

kyubisation: Ok, that's reasonable. 👍

PR #5209, opened 2026-09-01
fix(dialog): various mobile view optimizations

jeripeierSBB:
Improvements in this PR:
 - blocking body does no longer cut the background, if there is any background color set
 - safe area of mobile applications is respected
 - backdrop fade in also works on first time
 - scrolling of background with open dialog in iOS is prevented
 - minor CSS optimizations
 - Shadow of dialog actions and dialog title correctly cut so they do not bleed behind the dialog

Relates to https://github.com/sbb-design-systems/lyne-angular/issues/562


DavideMininni-Fincons (approved):
LGTM, thanks!

kyubisation (approved):
LGTM 👍

PR #5223, opened 2026-09-04
fix(form-field): increase label space for textarea in 's' size

### LINKED ISSUE #5196: story(textarea): improve text-position in size S
### Feature Description

To prevent text from overlapping in the text area (Size S only), the text should be shifted down by 4px. This makes the text area 4px higher.
Figma spec: https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?node-id=15292-219425&t=7qP1iwJxEgzRyN7P-11
related to: https://github.com/sbb-design-systems/lyne-components/issues/5153

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

DavideMininni-Fincons:
Closes #5196


src/elements/form-field/form-field/form-field.scss
```diff
@@ -1,6 +1,9 @@
 @use '../../core/styles' as sbb;
 
 :host {
+  --_sbb-form-field-spacer-margin-block-end-textarea: calc(
```
jeripeierSBB, line 4: I think I'd slightly change the concept to avoid the doubled margin-block-end definition.
 - Create a token for the spacing-fixed-1x in global (...-textarea-spacer or similar)
 - Assign this token to a "--_" variable here with `:host([size='s']:where(:state(input-element-textarea))`
 - There, where the margin-block-end is defined, use a calc that includes the variable (with fallback to 0, in case it's not a textarea).

jeripeierSBB (approved):
LGTM, thanks for the update!

mcilurzo (approved):

Issue #5214, opened 2026-09-02
[Bug]: It must be possible to stretch the calendar across the full width in the dialog


florianstuerzinger:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/sbb-design-systems/lyne-components/blob/main/docs/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/sbb-design-systems/lyne-components/blob/main/docs/CODE_OF_CONDUCT.md) that this project adheres to.
- [x] I have searched the [issue tracker](https://github.com/orgs/sbb-design-systems/projects/4) for an issue that matches the one I want to file, without success.

### Bug type

Other

### It affects the following packages

Components

### Which version of Lyne Components are you using

5

### Input mode

None

### What happened?

**Describe in detail the issue you are having.**
In the dialog, content such as the calendar can no longer be displayed across the full width.

**What did you expect to happen? What happened instead? What would you like to see changed?**
In the dialog, content such as the calendar can be stretched to fill the entire width.

**Is the issue related to a specific component?**
Dialog, Calendar

### Steps to reproduce the issue

1. Step one: Open https://feature-sbbch-23968-lyne-5.dev-www.sbb.ch/de?stops=Bern_I8507000~Basel_I22&day=2026-09-02&time=15_49&moment=dep
2. Step two: Click on the date of travel button
<img width="618" height="706" alt="Image" src="https://github.com/user-attachments/assets/55407578-c778-4bbe-9c01-e130801db352" />

### Relevant log output

```shell

```

PR #5228, opened 2026-09-07
style(calendar): improve paddings of stretched calendar

jeripeierSBB:
Relates to #5214


kyubisation (commented):
As it is, I object to this. 1kb of payload for a single usage use case is not something I am ok with.

PR #5234, opened 2026-09-07
feat(menu): allow menu actions to hide icon space

### LINKED ISSUE #5050: [Bug]: Menu-Item: Icon slot remains if no Icon is provided
### Preflight Checklist

### Bug type

Functionality

Components

4.11

### Operating system

any

### Browser / Browser version

Chromium

### Input mode

Mouse/Keyboard

### Additional settings

_No response_

### What happened?

Describe in detail the issue you are having.
What did you expect to happen? What happened instead? What would you like to see changed?
Is the issue related to a specific component?

When Menu-Items are used without Icon, the icon slot remains in the layout and produces empty spaces in the menu (see screenshot).

Image

### Relevant log output

```shell

```

MarioCastigliano:
## Issue

This PR Closes #5050 

## Changes

Changes in this pull request:

- add `hide-icon-space` property to menu action components
- if no icons are set on a menu's items, hide-icon-space property is set to true on all children actions

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes

- [ ] Yes
- [x] No


src/elements/menu/menu/menu.component.ts
```diff
@@ -263,6 +265,24 @@ export class SbbMenuElement extends SbbOpenCloseBaseElement {
     this._scrollHandler.enableScroll();
   }
 
+  private _checkIcons(): void {
+    const menuItems = Array.from(
+      this.querySelectorAll?.<SbbMenuActionCommonElementMixinType>(
+        'SBB-MENU-LINK, SBB-MENU-BUTTON',
```
jeripeierSBB: nitpick, we usually use lower case
```suggestion
        'sbb-menu-link, sbb-menu-button',
```

src/elements/menu/menu/menu.component.ts
```diff
@@ -263,6 +265,24 @@ export class SbbMenuElement extends SbbOpenCloseBaseElement {
...
+
+    const hasIcons = menuItems.some((e) =>
+      e.matches(':state(has-icon-name), :state(slotted-icon)'),
+    );
+
+    if (!menuItems.length || hasIcons) {
+      return;
+    }
+
+    menuItems.map((e) => (e.hideIconSpace = true));
```
jeripeierSBB:
```suggestion
    menuItems.forEach((e) => (e.hideIconSpace = true));
```

src/elements/menu/menu/menu.component.ts
```diff
@@ -39,6 +39,7 @@ import {
   ɵstateController,
```
jeripeierSBB: I think the concept would work better, if the hideIconSpace flag is on the menu and not on the action. so like for the autocomplete panel, the consumer decides on a menu, whether he wants to preserve (in this case hide) the space or not. Furthermore, I would also not hide the icon if there is one, but just have no ident if there is no icon (similar to autocomplete panel).

in your current solution, the icon check only works initially but not if anything changes later. when the option is on the menu, this problem has gone.

Like for the autocomplete panel, the state propagation could be done via css variable.

jeripeierSBB (changes requested):
Thank you for the addition!

After a discussion with at kyubisation, we think the concept would work better, if the hideIconSpace flag is on the menu and not on the action. Like for the autocomplete panel, the consumer decides on a menu, whether he wants to preserve (in this case hide) the space or not.
Furthermore, I would also not hide the icon if there is one, but just have no ident if there is no icon (similar to autocomplete panel).

With your current solution, the icon check only works initially but not if anything changes later. Like for the autocomplete panel, the state propagation could be done via css variable.

Recommendations:
 - hideIconSpace property on the menu, propagate state down via CSS variable.
 - the check, whether all icons are missing on all slotted actions should be updated when a slotchange event occurs on the menu, an icon slot change occurs on any slotted item, or if the iconName property of any slotted item changes. For that we recommend dispatching an internal event that bubbles up to the menu. Always check that the event origin is not from a nested menu.
 - the icon check should not mutate the hideIconSpace property itself but act just as a second condition.

If you have doubts, just reach out to me :)

src/elements/menu/common/menu-action-common.ts
```diff
@@ -19,14 +25,30 @@ export const SbbMenuActionCommonElementMixin = <
     extends SbbIconNameMixin(SbbDisabledMixin(superClass))
     implements SbbMenuActionCommonElementMixinType
   {
+    public static readonly events = {
```
jeripeierSBB: for private events the static block is not needed

src/elements/menu/menu-button/menu-button.component.ts
```diff
@@ -14,6 +14,9 @@ export class SbbMenuButtonElement extends SbbDisabledTabIndexActionMixin(
 ) {
   public static override readonly elementName: string = 'sbb-menu-button';
   public static override readonly role: ElementInternals['role'] = 'menuitem';
+  public static readonly events = {
```
jeripeierSBB: You can also skip this block here

src/elements/menu/menu-link/menu-link.component.ts
```diff
@@ -13,6 +13,9 @@ export class SbbMenuLinkElement extends SbbDisabledInteractiveMixin(
   SbbMenuActionCommonElementMixin(SbbLinkBaseElement),
 ) {
   public static override readonly elementName: string = 'sbb-menu-link';
+  public static readonly events = {
```
jeripeierSBB: and here :)

src/elements/menu/menu/menu.component.ts
```diff
@@ -263,6 +277,21 @@ export class SbbMenuElement extends SbbOpenCloseBaseElement {
     this._scrollHandler.enableScroll();
   }
 
+  private _checkIcons(): void {
+    console.trace();
```
jeripeierSBB:
```suggestion
```

src/elements/menu/menu/menu.component.ts
```diff
@@ -115,6 +124,11 @@ export class SbbMenuElement extends SbbOpenCloseBaseElement {
   public constructor() {
     super();
     this.addEventListener?.('keydown', (e) => this._handleKeyDown(e));
+    this.addEventListener?.(SbbMenuButtonElement.events.iconchange, (ev) => {
```
jeripeierSBB:
```suggestion
    this.addEventListener?.('iconchange', (ev) => {
```

src/elements/menu/menu/menu.component.ts
```diff
@@ -263,6 +277,21 @@ export class SbbMenuElement extends SbbOpenCloseBaseElement {
     this._scrollHandler.enableScroll();
   }
 
+  private _checkIcons(): void {
+    console.trace();
+    const menuItems = Array.from(
+      this.querySelectorAll?.<SbbMenuActionCommonElementMixinType>(
+        'SBB-MENU-LINK, SBB-MENU-BUTTON',
```
jeripeierSBB: nitpicky, but uppercase is not needed
```suggestion
        'sbb-menu-link, sbb-menu-button',
```

src/elements/menu/menu/menu.component.ts
```diff
@@ -263,6 +277,21 @@ export class SbbMenuElement extends SbbOpenCloseBaseElement {
     this._scrollHandler.enableScroll();
   }
 
+  private _checkIcons(): void {
```
jeripeierSBB: During the review we found out, that it's probably not the best idea to automatically detect whether there are any icons slotted. We are sorry!
Could you please remove the logic and only handle the explicit property?
With that, the checkIcons() and the event could probably be removed.

The problematic case:
If there is a menu that should have the reserved space, but initially don't have any icon (like the language in the menu (although here, there is at least one checked)), then the space would only appear whenever a consumer slots an icon. But, there is no way left to control that for the consumer. 

Future:
With the next breaking change, we should call the property `preserveIconSpace` and change the logic. From then on, it's also possible again to set the space if any element with an icon have been detected.

Could you add a TODO at the property?

src/elements/menu/common/menu-action.scss
```diff
@@ -115,6 +115,11 @@ $disabled: '[disabled], :disabled, [disabled-interactive]';
   min-height: var(--sbb-size-icon-ui-small);
 }
 
+.sbb-menu-action__icon {
```
jeripeierSBB, line 117: nice!

src/elements/menu/common/menu-action.scss
jeripeierSBB: Will be overridden in any case
```suggestion
```

src/elements/menu/menu/menu.visual.spec.ts
```diff
@@ -145,6 +145,32 @@ describe(`sbb-menu`, () => {
...
+            </sbb-menu>
+          `,
+          { minHeight: '400px' },
+        );
+        setup.withPostSetupAction(async () => {
+          const button = setup.snapshotElement.querySelector<SbbButtonElement>('#menu-trigger-1')!;
+          button.click();
+
+          // Test focus outline
+          await sendKeys({ press: tabKey });
```
jeripeierSBB, line 170: is the focus outline relevant here?

src/elements/menu/menu/menu.visual.spec.ts
```diff
@@ -145,6 +145,32 @@ describe(`sbb-menu`, () => {
...
       }),
     );
+
+    it(
+      'no icons',
+      visualDiffDefault.with(async (setup) => {
+        await setup.withFixture(
+          html`
+            <sbb-button id="menu-trigger-1">Menu trigger</sbb-button>
+            <sbb-menu trigger="menu-trigger-1">
```
jeripeierSBB: With the upcoming change, you could just set the flag here and the test should stay valid

src/elements/menu/menu.stories.ts
```diff
@@ -250,6 +272,12 @@ export const Ellipsis: StoryObj = {
   args: { ...defaultArgs },
 };
 
+export const NoIcons: StoryObj = {
```
jeripeierSBB, line 275: Here we could also just set the hideIconFlag, so everything else can be kept.

jeripeierSBB (changes requested):
During the review we found out, that it's probably not the best idea to automatically detect whether there are any icons slotted. We are sorry!
Could you please remove the logic and only take the explicit property as source?
With that, the checkIcons() and the event could be removed.

The problematic case:
If there is a menu that should have the reserved space, but initially don't have any icon (like the language in the menu (although here, there is at least one checked)), then the space would only appear whenever a consumer slots an icon. But, there is no way left to control that for the consumer that it has some pre reserved space.

Future:
With the next breaking change, we should call the property preserveIconSpace and change the logic. From then on, it's also possible again to set the space if any element with an icon have been detected.

Could you add a TODO at the property and create a github issue with milestone 6 assigned?

THANK YOU!

src/elements/menu/menu/menu.visual.spec.ts
MarioCastigliano: It's to highlight the inline padding, that changes from 8px to 12px when there's no icon

src/elements/menu/menu.stories.ts
```diff
@@ -92,6 +103,17 @@ const DefaultTemplate = (args: Args): TemplateResult => html`
   </sbb-menu>
 `;
 
+const NoIconsTemplate = (args: Args): TemplateResult => html`
+  ${triggerButton('menu-trigger-1')}
+  <sbb-menu trigger="menu-trigger-1">
```
TomMenga, line 108:
```suggestion
  <sbb-menu trigger="menu-trigger-1" ?hide-icon-space=${args['hide-icon-space']}>
```

TomMenga (approved):
LGMT 👍 

just a fix in the 'no-icon' story

PR #5239, opened 2026-09-08
feat(logo): add 125 years anniversary logo

jeripeierSBB:


src/elements/logo/logo-anniversary/logo-anniversary.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  public static override readonly elementName: string = 'sbb-logo-anniversary';
+  public static override styles: CSSResultGroup = [unsafeCSS(style)];
+
+  /** Accessibility label which will be forwarded to the SVG logo. */
+  @forceType()
+  @property({ attribute: 'accessibility-label' })
+  public accessor accessibilityLabel: string = '';
+
+  /** The enabled animations. */
+  @property({ reflect: true }) public accessor animation: 'all' | 'none' = 'all';
```
jeripeierSBB: Insecure whether this pattern is good? @kyubisation

kyubisation: We do not have prior art, so not sure.
Could we just support the usual css class for disabling animation?

DavideMininni-Fincons (approved):
Generally looks good :)

Copilot references this WAI rule https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html pointing out that the total time of animation is 5.2 seconds; should we slightly decrease the initial delay?

jeripeierSBB:
> Generally looks good :)
> 
> Copilot references this WAI rule https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html pointing out that the total time of animation is 5.2 seconds; should we slightly decrease the initial delay?

Thank you! You are right. We anyways have problems with a11y because of the ultra small font. But it seems that this has to be so and isn't negotiable. Manuel already tried to figure out better ways.

PR #5268, opened 2026-09-18
feat(pearl-chain): implement component

### LINKED ISSUE #5112: story(pearl-chain): initial implementation
### Feature Description

**Specification in progress**

We want to implement the pearl chain: https://www.figma.com/design/mWknI2rC5DJmOgRO61WKai/Lyne-Components?m=dev&node-id=7491-46

Important to note; The pearl chain has no accessibility internally. If accessibility should be provided, this can be done via aria-label or aria-describedby on `<sbb-pearl-chain>`.

Also rename the `<sbb-pearl-chain>` in experimental to `<sbb-pearl-chain-legacy>`. (The other components can be left as is, as they do not have a name conflict.)

Also note that icons might change; They are currently a draft from @Frenggi

**Spec**

- `sbb-pearl-chain` component
  - Renders the pearl chain as an absolute positioned `<svg>`
  - Should expose methods `add(node: SbbPearlChainNode)` and `remove(node: SbbPearlChainNode)` for adding/removing nodes
    - The order should be calculated based on DOM order
  - Should have both ResizeObserver and IntersectionObserver to observe both itself and any added node and re-render the svg on any change
  - Renders the lines between nodes
    - If any node is more horizontally offset than vertically, we switch to horizontal mode (`:state(horizontal)`), which adds a gab in the line before any node (except _empty_, `start` or `end`) and rotates the svg in the nodes (suggestion: add CSS variable for rotation and set this variable in the pearl chain depending on state)
    - The state of a line is based on the two adjacent nodes
    - The priority is `irrelevant` + `disrupted` > `disrupted` > `irrelevant` > `walk` > `past` (dynamic state) > default
    - If the current time is between the `departure` datetime of the line starting node and the `arrival` datetime of the line ending node, the line is rendered as two parts (except when `disrupted` or `walk`), where proportionally the line before the current time is rendered gray and the line after the current time is rendered normally
      - Additionally, at the current time a red, plusing dot is rendered (the pulsing might be achievable via CSS and can overlap the node bullets) (this is not rendered, if the time is between arrival and departure of a node)
- `sbb-pearl-chain-node`
  - Represents a station/point of interest
  - Is an `inline-block` with a width of 12px, the bullet is always centered and the svg should be sized via `em`, in order for the component to grow with text size (requirement KIS)
  - Are added to the ancestor `<sbb-pearl-chain>` on connectedCallback and removed on `disconnectedCallback`
  - Has property `type?: string` (or instead of string the types below), which decides how to render the bullet
    - _empty_ (Default) Bullet will not be rendered
    - `start`/`end`: simple bullet, full color (the only one with size 10x10px, all others are 8x8px)
      `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 10"><path fill="currentcolor" stroke="currentcolor" d="M5 .5a4.5 4.5 0 1 1 0 9 4.5 4.5 0 0 1 0-9Z"/></svg>`
    - `stop`: hollowed circle
    - `boarding`: hollowed circle with lower half filled in
    - `alighting`: hollowed circle with upper half filled in
    - `skip`: hollowed circle with a diagonal red/white rectangle
    - `stop-duty`: hollowed circle, with dashed border
    - `stop-on-demand`: hollowed circle with bullet inside
    - `boarding-on-demand`: hollowed circle with half moon in the lower half
    - `alighting-on-demand`: hollowed circle with half moon in the upper half
  - Has property `disrupted?: 'arrival' | 'departure' | true | null` (where empty string should be converted to `true`, as a convenience to mark it as both ways), which colors the bullet (if it exists and the value is not null) red
  - Has property `walk?: 'arrival' | 'departure' | true | null` (where empty string should be converted to `true`, as a convenience to mark it as both ways), which DOES NOT color the bullet
  - Has property `irrelevant?: 'arrival' | 'departure' | true | null` (where empty string should be converted to `true`, as a convenience to mark it as both ways), which colors the bullet (if it exists and the value is not null) gray (supersedes disruption color)
  - Has properties `arrival?: Date` and `departure?: Date`, which are parsed as a Date with time (should accept ISO formatted datetime strings)
    - If the current time is after `departure` (if it has been defined), the color of the bullet is set to gray
  - After adding itself to the `<sbb-pearl-chain>` calls `requestUpdate()` on the `<sbb-pearl-chain>` whenever something changes that has an effect on rendering (type, disrupted, irrelevant, walk, arrival, departure)

Example:

```html
<sbb-pearl-chain>
  <table>
    <tr>
      <td>11:00</td>
      <td><sbb-pearl-chain-node type="start" departure="2026-07-21T11:00:00"></sbb-pearl-chain-node></td>
      <td>Bern</td>
    </tr>
    <tr>
      <td>12:00</td>
      <td><sbb-pearl-chain-node type="stop" arrival="2026-07-21T11:58:00" departure="2026-07-21T12:00:00"></sbb-pearl-chain-node></td>
      <td>Zürich HB</td>
    </tr>
    <tr>
      <td>12:30</td>
      <td><sbb-pearl-chain-node type="end" arrival="2026-07-21T12:30:00"></sbb-pearl-chain-node></td>
      <td>Zürich HB</td>
    </tr>
  </table>
</sbb-peal-chain>
```
```html
<sbb-pearl-chain>
  <div class="flex">
    <sbb-pearl-chain-node ...></sbb-pearl-chain-node>
    <span class="weight1"></span>
    <sbb-pearl-chain-node ...></sbb-pearl-chain-node>
    <span class="weight2"></span>
    <sbb-pearl-chain-node ...></sbb-pearl-chain-node>
  </div>
</sbb-pearl-chain>
```

### Definition of Done

- [ ] Tests are implemented
- [ ] High-contrast is tested
- [ ] UX approved

TomMenga:
This PR Closes #5112


src/elements/pearl-chain/pearl-chain-node/pearl-chain-node.scss
```diff
@@ -0,0 +1,77 @@
...
+  color: var(--sbb-pearl-chain-bullet-color);
+  background-color: var(--sbb-pearl-chain-bullet-background-color);
+  rotate: var(--sbb-pearl-chain-bullet-rotation, 0deg);
+
+  circle {
+    cx: 50%;
+    cy: 50%;
+  }
+
+  &--past {
```
kyubisation: We generally avoid using & for name extensions, as it is easier to search in full name.
What do you think?

PR #5271, opened 2026-09-21
feat(calendar): improve keyboard navigation

### LINKED ISSUE #5253: story(calendar): change week number/day keyboard navigation
In multiple mode the keyboard navigation of the week number/day button navigation should be changed as follows.

First tab into calendar lands onto the top row of week numbers or days and you can navigate to the other numbers/days via arrow keys.
Next tab inside the calendar navigates to the side row of week numbers or days (if available) and you can navigate to the other numbers/days via arrow keys.
Next tab inside the calendar navigates to the days inside the calendar.

DavideMininni-Fincons:
The PR improves the keyboard navigation for the calendar component: 
- if multiple=true, is possible to navigate with arrows and special keys in weekdays;
- if weekNumber=true, is possible to navigate with arrows and special keys in weeknumbers;
- the Tab navigation is split in groups; first, the weekdays (horizontal) or the weeknumbers (vertical), then the opposite group, then the days;
- if the focus doesn't exit the groups, the last focused element is kept and can be reached with Shift+Tab, otherwise the first of each group is the first focusable.


src/elements/calendar/calendar/calendar.component.ts
```diff
@@ -355,8 +404,46 @@ export class SbbCalendarElement<T = Date> extends SbbFormAssociatedMixin(SbbElem
...
+          this._handleKeyboardEvent(event, day);
+        }
+        if (weekday) {
+          this._handleKeyboardEventWeek<SbbCalendarWeekdayElement>(
+            event,
+            weekday,
+            this._weekdayCells,
+          );
+        }
+        if (weekNumber) {
```
kyubisation: Nitpick
```suggestion
        } else if (weekday) {
          this._handleKeyboardEventWeek<SbbCalendarWeekdayElement>(
            event,
            weekday,
            this._weekdayCells,
          );
        } else if (weekNumber) {
```

src/elements/calendar/calendar/calendar.component.ts
```diff
@@ -1011,12 +1102,18 @@ export class SbbCalendarElement<T = Date> extends SbbFormAssociatedMixin(SbbElem
...
     );
+    this._setWeekFocus();
     const firstFocusable = this._getFirstFocusable();
     if (firstFocusable) {
       firstFocusable.tabIndex = 0;
     }
   }
 
+  private _setWeekFocus(): void {
+    Array.from(this._weekdayCells).forEach((e, i) => (e.tabIndex = i === 0 ? 0 : -1));
```
kyubisation: Array.from should not be necessary here as the getter already returns an array?

kyubisation (approved):
LGTM 👍 
Thanks!

PR #5272, opened 2026-09-21
fix(dialog): always display divider for actions

kyubisation:
This change allows a clearer visual distinction between the dialog content and actions.


TomMenga (approved):
LGTM 👍

